PR #32, opened 2021-11-01
feat(preset-uno): Supplementary transform style attribute

edwardnyc:
Minor adjustment  [Supplementary transform style attribute]


packages/preset-uno/src/rules/transform.ts
```diff
@@ -16,6 +16,9 @@ export const transforms: Rule[] = [
       'transform': 'rotate(var(--un-rotate)) scaleX(var(--un-scale-x)) scaleY(var(--un-scale-y)) scaleZ(var(--un-scale-z)) skewX(var(--un-skew-x)) skewY(var(--un-skew-y)) translateX(var(--un-translate-x)) translateY(var(--un-translate-y)) translateZ(var(--un-translate-z))',
     },
   ],
+  [/^(?:preserve-)?(3d|flat)$/, ([, value]) => ({
```
antfu, line 19:
```suggestion
  [/^preserve-(3d|flat)$/, ([, value]) => ({
```

Any reason that we support the unprefixed one here? I'd feel it might be too general.

antfu: I will made the change for now, and feel free to bring the discussion later

PR #39, opened 2021-11-02
feat: typography utilities

hannoeru:


packages/preset-uno/src/rules/static.ts
```diff
@@ -84,14 +84,100 @@ export const textDecorations: Rule[] = [
...
+export const writingModes: Rule[] = [
+  ['write-normal', {
+    '-webkit-writing-mode': 'horizontal-tb',
+    '-ms-writing-mode': 'lr-tb',
+    'writing-mode': 'horizontal-tb',
+  }],
+  ['write-vertical-right', {
+    '-webkit-writing-mode': 'vertical-rl',
+    '-ms-writing-mode': 'tb-rl',
+    'writing-mode': 'vertical-rl',
```
antfu, line 156: Most of them are actually suppress 95% coverage on caniuse, I think we could omit those prefixed ones.

packages/preset-uno/src/rules/static.ts
```diff
@@ -84,14 +84,67 @@ export const textDecorations: Rule[] = [
...
+]
+
+export const fontSmoothings: Rule[] = [
+  ['antialiased', {
+    '-webkit-font-smoothing': 'antialiased',
+    '-moz-osx-font-smoothing': 'grayscale',
+  }],
+  ['subpixel-antialiased', {
+    '-webkit-font-smoothing': 'auto',
+    '-moz-osx-font-smoothing': 'auto',
```
antfu, line 113: Could we add a unprefixed one here as well?

hannoeru: Looks like there is no browser support unprefixed ones.

https://caniuse.com/?search=font-smoothing

antfu: Yes, but in the future right? 👀

PR #40, opened 2021-11-02
fix(preset-uno): negative space utilities

hannoeru:
close https://github.com/antfu/unocss/issues/37

### DIFF
--- a/packages/preset-uno/src/variants/index.ts
+++ b/packages/preset-uno/src/variants/index.ts
@@ -36,8 +36,6 @@ export const variantNegative: Variant = {
         body: (body) => {
           body.forEach((v) => {
             v[1] = v[1]?.toString().replace(/[0-9.]+[a-z]+/, i => `-${i}`)
-            if (!v[0].includes('reverse') && v[1]?.toString().match(/^\d/))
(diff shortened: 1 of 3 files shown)


packages/preset-uno/src/variants/index.ts
```diff
@@ -35,7 +35,8 @@ export const variantNegative: Variant = {
         matcher: matcher.slice(1),
         body: (body) => {
           body.forEach((v) => {
-            if (v[1]?.toString().match(/^\d/))
+            v[1] = v[1]?.toString().replace(/[0-9.]+[a-z]+/, i => `-${i}`)
+            if (!v[0].includes('reverse') && v[1]?.toString().match(/^\d/))
```
antfu, line 39:
```suggestion
            if (!v[0].includes('reverse') && v[1]?.toString().match(/^\d/))
```

`v[0].includes('reverse')` this one feels a bit tricky to me. Can we have a better workaround?

hannoeru: Actually we can remove L39-40, and all tests still passed.

antfu: Sure, let's do that then

PR #44, opened 2021-11-02
feat(preset-uno): behaviors utilities

edwardnyc:
- Appearance
-  caretOpacity
-  caretColors
-  boxDecorationBreaks

@userquin Thanks for reminding.


hannoeru (commented):
I think we can reuse `colorResolver` in https://github.com/antfu/unocss/blob/main/packages/preset-uno/src/rules/color.ts

packages/preset-uno/src/rules/behaviors.ts
```diff
@@ -27,3 +30,87 @@ export const listStyle: Rule[] = [
...
+
+export const imageRenderings: Rule[] = [
+  ['image-render-auto', { 'image-rendering': 'auto' }],
+  ['image-render-edge', { 'image-rendering': 'crisp-edges' }],
+  ['image-render-pixel', [
+    ['-ms-interpolation-mode', 'nearest-neighbor'],
+    ['image-rendering', '-webkit-optimize-contrast'],
+    ['image-rendering', '-moz-crisp-edges'],
+    ['image-rendering', '-o-pixelated'],
+    ['image-rendering', 'pixelated'],
```
edwardnyc, line 108: maybe autoprefixer can resolve the prefixs?

antfu: Yes, but that would require an extra layer of postcss and autoprefixer, which I would actually want to avoid by default (but you can still have if you want to).

edwardnyc:
> I think we can reuse `colorResolver` in https://github.com/antfu/unocss/blob/main/packages/preset-uno/src/rules/color.ts

Yeah, I'll change it

packages/preset-uno/src/rules/behaviors.ts
hannoeru: Without prefix it can cover 95.64%

https://caniuse.com/?search=image-rendering

userquin:
> 
> 
> > I think we can reuse `colorResolver` in https://github.com/antfu/unocss/blob/main/packages/preset-uno/src/rules/color.ts
> 
> Yeah, I'll change it

beware, on `colors.ts` also handles `inherits` and it is not covered on caret colors.

packages/preset-uno/src/rules/behaviors.ts
```diff
@@ -27,3 +29,34 @@ export const listStyle: Rule[] = [
...
+  ]],
+]
+
+export const Appearance: Rule[] = [
```
antfu, line 58:
```suggestion
export const appearance: Rule[] = [
```

packages/preset-uno/src/rules/index.ts
```diff
@@ -17,7 +17,7 @@ import { paddings, margins } from './spacing'
 import { verticalAligns, appearances, breaks, cursors, displays, pointerEvents, resizes, textAligns, textDecorations, textOverflows, textTransforms, userSelects, whitespaces } from './static'
 import { tables } from './table'
 import { transforms } from './transform'
-import { listStyle } from './behaviors'
+import { listStyle, caretColors, boxDecorationBreaks, caretOpacity, imageRenderings, Appearance } from './behaviors'
```
antfu, line 20:
```suggestion
import { listStyle, caretColors, boxDecorationBreaks, caretOpacity, imageRenderings, appearance } from './behaviors'
```

packages/preset-uno/src/rules/index.ts
```diff
@@ -57,6 +57,11 @@ export const rules: Rule[] = [
   userSelects,
   whitespaces,
   listStyle,
+  caretColors,
+  boxDecorationBreaks,
+  caretOpacity,
+  imageRenderings,
+  Appearance,
```
antfu, line 64:
```suggestion
  appearance,
```

edwardnyc:
> > > I think we can reuse `colorResolver` in https://github.com/antfu/unocss/blob/main/packages/preset-uno/src/rules/color.ts
> > 
> > 
> > Yeah, I'll change it
> 
> beware, on `color.ts` also handles `inherit` and it is not covered on caret colors.

I tried. It's OK. here,  [w3c](https://www.w3schools.com/cssref/css3_pr_caret-color.asp)

userquin:
I add a reminder to add to this module: 
- #45

PR #46, opened 2021-11-02
perf: encodeSvg

HomyeeKing:
have look through [this article](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris)

the core step are follows :
image

so we can just use `encodeURI` directly to escape unsafe-url to avoid the redundant regexp match 

also it seems that remain the single quote to reduce the size, so I just keep it 

this is test code for benchmark
```js
function encodeSvg(svg) {
    return svg
      .replace(
        "<svg",
        ~svg.indexOf("xmlns") ? "<svg" : '<svg xmlns="http://www.w3.org/2000/svg"'
      )
      .replace(/"/g, "'")
      .replace(/%/g, "%25")
      .replace(/#/g, "%23")
      .replace(/{/g, "%7B")
      .replace(/}/g, "%7D")
      .replace(/</g, "%3C")
      .replace(/>/g, "%3E");
  }
  
  function encodeSvg1(svg) {
    return encodeURI(svg.replace(/"/g, "'"));
  }
  
  const str = `
    <svg version="1.1"
         width="300" height="200"
         xmlns="http://www.w3.org/2000/svg">
    
      <rect width="100%" height="100%" fill="red" />
    
      <circle cx="150" cy="100" r="80" fill="green" />
    
      <text x="150" y="125" font-size="60" text-anchor="middle" fill="white">SVG</text>
    
    </svg>
    `;
  
  function doTest() {
    for (let i = 0; i < 10; i++) {
      console.time("encodeSvg");
      for (let i = 1; i < 100_0000; i++) {
        encodeSvg(str);
      }
  
      console.timeEnd("encodeSvg");
  
      console.time("encodeSvg1");
      for (let i = 1; i < 100_0000; i++) {
        encodeSvg1(str);
      }
  
      console.timeEnd("encodeSvg1");
    }
  }
  
  doTest()

```

### DIFF
--- a/packages/preset-icons/src/utils.ts
+++ b/packages/preset-icons/src/utils.ts
@@ -11,12 +11,5 @@ export function warnOnce(msg: string) {
 
 // https://bl.ocks.org/jennyknuth/222825e315d45a738ed9d6e04c7a88d0
 export function encodeSvg(svg: string) {
-  return svg.replace('<svg', (~svg.indexOf('xmlns') ? '<svg' : '<svg xmlns="http://www.w3.org/2000/svg"'))
-    .replace(/"/g, '\'')
-    .replace(/%/g, '%25')
-    .replace(/#/g, '%23')
-    .replace(/{/g, '%7B')
-    .replace(/}/g, '%7D')
-    .replace(/</g, '%3C')
-    .replace(/>/g, '%3E')
+  return encodeURI(svg.trim().replace(/"/g, '\''))
 }


userquin:
you have the same perf if you remove the `xmlns` replacement on `encodeSvg` or add it to `encodeSvg1`, for example, removing the replacement from  `encodeSvg`:
```shell
encodeSvg: 2693.201171875 ms
encodeSvg1: 2544.139892578125 ms
encodeSvg: 2557.42919921875 ms
encodeSvg1: 2438.81298828125 ms
encodeSvg: 2611.39990234375 ms
encodeSvg1: 2368.839111328125 ms
encodeSvg: 2659.643310546875 ms
encodeSvg1: 2596.5078125 ms
encodeSvg: 2741.411865234375 ms
encodeSvg1: 2600.990966796875 ms
encodeSvg: 2654.531982421875 ms
encodeSvg1: 2593.3798828125 ms
encodeSvg: 2684.7890625 ms
encodeSvg1: 2579.383056640625 ms
encodeSvg: 2542.64111328125 ms
encodeSvg1: 2434.47509765625 ms
encodeSvg: 2513.14306640625 ms
encodeSvg1: 2536.263916015625 ms
encodeSvg: 2486.60595703125 ms
encodeSvg1: 2425.113037109375 ms
```

antfu:
The test is failing. From the result, I can see the space is been encoded into `%20` which I don't think it's optimal. Given the perf improvement is not significant, I would prefer to generate smaller code for users. Thanks for bringing this up.

PR #47, opened 2021-11-03
feat(preset-uno): support variables

edwardnyc:
Support variables
fix [38](https://github.com/antfu/unocss/issues/38)


packages/preset-uno/src/rules/behaviors.ts
```diff
@@ -60,3 +60,26 @@ export const appearance: Rule[] = [
     'appearance-none': 'none',
   }],
 ]
+
+// There is no match in the rule，like `bg`
+const unMatchedAbbreviation: Record<string, string> = {
+  bg: 'background-color',
+}
```
antfu, line 67: I think we better maintain a map for known keys instead of bypassing everything

edwardnyc: I added some abbreviation attributes by referring to the existing rules. What do you think of this time~

PR #48, opened 2021-11-03
feat(reset): add sanitize.css

johannschopplich:


packages/reset/scripts/copy.ts
```diff
@@ -4,3 +4,20 @@ fs.copyFileSync(
...
+  'forms.css',
+  'assets.css',
+  'typography.css',
+  'reduce-motion.css',
+  'system-ui.css',
+  'ui-monospace.css',
+]) {
+  fs.copyFileSync(
+    `node_modules/sanitize.css/${stylesheet}`,
+    `sanitize.css/${stylesheet}`,
```
antfu, line 21:
```suggestion
    `sanitize/${stylesheet}`,
```

packages/reset/sanitize.css/assets.css
```diff
@@ -0,0 +1,8 @@
+/**
```
antfu, line 1: I guess we'd better exclude them from the source control.

packages/reset/scripts/copy.ts
johannschopplich: Good idea.

packages/reset/sanitize.css/assets.css
johannschopplich: True. 🙂

johannschopplich: Have pushed a commit addressing your suggestion.

PR #49, opened 2021-11-03
feat(preset-uno): outline utilities

edwardnyc:


edwardnyc:
> It would be nice if we can align these utilities with [Tailwind v3's upcoming outline utilities](https://github.com/tailwindlabs/tailwindcss/pull/5887) which is have very similar composition like ring utilities.

Thank you for your suggestion, but I think this practical change should depend on the overall implementation of the current library. This may be better with the upgrading of the warehouse body?

I think it's painful for users to understand different specifications.    @sibbng

sibbng:
> I think it's painful for users to understand different specifications. @sibbng

If you are talking about having multiple outline utilities with similar API, there is nothing "painful" with that since both utilities are have nearly exact same API. Also I'm curious about which users are you talking about. UnoCSS currently doesn't have detailed documentation and all its users are diving into the source code to understand its capacity. Once we have documentation we can document differences between these two utilities to make it less "painful" for users:

- Ring utilities can have offset color, outline utilities can't
- Outline utilities have very sluggish animations _(tested on on Chrome)_

Also I don't even think Tailwind team will remove ring utilities. Because Tailwind team removed AOT engine on v3 and they are on the way of pushing Tailwind's capabilities to what JIT engine can do. This is just a upgrade for outline properties which was not quite ideal while living with AOT engine.

Also I would suggest removing `outline-color` handler and handle colors in `colors.ts` like how ring utilities do without `-colors` appendix:
https://github.com/antfu/unocss/blob/main/packages/preset-uno/src/rules/color.ts#L122

packages/preset-uno/src/rules/behaviors.ts
```diff
@@ -2,6 +2,60 @@ import { Rule } from '@unocss/core'
...
+      return {
+        'outline-width': size,
+      }
+    },
+  ],
+  [
+    /^outline-color-(.+)$/, colorResolver('outline-color', 'outline-color'),
+  ],
+  [
+    new RegExp(`^outline-(${outlineStyle.join('|')})-(.+)$`), (reg, config) => {
```
antfu, line 43: I think it would be better to avoid creating dynamic RegExp if possible. We could do the check inside function body instead.

antfu:
I would also prefer to have the new capability. For `preset-uno`, it's goal is trying to blend many different conventions together, not only tailwind and windi but probably a few others. The idea here is that, since we are on-demand, there is almost nothing to sacrifice supporting them all. It's also kinda a showcase preset to see the limitation of UnoCSS, and for those who don't care about the "strict constariant" (like me). For those who really prefer a "spec", I would recommend creating their own presets (which also give them the full control).

packages/preset-uno/src/rules/behaviors.ts
```diff
@@ -2,6 +2,83 @@ import { Rule } from '@unocss/core'
...
+  }
+}
+
+export const outline: Rule[] = [
+  ['outline', { 'outline-style': 'solid' }],
+  [
+    /^outline-(.+)$/, (match, config) => {
+      const [, d] = match
+
+      const handlerList = [
```
antfu, line 38: Why handlers are needed? Could we use `if ... else` directly?

antfu: I have updated it

edwardnyc: EMM, because I don't think these blocks are coupled, when I find logic, I need to read it from top to bottom, rather than directly find the corresponding branch with the problem. 

However, there is not much logic in this function, which will also cause the problem of seemingly over design. Then listen to your opinion!

It seems that our biological clock is the opposite. I will deal with the mail immediately when I see it. I'm very sorry that I didn't repair it according to your opinion.

antfu: No worry about the timezone at all, I was just a bit rushed to push the new release :P

PR #50, opened 2021-11-03
feat(preset-uno): overscrolls utilities

edwardnyc:


packages/preset-uno/src/rules/behaviors.ts
```diff
@@ -60,3 +60,17 @@ export const appearance: Rule[] = [
...
+  'hidden',
+  'visible',
+  'scroll',
+  'contain',
+]
+
+export const overscrolls: Rule[] = [
+  [/^overscroll-(.+)$/, ([, v]) => overflowValues.includes(v) ? { 'overscroll-behavior': v } : undefined],
+  [/^overscroll-([xy])-(.+)$/, ([, d, v]) => overflowValues.includes(v) ? { [`overscroll-behavior-${d}`]: v } : undefined],
+]
```
hannoeru, line 76: Instead of dynamic rules, I think we can map `overflowValues` to create a list of static rule, which is faster.

edwardnyc: Thank you for your guidance. I think such a small amount of code has little impact on performance, but if you use a static scheme, you need to maintain a long string of configuration code.


You are a team member and have more experience than me. If you insist, I will change. @hannoeru

antfu: I don't mind keeping the current format. We could do some pref improve when it becomes an issue.

Issue #64, opened 2021-11-05
Support stacking group variants


sibbng:
Tailwind's [new simplified addVariant API](https://github.com/tailwindlabs/tailwindcss/pull/5809) have a custom pseudo/function that merges same classes into one class. This is especially useful for group variants. You can search `group-hover:group-focus:text-center` class on page to jump to example.

PR #67, opened 2021-11-05
feat: :merge() function

sibbng:
Follow up of #64


antfu:
I'd feel this might be a bit too magical and less intuitive?

sibbng:
> I'd feel this might be a bit too magical and less intuitive?

I agree. Lets hold this until we get more resource about use cases.

PR #69, opened 2021-11-06
feat(preset-uno): additional animation utilities

### LINKED ISSUE #45: add more behaviors
Just a reminder we need to add (from https://windicss.org/utilities/behaviors.html):
- **Outline**: `Outline Solid` and `Outline Dotted`
- **Overflow** and **Overscroll Behavior**
- **Placeholder Color**
- **Placeholder Opacity**
- **Resize**
- **User Select**
- **Screen Readers Access**

**Animations** will be included in its own module, @antfu is working to `hoist` `keyframes`.

edwardnyc:
- animation-{number}  Default to animation-duration
- Define speed through cssVar

[Last todo](https://github.com/antfu/unocss/issues/45#issuecomment-959991928)

closes #45


packages/preset-uno/src/rules/animation.ts
```diff
@@ -14,4 +15,14 @@ export const animations: Rule[] = [
...
   }],
   ['animate-none', { animation: 'none' }],
+  [/^animation(?:-duration)?-((\d+)(?:(s|ms)?))$/, ([, d]) => {
+    return {
+      'animtion-duration': h.bracket.time(d.replace(/-duration/, '')),
+    }
+  }],
+  [/^animation-delay-((\d+)(?:(s|ms)?))$/, ([, d]) => {
+    return {
+      'animtion-delay': h.bracket.time(d),
```
userquin, line 25: review all keys, misssing `a`

userquin:
missing speed: `/^animate-speed-(inherit|initial|revert|unset|linear|ease|ease-in|ease-out|ease-in-out)$/`, we should cover  `cubic-bezier` and `steps` (https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timing-function#syntax)?

packages/preset-uno/src/rules/animation.ts
edwardnyc: Yes, I'm so stupid! 
resolved~

userquin:
For `steps` we can handle in another rule: 
`/^animation-steps-([1-9][0-9]*)-(jump-start|jump-end|jump-none|jump-both|start|end)$/`

userquin:
update test snapshots

userquin:
> jump

Can be included in `speed`:

```ts
/^animate-speed-(inherit|initial|revert|unset|linear|ease|ease-in|ease-out|ease-in-out|steps-([1-9][0-9]*)-(jump-start|jump-end|jump-none|jump-both|start|end))$/
```

edwardnyc:
> For `steps` we can handle in another rule: `/^animation-steps-([1-9][0-9]*)-(jump-start|jump-end|jump-none|jump-both|start|end)$/`

Thank you for your advice.  
I put this attribute in CSS variables. Because I don't think even the attributes you provide can fully meet the needs of users. Maybe CSS variables can allow them to customize and implement animation attributes

userquin:
here basic regex for `bezier-cubic`:
 ```ts
/^animate-cubic-bezier-(0(\.[1-9])?|1)-((-)?[0-9]+(\.[0-9]+)?)-(0(\.[0-9])?|1)-((-)?[0-9]+(\.[0-9]+)?)$/
```

**Comment by userquin** [comment]
can you add also https://developer.mozilla.org/en-US/docs/Web/CSS/animation-play-state#syntax?

File: packages/preset-uno/src/rules/animation.ts [hunk 2]
```diff
@@ -14,4 +15,37 @@ export const animations: Rule[] = [
...
+    }
+  }],
+  [/^animation-delay-((\d+)(?:(s|ms)?))$/, ([, d]) => {
+    return {
+      'animation-delay': h.bracket.time(d),
+    }
+  }],
+  [
+    /^animation-(?:fill-)?mode-(none|forwards|backwards|both)$/, ([, d]) => ({
+      'animation-fill-mode': d,
```
userquin, line 30: what about global values https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode#syntax?

packages/preset-uno/src/rules/animation.ts
```diff
@@ -14,4 +15,37 @@ export const animations: Rule[] = [
...
+    }),
+  ],
+  [
+    /^animation-direction-(normal|reverse|alternate|alternate-reverse)$/, ([, d]) => ({
+      'animation-direction': d,
```
userquin, line 35: global values https://developer.mozilla.org/en-US/docs/Web/CSS/animation-direction#syntax?

packages/preset-uno/src/rules/animation.ts
```diff
@@ -14,4 +15,37 @@ export const animations: Rule[] = [
...
+    }),
+  ],
+  [
+    /^animation-(?:iteration-)?count-(.+)$/, ([, d]) => {
+      const value = d.replace(/\-/g, ', ')
```
userquin, line 40: global values https://developer.mozilla.org/en-US/docs/Web/CSS/animation-iteration-count#syntax?

packages/preset-uno/src/rules/animation.ts
userquin: I don't know if we want to handle global values, but the spec is there

edwardnyc:
Emmm....

In fact, I have never used these global attributes to write animation. 

It's hard to understand the purpose of these global APIs(for animation)@userquin

userquin:
> 
> 
> Emmm....
> 
> In fact, I have never used these global attributes to write animation.
> 
> It's hard to understand the purpose of these global APIs(for animation)@userquin

I have never use it... I just see the specs and ask

edwardnyc:
> here basic regex for `bezier-cubic`:
> 
> ```ts
> /^animate-cubic-bezier-(0(\.[1-9])?|1)-((-)?[0-9]+(\.[0-9]+)?)-(0(\.[0-9])?|1)-((-)?[0-9]+(\.[0-9]+)?)$/
> ```
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timing-function#values

In my opinion, we deal with so much logic to restrict users, or add an animation step we define. It does not bring any convenience to users. I think I'd prefer to implement it by customizing variables  @userquin 

Global APIs has been added

packages/preset-uno/src/rules/animation.ts
```diff
@@ -14,4 +15,42 @@ export const animations: Rule[] = [
     return `${keyframes[name]}\n${constructCSS({ animation: `${name} 1s linear infinite` })}`
   }],
   ['animate-none', { animation: 'none' }],
+  [/^animate(?:-duration)?-((\d+)(?:(s|ms)?))$/, ([, d]) => {
+    return {
+      'animation-duration': h.bracket.time(d.replace(/-duration/, '')),
+    }
+  }],
+  [/^animate-delay-((\d+)(?:(s|ms)?))$/, ([, d]) => {
+    return {
```
userquin, line 24: can we support negative values?: for context see https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay#values

packages/preset-uno/src/rules/animation.ts
```diff
@@ -14,4 +15,42 @@ export const animations: Rule[] = [
...
+      'animation-delay': h.bracket.time(d),
+    }
+  }],
+  [
+    /^animate-(?:fill-)?mode-(none|forwards|backwards|both|inherit|initial|revert|unset)$/, ([, d]) => ({
+      'animation-fill-mode': d,
```
userquin, line 30: should remove `fill-` from value

packages/preset-uno/src/rules/animation.ts
```diff
@@ -14,4 +15,42 @@ export const animations: Rule[] = [
...
+    }),
+  ],
+  [
+    /^animate-(?:direction)?-(normal|reverse|alternate|alternate-reverse|inherit|initial|revert|unset)$/, ([, d]) => ({
+      'animation-direction': d,
```
userquin, line 35: should include second `-` inside `direction` captring group and remove  `direction-` from value

packages/preset-uno/src/rules/animation.ts
```diff
@@ -14,4 +15,42 @@ export const animations: Rule[] = [
...
+    }),
+  ],
+  [
+    /^animate-(?:iteration-)?count-(.+)$/, ([, d]) => {
+      const value = d.replace(/\-/g, ', ')
```
userquin, line 40: shoukd remove `iteration-` from value

userquin: upps, sorry

packages/preset-uno/src/rules/animation.ts
userquin: forgot about remove value, just the second `-`

packages/preset-uno/src/rules/animation.ts
userquin: upps, sorry

packages/preset-uno/src/rules/animation.ts
userquin: should we also force the unit? on the spec it is required and we have `time` with `ms` as default value: it would be better `seconds` since it is an animation?

also, this regex will not support decimals, `2.3s` and `150.25ms` are valid values: https://developer.mozilla.org/en-US/docs/Web/CSS/time

userquin: time spec from here : https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay#formal_syntax

edwardnyc: > can we support negative values?: for context see https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay#values

I considered this problem when implementing. Multiple groups of attributes only appear when there are multiple groups of animations on the same element. This user scenario is too customized. Even if preset supports multiple groups of attributes, it is still difficult for users to easily complete animation. In this complex scenario, implementing custom `shortcuts` according to requirements should make them easier to implement

edwardnyc: > should we also force the unit? on the spec it is required and we have `time` with `ms` as default value: it would be better `seconds` since it is an animation?
> 
> also, this regex will not support decimals, `2.3s` and `150.25ms` are valid values: https://developer.mozilla.org/en-US/docs/Web/CSS/time

Excellent point of view! I never used negative numbers, thought there was no scene. Learned from you. it had been solved

userquin (approved):

userquin:
@edwardnyc can you remove this comment `// TODO: support custom speed, duration, delay` from `animation.ts` module?

userquin:
@edwardnyc FYI: I'm including https://windicss.org/plugins/community/animations.html

packages/preset-uno/src/rules/animation.ts
```diff
@@ -7,11 +8,56 @@ const keyframes: any = {
   bounce: '@keyframes bounce {0%, 100% {transform:translateY(-25%);animation-timing-function:cubic-bezier(0.8,0,1,1);} 50% {transform:translateY(0);animation-timing-function:cubic-bezier(0,0,0.2,1);}}',
 }
 
-// TODO: support custom speed, duration, delay
 // https://windicss.org/plugins/community/animations.html
 export const animations: Rule[] = [
   [/^animate-(none|spin|ping|pulse|bounce)$/, ([, name], { constructCSS }) => {
```
userquin, line 13: remove `none`

antfu, line 13:
```suggestion
  [/^animate-(spin|ping|pulse|bounce)$/, ([, name], { constructCSS }) => {
```

PR #71, opened 2021-11-07
proposal(core): support onTrigger

edwardnyc:
This is just a proposal. I just wrote a little to show my intention. 

In my plan, it is used to customize the hit or match of troubleshooting attributes.In Env mode, let users know how the set matching rules work in the generator

Do you think it is necessary to implement this function? If not, I'll give it up


antfu:
Thanks for working on that. I got your point and I think I have a better solution for this, keep turned :)

PR #86, opened 2021-11-09
fix(preset-uno): add screenReadersAccess to rules

### LINKED ISSUE #84: sr-only not working
I hope I didn't get it wrong, but shouldn't the following work already?

`<div class="sr-only">Test</div>`

The text is still visible and no css is generated.

I found the rule here:
https://github.com/antfu/unocss/blob/e6c53687da7f073bd89d3a0e968f33a68de0fb63/packages/preset-uno/src/rules/static.ts#L125-L152

Reproduction:
https://unocss.antfu.me/?html=DwEwlgbgBAzgTgWgPYDsA2BPAfAKClAFQFMYAXHYAenAl1EigGM0BDGGAXgCJ5l0MuufMTIVqkLEA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmcAvnOlBCHAOQCuWEAxgM6u0BQnqAHpLBQYAhvQA28NJhz5CwIgAoEnOHCjjUrAFxwA2itV7azeqxjUAtOrGpaAGkRxmEMdB20oqZLTIBdA-6kAJRAA&options=N4IgzgLgTglgxhEAuaBXApgXyA

I don't know if the playground is on latest unocss, but I guess so?!
I also tried it with vitesse-nuxt3, with unocss 0.6.4, where this rule should be included already. But still all sr-only elements are visible in browser. 

Please let me know, if more information is needed, I'm pretty new to this

lstegger:
If I got it right, this would close #84

My first pull request ever, hope everything is alright?!

### DIFF
--- a/packages/preset-uno/src/rules/index.ts
+++ b/packages/preset-uno/src/rules/index.ts
@@ -24,6 +24,7 @@ import { questionMark } from './question-mark'
 import { textAligns, verticalAligns } from './align'
 
 export const rules: Rule[] = [
+  screenReadersAccess,
   cssVariables,
   paddings,
   margins,
@@ -108,7 +109,6 @@ export const rules: Rule[] = [
   filters,
   tables,
   transforms,
-  screenReadersAccess,
 
   // should be the last
   questionMark,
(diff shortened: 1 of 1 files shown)


packages/preset-uno/src/rules/index.ts
```diff
@@ -108,6 +108,7 @@ export const rules: Rule[] = [
   filters,
   tables,
   transforms,
+  screenReadersAccess,
```
antfu, line 111: I guess this should go in the beginning of the list so other rules can override it.

lstegger: I moved it to the beginning, please check! Thanks for your quick reply, highly appreciated!

PR #90, opened 2021-11-09
feat(core): utilities preprocess

schelmo:
i dont know if the type for prefix is good
tried with 
`prefix?: string  | undefined`
but this didnt work

do you think we should ditch the prefix for preset-icons, since we got now a global prefix?
i think its still useful for readability to keep it there.


antfu:
Generally, I think this feature makes sense. But I am not sure if it's clear enough to have the prefix works after applying the variants. I guess it might be better to apply before variants?

```
// ok
uno:hover:text-red4
uno:!p-4
uno:btn

// no
!uno:p-4
hover:uno:text-red4
```

What do you think?

> Re: icon prefix

I think we should still have it, the prefix for icons is more like to disguise from the normal utils, while the global prefix is for whether it should be processed by UnoCSS.

schelmo:
i implemented it like the windi/tailwind behaviour.
maybe we can have an option (smth like `prefixUtilities: boolean`) to still go with the windi/tw behaviour?

the problem i see with the prefix at start is the attributify mode.
we have to check the prefix before the matchVariants call, like this then:
```ts
matchPrefix(raw: string): string | undefined {
  const matches = raw.match(new RegExp('^\\[?' + this.config.prefix + '(.*)')
  if (!matches) return
  return matches[1]
}
```

also we must store both the prefixed and unprefixed selector in the `VariantMatchedResult` to generate the right classes later
```ts
export type VariantMatchedResult = readonly [
  raw: string,
  unprefixed: string,
  current: string,
  variants: VariantHandler[]
]
```

schelmo:
maybe this way?

antfu:
I would avoid adding yet another option (want to keep the config minimal). Please give me some more time to think about it and I will see if I can find some way better. Thanks

antfu:
@schelmo I decided to expose a `preprocess` function and let users have fine-grain control of whether utilities should be included or not. Updated the test to show case the implementation of both styles of prefixing.

PR #97, opened 2021-11-10
feat(preset-uno): fixes typo in statics

### LINKED ISSUE #96: potential typo in preset-uno 'break-works'
is [this](https://github.com/antfu/unocss/blob/main/packages/preset-uno/src/rules/static.ts#L47) supposed to be `break-word` instead?

```js
export const breaks: Rule[] = [
  ['break-normal', { 'overflow-wrap': 'normal', 'word-break': 'normal' }],
  ['break-works', { 'overflow-wrap': 'break-word' }],
  ['break-all', { 'word-break': 'break-all' }],
]
```

benkroeger:
fixes #96


userquin:
@benkroeger can you add some test on preset-uno?

PR #101, opened 2021-11-11
feat: @unocss/cli

johannschopplich:
Draft of the CLI. ☀️

Notes/questions:
- I'm having trouble executing the CLI inside the monorepo setup. No tsup build error thrown, but executing won't work... Works fine standalone over at [unocss-cli](https://github.com/johannschopplich/unocss-cli). Am I importing the UnoCSS deps correctly? 
- Consequently, tests are failing at the moment.
- Is the cli name `unocss` ok with you? Would you prefer `unocss-cli` or rather something else?

Some help to migrate is much appreciated. Thanks in advance! 🙂


johannschopplich:
Thank you very much! Should have found those minor issues myself. 🙈

johannschopplich:
Tests are actually working locally. As the CLI depends on `@unocss/core/dist/index`, deps have to be built first. Afterwards, the CLI works and tests pass.

userquin:
it seems you also need to build `config` package...

userquin:
just configure `pnpm run build && pnpm -C cli run build`, it will build all packages and then the `cli`

johannschopplich:
> just configure `pnpm run build && pnpm -C cli run build`, it will build all packages and then the `cli`

OK. That didn't work locally. I may have to pass the complete package name.

userquin:
it is working, just run `test:update` script

userquin:
maybe you will need to merge main

package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "@unocss/monorepo",
-  "version": "0.7.6",
+  "version": "0.7.7",
   "private": true,
   "scripts": {
     "bench": "npm -C bench run bench",
-    "build": "pnpm -r --parallel --filter !playground run build && pnpm -r run post",
+    "build": "pnpm -r --parallel --filter !playground run build && pnpm -C packages/cli run build && pnpm -r run post",
```
userquin, line 7: should be:
`"build": "pnpm -r --parallel --filter !playground --filter !cli run build && pnpm -C cli run build && pnpm -r run post"`

you also need to run `update:test` script since the snapshots are changed: maybe will be better to merge main into your branch

johannschopplich:
@userquin Yes, that was the case. Also, thanks for the hint to use `pnpm -C cli run build` – I just had to replace `cli` with `packages/cli` and it worked. 

I will update the build command right away.

userquin:
finally 🥳

johannschopplich:
Sorry it took that many attempts. 🙈

userquin:
> 
> 
> Sorry it took that many attempts. 🙈

you should remove `package/cli` from the build script: `-C` will do it

johannschopplich:
@userquin Locally the following error is thrown if just using `cli`, rather than `packages/cli`:

```
 ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND  No package.json (or package.yaml, or package.json5) was found in "/Users/johann/Git/unocss/cli".
 ELIFECYCLE  Command failed with exit code 1.
```

userquin:
> 
> 
> @userquin Locally the following error is thrown if just using `cli`, rather than `packages/cli`:
> 
> ```
>  ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND  No package.json (or package.yaml, or package.json5) was found in "/Users/johann/Git/unocss/cli".
>  ELIFECYCLE  Command failed with exit code 1.
> ```

But the `package.json` is there: see your path `/Users/johann/Git/unocss/cli`  should be `/Users/johann/Git/unocss/packages/cli` you have something wrong in your codebase.

userquin:
Maybe you need to add the cli to root `tsconfig.json` file: can you test it?

johannschopplich:
I'm running from root. [The docs](https://pnpm.io/pnpm-cli#-c-path---dir-path) state for using `-C`:
> Run as if pnpm was started in `<path>` instead of the current working directory. 

Thus, `package/cli` has to be the path. Am I right? 🤔 

Adding the path to the `tsconfig.json` is only plausible if the package is imported somewhere else. Just like `nuxt` for example it doesn't need to be added imho.

userquin:
> 
> 
> I'm running from root. [The docs](https://pnpm.io/pnpm-cli#-c-path---dir-path) state for using `-C`:
> 
> > Run as if pnpm was started in `<path>` instead of the current working directory.
> 
> Thus, `package/cli` has to be the path. Am I right? 🤔
> 
> Adding the path to the `tsconfig.json` is only plausible if the package is imported somewhere else. Just like `nuxt` for example it doesn't need to be added imho.

ok, it should be `npm -C...`, but running it with `pnpm` it is ok (sorry man)

johannschopplich:
@antfu I messed up the PR while rebasing on `main`. Sorry about that. 😑

PR #107, opened 2021-11-13
fix: add explicit color types

ydcjeff:


antfu:
I don't think it's good to type them as users might extend their own color names.

ydcjeff:
But, currently accessing `colors.red` gives me type error in ts.

antfu:
What's your usage?

PR #117, opened 2021-11-17
fix(preset-uno): fix zIndex can't set negative number and 0

chris-zhu:
I find some problems. when i use default ruels about `zIndex`,  i can't use `z-0` and `-z-1`, and i think this is a bug. but now i resolve it.

### DIFF
--- a/packages/preset-uno/src/utils/handlers/shorthand.ts
+++ b/packages/preset-uno/src/utils/handlers/shorthand.ts
@@ -19,7 +19,7 @@ const handler = function(
   this.__options.sequence = []
   for (const n of s) {
     const res = handlers[n](str)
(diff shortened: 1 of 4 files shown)


chris-zhu:
image
this is unreasonable

packages/preset-uno/src/utils/handlers/shorthand.ts
```diff
@@ -19,7 +19,7 @@ const handler = function(
   this.__options.sequence = []
   for (const n of s) {
     const res = handlers[n](str)
-    if (res)
+    if (res !== undefined)
```
antfu, line 22:
```suggestion
    if (res != null)
```

PR #118, opened 2021-11-18
fix(line-clamp): fix line-clamp rely on break-all

chris-zhu:
before, we complete `line-clamp-${num}` rely on `break-all`
image

now.
image

i think  `line-clamp-${num}`  has a strong purpose and should carry break-all itself.


antfu:
Tailwind's line clamp does not include it https://github.com/tailwindlabs/tailwindcss-line-clamp I guess we better align with it?

PR #126, opened 2021-11-19
feat(preset-uno): width height theme size

hannoeru:
#125﻿


antfu:
Maybe we could have more entries in the test?

PR #145, opened 2021-11-19
feat(preset-uno): add container

### LINKED ISSUE #133: `container` missing rule: breakpoint variant `max-width` variant

userquin:
closes #133


userquin:
upps, it should be `max-with`, changing it...

hannoeru:
I think container is more like a preset of `max-width` in different viewport?

userquin:
@hannoeru It is here: https://tailwindcss.com/docs/container, we can move to here layout module.

userquin:
Should I change to?:
- `layouts.ts`  renamed to `overflow.ts`
- `container.ts` with the 2 rules here.

userquin:
> I think `container` is more like a preset of `max-width` in different viewport?
> 

yeah, sorry, testing...

userquin:
I keep `container` rule on `size.ts` module.

test/__snapshots__/preset-uno.test.ts.snap
```diff
@@ -352,9 +353,14 @@ exports[`targets 1`] = `
 }
 @media (min-width: 640px){
 .sm\\\\:m-1,.sm\\\\:m1{margin:0.25rem;}
+.sm\\\\:container{width:100%;}
```
userquin, line 356: @hannoeru How can I avoid generating this style?

userquin: I mean,is  it correct?

userquin:
ok, it is wrong, updating again...

userquin:
@hannoeru  Sorry for the delay, I switch to my work, I'm back.

So the `container` rule will add `{ 'width': '100%'}` and all the corresponding `@media max-width` declared on `breakpoints`, but it also needs to be itselft responsive, that is, we need also add bp variants to `container`, for example `lg:container`. In that case we need to rewrite the styles using only `breakpoints` from the current and down in the current list (from current and up in size), following with the `lg:container` example, the resulting rules will contain only `lg`, `xl` and `2xl` styles for `@media max-width`.

Should only handle `:` or also support `-` variant?

userquin:
@hannoeru @antfu The problem with this rule is that is not a normal breakpoint rule, by default , it will configure the `max-width` for each `theme breakpoint`, but can also be used with `breakpoint` variant (from the `breakpoint` and up, see tailwind docs here https://tailwindcss.com/docs/container#responsive-variants).

The `container` rule will result on this:
```css
/* class="container" or container attr */ 
.container,[container=""]{width:100%}
@media (min-width:640px) {.container,[container=""]{max-width:640px}}
@media (min-width:768px) {.container,[container=""]{max-width:768px}}
@media (min-width:1024px) {.container,[container=""]{max-width:1024px}}
@media (min-width:1280px) {.container,[container=""]{max-width:1280px}}
@media (min-width:1536px) {.container,[container=""]{max-width:1536px}}
/* class="md:container" or md:container attr */ 
@media (min-width:1024px) {.lg\:container,[lg\:container=""]{width:100%;max-width:1024px}}
@media (min-width:1280px) {.lg\:container,[lg\:container=""]{max-width:1280px}}
@media (min-width:1536px) {.lg\:container,[lg\:container=""]{max-width:1536px}}
```

I have changed `breakpoint` variant to exclude `container` to be processed on `variantBreakpoints`, then added its own module on `rules` and handle the `breakpoint` variant on the new module. It also includes a cache, but we can remove it.

I don't know the implications aboout reusing the `@media (min-width...) {}`: I cannot use Variant and cannot allow breakpoints variant to process it, so the only way I have found is to exclude from latter.

userquin:
we should also review if we need to add the center and padding options: https://tailwindcss.com/docs/container#customizing

hannoeru (commented):
Missing `;` in end of css property?

```css
.container,[container=\\"\\"]{width:100%}
```
```css
.container,[container=\\"\\"]{width:100%;}
```

userquin:
> Missing `;` in end of css property?
> 
> ```css
> .container,[container=\\"\\"]{width:100%}
> ```
> 
> ```css
> .container,[container=\\"\\"]{width:100%;}
> ```

I don't add it, see for example `animate` for single or last entries.

packages/preset-uno/src/rules/container.ts
```diff
@@ -0,0 +1,31 @@
...
+      if (idx !== -1) {
+        return keys.slice(idx).map((bp, idx) => {
+          return `@media (min-width:${bps[bp]}) {.${useBp}\\:container,[${useBp}\\:container=""]{${idx === 0 ? 'width:100%;' : ''}max-width:${bps[bp]}}}`
+        }).join('\n')
+      }
+      return undefined
+    }
+
+    const entries = keys.map(bp => `@media (min-width:${bps[bp]}) {.container,[container=""]{max-width:${bps[bp]}}}`)
+    return `.container,[container=""]{width:100%}\n${entries.join('\n')}`
```
antfu, line 29: Can we use `constructCustomCSS` here? So we could apply variants on it

userquin: also on variant?

userquin: I think it is not a good idea, read https://github.com/antfu/unocss/pull/145#issuecomment-974489948

antfu: ok, let me handle it then

PR #153, opened 2021-11-20
feat(preset-uno): add missing position sticky and static

userquin:
thx to @chris-zhu


userquin:
the vendor prefix not included since it has a 96% of global coverage: https://caniuse.com/css-sticky

PR #179, opened 2021-11-29
feat: Support font family customization and font compression configuration

edwardnyc:
This is a proposal, I think of an idea

input:
```
src:https://tx2.a.kwimgs.com/udata/pkg/AUTO-KUAISHOU-COM-STATIC-FONT/DINCondensed.ttf
dest: '/build/fonts'
family: 'din1451'
text: '新年快乐'
```

usage
```<div class=" font-din1451"> 快乐 </div>```

If you agree with me, I will continue @antfu @userquin

### DIFF
--- a/packages/preset-font/package.json
+++ b/packages/preset-font/package.json
@@ -1,0 +1,57 @@
+{
+  "name": "@unocss/preset-font",
+  "version": "0.12.13",
+  "description": "Pure CSS font for UnoCSS",
+  "keywords": [
+    "unocss",
+    "unocss-preset",
+    "font",
+    "css-font",
+    "iconify"
+  ],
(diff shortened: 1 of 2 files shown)


edwardnyc:
Our project is now used under the webpack plugin and has gained a good experience. Fontmin this NPM package does not support ESM, so I intend to call it by cp.exec

The picture below is the result of WordArt compression

image

antfu:
Can you describe a bit what you are trying to achieve?

edwardnyc:
First of all, we don't have the ability to support `@font-face` statement at present,

It's easy for us to implement `@font-face`. However, in general, the size of `WordArt` is very large, so developers will compress it on demand. I think it would be a great experience if we could implement the introduction and compression through a few lines of configuration.


Simply put, the goal is to support the configuration of `WordArt` and extract the text of the font package on demand

PR #182, opened 2021-11-30
Extract regexes

chu121su12:
Extract regexes inside callback to external constant


antfu:
I found it's hard to read after this refactoring, thanks for your contributions but I don't think we want it.

PR #183, opened 2021-11-30
feat(preset-uno + preset-icons): extract and update regexes

chu121su12:
Compared to #182, this PR targets part of the code base that requires some logic update. To note:
- `encodeSvg()`: extract to array and use reduce.
- `time()`: changed the second regex (`/ms|s$/`) to match the first one (`/(s|ms)$/`)

### DIFF
--- a/packages/preset-icons/src/utils.ts
+++ b/packages/preset-icons/src/utils.ts
@@ -1,6 +1,6 @@
 export const isNode = typeof process < 'u' && typeof process.stdout < 'u'
 
-const svgCharRe = [
+const svgCharRe: [RegExp, string][] = [
     [/"/g, '\''],
     [/%/g, '%25'],
     [/#/g, '%23'],
@@ -9,6 +9,7 @@ const svgCharRe = [
     [/</g, '%3C'],
     [/>/g, '%3E'],
 ]
+
 // https://bl.ocks.org/jennyknuth/222825e315d45a738ed9d6e04c7a88d0
 export function encodeSvg(svg: string) {
   return svgCharRe.reduce(
--- a/packages/preset-icons/src/utils.ts
+++ b/packages/preset-icons/src/utils.ts
@@ -1,19 +1,13 @@
 export const isNode = typeof process < 'u' && typeof process.stdout < 'u'
 
-const svgCharRe: [RegExp, string][] = [
-    [/"/g, '\''],
-    [/%/g, '%25'],
-    [/#/g, '%23'],
-    [/{/g, '%7B'],
-    [/}/g, '%7D'],
-    [/</g, '%3C'],
-    [/>/g, '%3E'],
-]
-
 // https://bl.ocks.org/jennyknuth/222825e315d45a738ed9d6e04c7a88d0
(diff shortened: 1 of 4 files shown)


packages/preset-icons/src/utils.ts
```diff
@@ -1,13 +1,19 @@
...
-    .replace(/%/g, '%25')
-    .replace(/#/g, '%23')
-    .replace(/{/g, '%7B')
-    .replace(/}/g, '%7D')
-    .replace(/</g, '%3C')
-    .replace(/>/g, '%3E')
+  return svgCharRe.reduce(
+    (str, [regexp, replacement]) => str.replace(regexp, replacement),
+    svg.replace('<svg', (~svg.indexOf('xmlns') ? '<svg' : '<svg xmlns="http://www.w3.org/2000/svg"'))
+  )
```
antfu, line 18: Is this way faster? If not, I would prefer the old one.

packages/preset-uno/src/utils/handlers/handlers.ts
```diff
@@ -67,13 +67,14 @@ export function cssvar(str: string) {
     return `var(--${str.slice(1)})`
 }
 
+const timeRe = /(s|ms)$/
```
antfu, line 70:
```suggestion
const timeRe = /(ms|s)$/
```

antfu (changes requested):

packages/preset-icons/src/utils.ts
userquin: avoid the replacement when the svg xml namespace present?

packages/preset-uno/src/rules/behaviors.ts
```diff
@@ -49,10 +49,13 @@ export const outline: Rule[] = [
...
+const listStyleRe1 = /-inside|-outside/
+const listStyleRe2 = /inside|outside/
+
 export const listStyle: Rule[] = [
-  [new RegExp(`^list-((${listStyleProps.join('|')})(?:(-outside|-inside))?)$`), ([, value]) => {
-    const style = value.split(/-outside|-inside/)[0]
-    const position = /inside|outside/.exec(value) ?? []
+  [new RegExp(`^list-((${listStyleProps.join('|')})(?:(-inside|-outside))?)$`), ([, value]) => {
+    const style = value.split(listStyleRe1)[0]
+    const position = listStyleRe2.exec(value) ?? []
```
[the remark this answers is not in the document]
chu121su12: This part only swaps inside|outside to match the surrounding style.

packages/preset-uno/src/utils/handlers/handlers.ts
chu121su12: Updated

packages/preset-icons/src/utils.ts
chu121su12: Reverted since I couldn't find noticeable speed difference

antfu:
I found it's hard to read after this refactoring, thanks for your contributions but I don't think we want it.

PR #184, opened 2021-11-30
feat(preset-uno): add peer variant support

chu121su12:
I want to add `.peer` support (see [tailwind's doc](https://tailwindcss.com/docs/just-in-time-mode#sibling-selector-variants) and [source](https://github.com/tailwindlabs/tailwindcss/blob/v3.0.0-alpha.2/src/corePlugins.js#L108)). However I couldn't figure out how to add the test, so the code is only based on copy-paste from the `Group` section.


chu121su12:
Found the test section thanks to failing test.

PR #185, opened 2021-11-30
Support content utility

chu121su12:
Source: https://tailwindcss.com/docs/just-in-time-mode#content-utilities


antfu:
Hmmm, I am actually reluctant to support this https://github.com/antfu/unocss/issues/124#issuecomment-981532905

PR #196, opened 2021-12-01
fix: improve selector regex

owlnai:
I was trying to add a rule that supported emoji selectors, but to my surprise unoCSS didn't parse it although native CSS does (e.g `[🙈] { color: red }`). I checked and the attribute RegExp seems to only support alphanumeric characters, colons, % and hyphens. 

[According to the W3C spec](https://www.w3.org/TR/CSS21/syndata.html#value-def-identifier), the syntax for identifiers is quite broader so I modified it according to the spec. I haven't been able to test it outside RegExp unit tests, but I believe it shouldn't break anything.


antfu:
Can you also add some tests? Thanks

owlnai:
Tests added 👍

Issue #195, opened 2021-12-01
Support for tailwind Font Variant Numeric


chu121su12:
Docs: https://tailwindcss.com/docs/font-variant-numeric

I tried to add the rules but it seems that the generator is prioritizing static rules and only returning the first rule that matches.

Is there planned update for the core to match more than 1 rule?

chu121su12:
I am implementing the `font-variant-numeric` rules, and using tailwind as a reference, there's this section:

*(other modifiers are trimmed for brevity)*
```css
/* base */
.ordinal, .slashed-zero /*, ... */ {
  --tw-ordinal: var(--tw-empty,/*!*/ /*!*/);
  --tw-slashed-zero: var(--tw-empty,/*!*/ /*!*/);
  /* ... */
  font-variant-numeric: var(--tw-ordinal) var(--tw-slashed-zero) /* ... */;
}
/* modifiers */
.normal-nums { font-variant-numeric: normal; }
.ordinal { --tw-ordinal: ordinal; }
/* ... */
```

With the intention to copy this behaviour, I tried adding these rules:
```ts
const fontVariantNumericBase = {
  '--un-ordinal': varEmpty,
  '--un-slashed-zero': varEmpty,
  // ...
  'font-variant-numeric': 'var(--un-ordinal) var(--un-slashed-zero)', //...
}
const fontVariantNumericV1: Rule[] = [
  [/^(?:ordinal|slashed-zero|___)$/, () => fontVariantNumericBase],
  ['normal-nums', { 'font-variant-numeric': 'normal' }],
  ['ordinal', { '--tw-ordinal': 'ordinal' }],
  ['slashed-zero', { '--tw-slashed-zero': 'slashed-zero' }],
  // ...
]
```

The generator prioritizing string rules only produced the css in `modifiers` section. By allowing multiple matches, this can be achieved with the V1 rules.

Although, while writing this comment, I remember you can just use this V2 rules:
```ts
const handleFontVariantNumeric = (style: object) => ({
  ...fontVariantNumericBase,
  ...style,
})
const fontVariantNumericV2: Rule[] = [
  ['ordinal', handleFontVariantNumeric({ '--tw-ordinal': 'ordinal' })],
  ['slashed-zero', handleFontVariantNumeric({ '--tw-slashed-zero': 'slashed-zero' })],
  // ...
  // 'normal-nums' moved to bottom as this should override the `base` css
  ['normal-nums', { 'font-variant-numeric': 'normal' }], 
]
```
[...]

PR #197, opened 2021-12-01
feat(core): multiple rule match

chu121su12:
Proposed implementation to support [#195](https://github.com/antfu/unocss/issues/195#issuecomment-983671511).


chu121su12:
Weird that locally the test is passing. Updating the snapshot manually to follow the test reverses the test result: it wants the duplicated rule removed.

I'm on windows 11 + node 16. I will try later in the ubuntu box.

PR #198, opened 2021-12-01
feat: allow a rule to return multiple css items, close #197

antfu:
Another solution of #197, /cc @chu121su12

Instead of allowing match multiple rules (which will make the overriding impossible and hard to resolve when there are conflicts), this PR allow rules to return an array of css object to multiple css selectors.

```ts
rules: [
  [/^scale-(\d+)$/, ([, d]) => {
    return [
      { transform: 'var(--scale)' },
      { --scale: d }
    ]
  }
]
```

Which make it generates 

```css
.scale-100 {
  transform: var(--scale);
}
.scale-100 {
  --scale: 100
}
```

And [Rules Merging](https://github.com/antfu/unocss#rules-merging) could merge the duplicated parts.


chu121su12:
I missed on that rules can return array

PR #210, opened 2021-12-03
fix: build script on windows environments

userquin:
Copying `README.md` on `unocss` and `LICENSE` on `vscode` breaks windows build.


userquin:
upps, `LICENSE` not being copied.

johannschopplich:
@userquin What about using [shx](https://github.com/shelljs/shx)? That way the shell scripts can stay readable and no extra scripts would have to be introduced.

Example usage: `shx cp ../../README.md .`

packages/unocss/scripts/copy-files.ts
```diff
@@ -0,0 +1,11 @@
+import { copyFileSync } from 'fs'
+import { resolve } from 'path'
+
+// relative to scripts directory
```
antfu, line 4: Could we move this to the top level so we copy them once on build

userquin: sure, to root script directory

userquin:
@johannschopplich we should this way to allow copying multiple files at once, and we can reuse the module for ohter things when required

PR #214, opened 2021-12-04
feat(preset-mini): transform rules align with tw

chris-zhu:


packages/preset-mini/src/rules/transform.ts
```diff
@@ -12,24 +22,20 @@ const transformBase = {
...
-  transform: 'rotate(var(--un-rotate)) scaleX(var(--un-scale-x)) scaleY(var(--un-scale-y)) scaleZ(var(--un-scale-z)) skewX(var(--un-skew-x)) skewY(var(--un-skew-y)) translate3d(var(--un-translate-x), var(--un-translate-y), var(--un-translate-z))',
-  [CONTROL_BYPASS_PSEUDO]: '',
+  ...transformCpu,
 }
 
 export const transforms: Rule[] = [
   ['transform', transformBase],
+  ['transform-gpu', transformGpu],
+  ['transform-cpu', transformCpu],
+  ['transform-none', { transform: 'none' }],
```
antfu, line 32: I guess they should be on the end of the list to overrides the base rules provided by each utils

chris-zhu: Ok, I have corrected my mistake

PR #221, opened 2021-12-05
Revert "feat(preset-mini): new text-[sizes] behaviour"

chu121su12:
Reverts antfu/unocss#218

~I messed up rule order with the other branch.~ Will redo this only in wind next time.

https://github.com/antfu/unocss/blob/2d7d1b029e276ac0a420605d75bc41a3ddbdfc4f/test/__snapshots__/preset-uno.test.ts.snap#L121-L122


antfu:
Why reverting? I think this change makes sense tho

chu121su12:
My bad. I wrongly assummed the rule order.

PR #222, opened 2021-12-05
feat(vite): add `svelte/sveltekit` and `shadow-dom` support

userquin:
This PR adds support for:
- [x] `vite + svelte`: add support for `class:foo={bar}` and `class:foo` (shorthand)
- [x] `sveltekit`: I need to add test for `sveltekit` and test it
- [x] `shadow-dom`: add support for `webcomponents`, I have it working but on build it fails (I need to fix it)
- [x] `shadow-dom test`: add example on `test/fixtures` for `lit`

/cc @dominikg


userquin:
I will close this PR to split them into 2 separated PR, the code on `feat/per-module-import` is working as a POC.

**keep it open, i don't want the branch removed, we are working on it to add some svelte enhancements**

userquin:
Here a video with `lit` playground in action: https://imgur.com/LVuknAd

PR #225, opened 2021-12-06
feat(preset-mini): support text-[types:x]

### LINKED ISSUE #202: JIT ambiguous values support
https://tailwindcss.com/docs/just-in-time-mode#arbitrary-value-support
(sorry, no direct anchor link in the docs)

As per docs
```html
<div class="text-[var(--mystery-var)]">
```
vs
```html
<div class="text-[color:var(--mystery-var)]">
```

> The supported types are:
    - \`length\`
    - \`color\`
    - \`angle\`
    - \`list\`

/cc @userquin #124

chu121su12:
Closes #202; ~introduces #226~.

### DIFF
--- a/test/__snapshots__/preset-mini.test.ts.snap
+++ b/test/__snapshots__/preset-mini.test.ts.snap
@@ -206,8 +206,11 @@ div:hover .group-\\\\[div\\\\:hover\\\\]-\\\\[combinator\\\\:test-4\\\\]{combina
 .peer:is(:placeholder-shown)~.peer-is-placeholder-shown\\\\:text-3xl,
 .peer:not(:placeholder-shown)~.peer-not-placeholder-shown\\\\:text-3xl{font-size:1.875rem;line-height:2.25rem;}
 .peer:not(:placeholder-shown)~.peer-not-placeholder-shown\\\\:text-2xl{font-size:1.5rem;line-height:2rem;}
+.text-\\\\[length\\\\:2em\\\\],
 .text-2em,
 .text-size-\\\\[2em\\\\]{font-size:2em;}
+.text-\\\\[length\\\\:calc\\\\(1em-1px\\\\)\\\\]{font-size:calc(1em - 1px);}
+.text-\\\\[length\\\\:var\\\\(--size\\\\)\\\\]{font-size:var(--size);}
 .text-base{font-size:1rem;line-height:1.5rem;}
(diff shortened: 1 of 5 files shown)


antfu:
Would we find a way to reuse tho modules instead of duplicating them? Like a factory to generate the `h` helper (and it could be a separate package if needed)

antfu:
@chu121su12 Please take a look at the changes in #234. Thanks!

chu121su12:
Tests still fails without https://github.com/antfu/unocss/pull/225/commits/e7cf4dfe4825dbd3b3fb63db09fd43417ca09601

antfu (approved):

johannschopplich (approved):

userquin (approved):

chu121su12:
The logic is now added to preset-mini instead because maintaining duplicate parseColor functions is less than ideal, especially it may be added to other utility with different color parsers (bg, shadow, etc.). However with the existing mini syntaxes (x-size-y) this feature maybe (?) less than useful anymore except for tailwind compatibility.

test/__snapshots__/preset-mini.test.ts.snap
```diff
@@ -206,8 +206,11 @@ div:hover .group-\\\\[div\\\\:hover\\\\]-\\\\[combinator\\\\:test-4\\\\]{combina
 .peer:is(:placeholder-shown)~.peer-is-placeholder-shown\\\\:text-3xl,
 .peer:not(:placeholder-shown)~.peer-not-placeholder-shown\\\\:text-3xl{font-size:1.875rem;line-height:2.25rem;}
 .peer:not(:placeholder-shown)~.peer-not-placeholder-shown\\\\:text-2xl{font-size:1.5rem;line-height:2rem;}
+.text-\\\\[length\\\\:2em\\\\],
 .text-2em,
 .text-size-\\\\[2em\\\\]{font-size:2em;}
+.text-\\\\[length\\\\:calc\\\\(1em-1px\\\\)\\\\]{font-size:calc(1em - 1px);}
+.text-\\\\[length\\\\:var\\\\(--size\\\\)\\\\]{font-size:var(--size);}
```
antfu, line 213: Why is it called length instead of `size`?

chu121su12: Just following tailwinds'
https://tailwindcss.com/docs/adding-custom-styles#resolving-ambiguities

PR #243, opened 2021-12-09
feat(preset-wind): add `tailwind` missing animations

### LINKED ISSUE #241: animate-pulse is scaling the object
The pulse keyframes are wrong and do not align with windicss or tailwind.
Instead of pulsing opacity it pulses scale.

userquin:
Added missing tailwind `pulse` and `bounce`, `tw-pulse` and `tw-bounce` respectively, `spin` and `ping` was already there.

closes #241

### DIFF
--- a/packages/preset-wind/src/rules/animation.ts
+++ b/packages/preset-wind/src/rules/animation.ts
@@ -107,6 +107,7 @@ const keyframes: Record<string, string> = {
 }
 
 const durations: Record<string, string> = {
+  'tw-pulse': '2s',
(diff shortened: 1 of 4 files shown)


antfu:
Maybe we should align the unprefixed one with Tailwind and have the one from animated.css as something like `pulse-alt`?

userquin:
ok

userquin:
updated also playground to use `animate-bounce-alt`

userquin:
@chris-zhu can you take a look to the changes I've made?

PR #248, opened 2021-12-10
fix(preset-uno/preset-wind): send `dark` option through `presetMini`

chu121su12:
https://github.com/antfu/unocss/blob/22eb37fb214fa05b4e85c65416585d6ca49a383e/test/preset-uno.test.ts#L423-L427


chu121su12:
This is obsoleted with #249 by not depending on `presetMini()`

Issue #124, opened 2021-11-18
review: align with tailwind


userquin:
I have tested main entries from here https://tailwindcomponents.com/ except Dashboard and Account Card entries.

I will be adding rules here that are not on unocss preset, or that are misaligned with tailwind:
- [x] #125
- [x] #134
- [x] #128
- [x] #133
- [x] #130
- [x] #170
- [x] #132 
- [x] #180
- [x] #178
- [x] #136
- [x] #167

userquin:
@antfu @hannoeru @edwardnyc

On weekend I'll do a second review.

userquin:
Ofc Man, I added the task for it... you can also asign yourself the issue

userquin:
@hannoeru  I assign it to you, see assignees from right panel on the issue, we need to check if the title can be changed or modify in some way, ot is huge

xiaojieajie:
Hello, I want to help solve the seventh task（ object-center missing rule: object-position: center;, we should add the variants (top, right...)）

hannoeru:
@xiaojieajie Click convert to issue button on right side of list to create issue and assign yourself, then you are good to go!

xiaojieajie:
> @xiaojieajie单击列表右侧的转换为问题按钮以创建问题并分配给自己，然后就可以开始了！


Ask me, I created issue, but I don't think the task can be selected?

chris-zhu:
Can i handle the last one?

hannoeru:
@xiaojieajie Task will auto checked when you close that issue.

hannoeru:
> Can i handle the last one?

Of course.

userquin:
wtf: I created the issue this very morning, I just woke up and almost all the tasks started / closed, you are all amazing.

Thank you very much to all.

UltraCakeBakery:
Can you add `translate-x-full`, `translate-y-full` and their negative variants to the list? @userquin

SIGUSR97:
[...]

PR #253, opened 2021-12-10
feat(preset-wind): update rule order

chu121su12:
This PR takes off from #249 and tries to match the ordering from tailwind (v3).

Due to the large number of rules shuffled, I verified uno and attributify snapshot by only by comparing the sorted lines. I hope this is enough.

There are also commented lines in `rules` for unimplemented features.
/cc @userquin #124


antfu:
I did some refactoring to the presets, you can help resolve the conflicts? Thanks

PR #260, opened 2021-12-11
fix(preset-mini): `question-mark` rule should use `userConfig.envMode`

### LINKED ISSUE #258: UnoCSS Renders `?` Key CSS When Using Ternary
I'm using UnoCSS with Vue 3 Script Setup Sugar but even though I'm not using `?` key, UnoCSS and UnoCSS VSCode plugin detects ternary's question mark as a key and renders it. The presets I use are `presetUno` and `presetIcons`. I don't use `attributify` mode.
## My Code (<script setup>)
image
## Built CSS (formatted)
image

userquin:
closes #258

### DIFF
--- a/packages/plugins-common/context.ts
+++ b/packages/plugins-common/context.ts
@@ -42,7 +42,7 @@ export function createContext<Config extends UserConfig = UserConfig>(
 
     rawConfig = result.config
     uno.setConfig(rawConfig)
-    uno.config.envMode = 'dev'
+    uno.config.envMode = rawConfig.envMode ?? 'build'
     rollupFilter = createFilter(
       rawConfig.include || defaultInclude,
       rawConfig.exclude || defaultExclude,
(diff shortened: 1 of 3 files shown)


antfu:
Not so sure. I expect reload only happens on dev?

userquin:
> Not so sure. I expect reload only happens on dev?

I just build the `vite-svelte` and the `?` is there, on hmr is not necessary, but it seems the line I modified is being called on `build`.

userquin:
it is being called here:
```ts
  async function getConfig() {
    await ready
    return rawConfig
  }
```

the `await ready` statement is just `const ready = reloadConfig()`

userquin:
@antfu called here https://github.com/antfu/unocss/blob/main/packages/vite/src/modes/global/build.ts#L52?

userquin:
@antfu maybe we can access the `userConfig` on the rule, changing it...

PR #266, opened 2021-12-13
feat(preset-attributify): support for peer & group

### LINKED ISSUE #229: Attributify: attribute: `group`'s selector doesn't match
The generated selector for the `group` attribute does not match. 

Input:
```html
<div group>
      <div opacity-0 group-hover:opacity-100
</div>
```

Output:
```css
.group:hover [group-hover~="opacity-100"],.opacity-100{opacity:1;}
```

chu121su12:
Closes #229, but I may need help on verifying how `variantAttributify` works.


chu121su12:
Forgive me I'm bad with npm and build process.

UltraCakeBakery:
Thanks for looking in to #229 <3

PR #268, opened 2021-12-14
feat(preset-attibutify): support for peer & group

### LINKED ISSUE #229: Attributify: attribute: `group`'s selector doesn't match

chu121su12:
Closes #229, but I may need help on verifying how `variantAttributify` works.

Redoing this since previous PR has too many unused commit.


packages/preset-attributify/src/variants/pseudo.ts
```diff
@@ -0,0 +1,37 @@
+import type { CSSEntries, VariantObject } from '@unocss/core'
+import { CONTROL_BYPASS_PSEUDO_CLASS, PseudoClasses } from '@unocss/preset-mini/variants'
```
antfu, line 2: I kinda want to avoid relying preset-mini on attributify (as it expects to empower any rules to attributify). I think it might be better to have this logic somehow in the preset-mini instead.

chu121su12: Something like this? https://github.com/antfu/unocss/pull/268/commits/2ddc566d99b9d63f3c64210d44754a7c7b449cd9

chu121su12: Getting ahead of myself, the spec says `:not`, `:is` and `:where` have similar syntax, so you can generalize NotRE as such:
https://github.com/chu121su12/unocss/compare/preset-attibutify-variant-2...chu121su12:pseudo-class-fn

https://developer.mozilla.org/en-US/docs/Web/CSS/:not#syntax
https://developer.mozilla.org/en-US/docs/Web/CSS/:where#syntax
https://developer.mozilla.org/en-US/docs/Web/CSS/:is#syntax

`:has` can be added too, but there's no browser support yet
https://developer.mozilla.org/en-US/docs/Web/CSS/:has#syntax

packages/preset-attributify/package.json
```diff
@@ -37,6 +37,7 @@
     "stub": "unbuild --stub"
   },
   "dependencies": {
-    "@unocss/core": "workspace:*"
+    "@unocss/core": "workspace:*",
+    "@unocss/preset-mini": "workspace:*"
```
antfu, line 41: I still want these two preset to be orthogonal.

test/__snapshots__/preset-attributify.test.ts.snap
```diff
@@ -102,22 +135,44 @@ exports[`attributify > fixture1 1`] = `
...
-[text-4xl=\\"\\"]{font-size:2.25rem;line-height:2.5rem;}
+[peer=\\"\\"]:checked~[peer-checked~=\\"text-4xl\\"],[text-4xl=\\"\\"]{font-size:2.25rem;line-height:2.5rem;}
 [text-5xl=\\"\\"]{font-size:3rem;line-height:1;}
 [font-100=\\"\\"]{font-weight:100;}
 [font-200=\\"\\"]{font-weight:200;}
+[leading-1rem=\\"\\"]{line-height:1rem;}
 [tracking-wider=\\"\\"]{letter-spacing:0.05em;}
-.group:hover [group-hover~=\\"text-teal-400\\"]{--un-text-opacity:1;color:rgba(45,212,191,var(--un-text-opacity));}
-.group:hover [group-hover~=\\"op-50\\"]{opacity:0.5;}
+[group=\\"\\"]:hover [group-hover~=\\"text-teal-400\\"]{--un-text-opacity:1;color:rgba(45,212,191,var(--un-text-opacity));}
```
antfu, line 168: Even when attributify is enabled, I think matching the class version of `.group` and `.peer` is till needed as people might mix them together

test/preset-wind-targets.ts
```diff
@@ -1,12 +1,86 @@
...
-  'object-ct',
-  'object-cb',
+  // animation
+  'animate-none',
+  'animate-pulse',
+  'animate-pulse-alt',
+  'animate-300',
+  'animate-100s',
+  'animate-duration-100',
+  'animate-duration-100.32',
```
antfu, line 9: Could you have the refactoring in another PR so we can have a clearer state? Thanks

PR #271, opened 2021-12-15
feat(preset-mini): remove `next:` variant (breaking)

chu121su12:
May be reused for tagged variant like group/peer


antfu:
What's the reason of this?

chu121su12:
Originally to change the behaviour to be similar to peer, but with `+` combinator instead of `~`. There's also alternative adding `peers` for `~` and changing `peer` to use `+` instead, but this will be incompatible with tailwind's syntax.

antfu:
I see no harm to leave it for a while, we could remove or change it when there is more reasonable `next:`

PR #275, opened 2021-12-17
feat(preset-mini): ease-linear

xiaojieajie:
When I was using Transition, I found that LINEAR was missing, which affected my use

I went to check tailwindcss

WeChate55788ab556dcce6f1842688515222e2

WeChat7235e1f74bfb27b83008ed5fb646200b

So I'm going to fill it in！

WeChat53bdd8d822cf3edbf891b980654ac4e1

WeChata45d1293a9d8b25296c1d16ee9b9daac


packages/preset-mini/src/rules/transition.ts
```diff
@@ -28,6 +28,7 @@ export const transitions: Rule[] = [
       'transition-duration': `${duration}ms`,
     }
   }],
+  ['ease-linear', { 'transition-timing-function': 'linear' }],
   ['ease', { 'transition-timing-function': 'cubic-bezier(0.4, 0, 0.2, 1)' }],
```
antfu, line 32: Better to be after `ease`

xiaojieajie: ok

PR #278, opened 2021-12-18
refactor(preset-mini,preset-wind): split and prioritize `x-none`, `x-normal`

chu121su12:
This PR reduces runtime file size by using rule generator functions. The sizes are:
- mini: 47.25kb -> 45.38kb
- uno: 82.22kb -> 75.84kb
- attr: 83.4kb -> 77.03kb

Breaking changes:
- Removed `m` regex flag
- fix `self-stretch` definition
- `x-none` rules get split and prioritized
- some rule order within its group are modified

Since this is a long PR, below contains some summary. Most of the commit are done on preset mini and wind, so they usually go by turns. Every commit was linted & tested. (Last commits in the main branch forces alpha-ordered import, so the lint may not pass anymore)

First commits is to reorder test section to match to the `.ts` filename order in the folder. This makes easier tracking of the classes vs rules. The next commits moves the classes available in uno into each of the mini/wind preset according to the sections. Last commits in the test section add the missing classes that can be matched.

Next business is to refactor some utility functions. First is to move the `colorResolver` and `parseColorUtil` into `utilities.ts`. Second is to split `auto` handler into its own function.

Next 2 commits refactor the rules in mini & wind. The details are in the commit message, but mostly small tweaks here and there. This first 2 commits preserve the tests, except for the `varEmpty` and `self-stretch` fix.

Next 2 commits then split the `x-none` and some `x-normal` classes then prioritize them by placing the rules below their original definition. Tests are updated to reflect this changes.

Before using the generator functions, I separate the commit for the definitions. Usage of the generator functions are again split into 2 commits for each mini and wind presets.

Most affected rules are colors & opacity that can be defined with `createColorAndOpacityRulePair`, `createColorRule` or `createColorOpacityRule`. This function should eliminate discrepancies in color rule definition by placing them in single place.
Side note: if the `m` flag in opacity is needed, it can be re-added globally in the utility.

Next type of rules that are modified a lot are prefixed static rules by replacing them with `createKeywordRules` or the alternative `createGlobalKeywordRules` if the original rule has `inherit`, `initial`, `revert` and `unset` in the definition. This helpers generate array of static rules, so order of the specified parameters matter. This can be changed to be not strict by updating the helper definition to output regex rule.

These last 3 commits only reorder the test output,

Benchmark classes includes the test classes from mini & wind tests. Benchmark result doesn't show noticeable difference:

This PR (before https://github.com/antfu/unocss/commit/f99b0cfd082c59779da5f081e101a105c2c837f3):
```
12/18/2021, 12:41:58 PM
2216 utilities | x50 runs (min build time)

none                             15.96 ms / delta.      0.00 ms 
unocss       v0.16.4             34.54 ms / delta.     18.58 ms (x1.00)
```

commit https://github.com/antfu/unocss/commit/54a84b6330eda39e90c1da1278eb46c6fe71bef3 :
```
12/18/2021, 12:43:56 PM
2216 utilities | x50 runs (min build time)

none                             15.78 ms / delta.      0.00 ms 
unocss       v0.16.4             35.36 ms / delta.     19.58 ms (x1.00)
```

I hope this is well received. Feedback are welcome.

---

Update:
Turned out not much can be shaved with good compression.

|| https://github.com/antfu/unocss/commit/54a84b6330eda39e90c1da1278eb46c6fe71bef3 | this pr |
| - | - | - |
@unocss/preset-mini
gzip|12.90 KiB|12.80 KiB
brotli|11.34 KiB|11.29 KiB
unocss/preset-wind
gzip|8.43 KiB|7.88 KiB
brotli|7.23 KiB|6.83 KiB
@unocss/runtime
attributify.global.js
gzip|22.88 KiB|22.13 KiB
brotli|19.42 KiB|19.05 KiB
mini.global.js
gzip|15.91 KiB|15.76 KiB
brotli|13.87 KiB|13.76 KiB
uno.global.js
gzip|22.31 KiB|21.56 KiB
brotli|18.97 KiB|18.56 KiB


packages/preset-mini/src/utils/utilities.ts
```diff
@@ -7,7 +9,177 @@ export function capitalize<T extends string>(str: T) {
...
+export const globalKeywordResolver = (props: string | string[], keywords: CSSPropertyValue[] = []) => {
+  const cache = Object.fromEntries(keywords.map(toArray))
+  return ([, s]: string[]) => {
+    const v = s in cache ? (cache[s] ?? s) : h.global(s)
+    if (v !== undefined)
+      return generateCssObject(props, v)
+  }
+}
+
+export const createKeywordRules = (bases: string | string[], props: string | string[], keywords: CSSPropertyValue[] = []): Rule[] =>
```
antfu, line 162: Overall looks good, but the abstraction makes the code hard to read. Can you write some jsdocs for each utilities and provide some input / output examples to them so we could better understand how each functions doing their jobs? Thanks

antfu:
Also, can you do a comparison with `npm run size`? I think `gzip` and `brotli` could optimize duplicate strings

chu121su12:
Updated post with `npm run size` result.

antfu:
@chu121su12 Do you think it's worth this tradeoff? (readbility, maintainability, a bit runtime cost for slightly smaller file size)

chu121su12:
> Also, can you do a comparison with `npm run size`? I think `gzip` and `brotli` could optimize duplicate strings

Due to small compressed size update, my consideration:
- The commits up to https://github.com/antfu/unocss/pull/278/commits/d6a8db354b664fda757b325bb71ca0c00b3c9cce are recommended. It doesn't change the behaviour of the codebase and makes it easier to adjust the tests classes. 
- The commits up to https://github.com/antfu/unocss/pull/278/commits/4aabbb43b29b4e4c256495104b5553c45d04b21a are optional. I placed `-none` last so it can override the related keywords.
- The rest of the commits can be reverted. I'll do another attempt to extract simple regex rules into static rules.

antfu:
Up to 4aabbb4 sounds good to me!

antfu (approved):

chu121su12:
Thank you for the guidance

PR #281, opened 2021-12-18
feat: add `shadow-dom` support

### LINKED ISSUE #169: lit support
Hi @antfu 

Does our unocss supports lit web components. I have tried using unocss with lit but unocss is unable to update the styles.

Please guide me how can we use unocss with lit.

thanks.

userquin:
we need to add some documentation, the plugin just replace the `@unocss-import` with the generated css

supersedes #222 (`shadow-dom` part)

closes #169

### DIFF
--- a/packages/plugins-common/constants.ts
+++ b/packages/plugins-common/constants.ts
@@ -1,2 +1,3 @@
 export const INCLUDE_COMMENT = '@unocss-include'
 export const INCLUDE_COMMENT_IDE = '@unocss-ide-include'
+export const CSS_PLACEHOLDER = '@unocss-placeholder'
--- a/packages/plugins-common/constants.ts
+++ b/packages/plugins-common/constants.ts
@@ -1,2 +1,3 @@
+export const IMPORT_CSS = '@unocss-import'
 export const INCLUDE_COMMENT = '@unocss-include'
 export const INCLUDE_COMMENT_IDE = '@unocss-ide-include'
--- a/packages/plugins-common/constants.ts
+++ b/packages/plugins-common/constants.ts
@@ -1,3 +1,2 @@
-export const IMPORT_CSS = '@unocss-import'
 export const INCLUDE_COMMENT = '@unocss-include'
 export const INCLUDE_COMMENT_IDE = '@unocss-ide-include'
--- a/packages/vite/src/modes/shadow-dom.ts
+++ b/packages/vite/src/modes/shadow-dom.ts
@@ -64,10 +64,10 @@ export function ShadowDomModuleModePlugin({ uno }: UnocssPluginContext): Plugin 
 
         return acc
       }, new Array<PartData>())
-      const partsToApply = new Map<string, Array<string>>()
       if (useParts.length > 0) {
         let useCode = code
         let element: RegExpExecArray | null
+        const partsToApply = new Map<string, Array<string>>()
         const { elementIdxMap, idxResolver } = idxMapFactory()
         // We need to traverse the code to find the web components using the original class/attr part.
         // We need traverse the code to apply the same order the components are on the code: we are using nth-of-type.
--- a/packages/vite/src/modes/shadow-dom.ts
+++ b/packages/vite/src/modes/shadow-dom.ts
@@ -15,6 +15,7 @@ export function ShadowDomModuleModePlugin({ uno }: UnocssPluginContext): Plugin 
     let { css, matched } = await uno.generate(code, { preflights: false })
 
     if (css && matched) {
+      // check if we need to apply some ::part
       const partsToApply = Array.from(matched).reduce((acc, e) => {
         if (e.match(partRegex)) {
           const cssEntryRegex = `\\\.${e.replace(
@@ -39,7 +40,7 @@ export function ShadowDomModuleModePlugin({ uno }: UnocssPluginContext): Plugin 
 
         return acc
       }, new Map<string, string>())
-      if (partsToApply) {
+      if (partsToApply.size > 0) {
         css = Array.from(partsToApply.entries()).reduce((k, [r, name]) => {
           let regexp = cache.get(r)
           if (!regexp) {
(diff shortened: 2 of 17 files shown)


packages/vite/src/modes/shadow-dom.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function ShadowDomModuleModePlugin({ uno }: UnocssPluginContext): Plugin {
+  async function transformWebComponent(code: string) {
+    const imported = code.match(IMPORT_CSS)
+    if (!imported)
+      return code
+
+    const { css } = await uno.generate(code, { preflights: false })
+    if (!css)
+      return code
```
userquin, line 13: should also replace the placeholder with empty string

userquin:
we should also add an example with vanilla/typescrypt web component on index.html page for example.

@antfu how about support for `::part`, `::theme` and  `:explainer` here on unocss?

userquin:
We should review this and maybe change the plugin name, I Will explore some other approachs like finding style tag instead just replace the import comment.

I Will Mark it as draft.

userquin:
adding initial support for `::part`...

userquin:
The solution will work only with `vite`, and we need to do it in 2 phases: on pseudo and on the plugin (unless I'm missing somehting we can change internally, just putting here my thougths). Maybe we can apply another approach.

The problem with `::part` is that we cannot style shadow dom parts using css selectors, we need to use the custom element name as prefix for the `::part`, for example, this will work:
```css
my-another-element::part(cool-part) {
  background-color: #fefefe;
}
```
while this will not:
```css
.part-\[cool-part\]\:bg-yellow-300::part(cool-part){--un-bg-opacity:1;background-color:rgba(253,224,71,var(--un-bg-opacity));}
```

Another problem with the initial approach is when we have more than one style to apply: the previous css style was generated by this markup:
```html
export let MyElement = class extends LitElement {
  ...
  render() {
    return html`
      ...
      <my-another-element class="part-[cool-part]:bg-yellow-300">Testing css part</my-another-element>
      ...
    `;
  }
  ...
```

As you can see you can also add another parts on `my-another-element ` class, but then we have the problem to replace all pseudo parts with the component name on the plugin and we cannot do that (can be done, but the resulting code will be huge).

So, to avoid previous problem, I think the best way to support `::part` will be via `shorcuts` and so problem solved.

Finally, for `::part` we can also use it with pseudo elements like `:focus`, `:hover`, `::placeholder` (see last link bellow), but we can address it later (if we can do it).

For context:
- https://css-tricks.com/styling-a-web-component/
- https://meowni.ca/posts/part-theme-explainer/

@antfu @edwardnyc @chu121su12 @hannoeru I need some hints/ideas/review for this and sorry to bother you all, I'll push the initial support for `::part` in a few minutes

userquin:
about the `::part` when multiple instances configired,  I'll try to fix using `:nth-of-type(n)` selector, this is the way it should work

packages/plugins-common/constants.ts
```diff
@@ -1,2 +1,3 @@
+export const IMPORT_CSS = '@unocss-import'
```
userquin, line 1: @antfu we can revert this and force to use `@unocss-include`

packages/vite/src/modes/shadow-dom.ts
```diff
@@ -4,8 +4,8 @@ import { IMPORT_CSS } from '../../../plugins-common'
 
 export function ShadowDomModuleModePlugin({ uno }: UnocssPluginContext): Plugin {
   const partRegex = /^part-\[.*]:/
+  const partExtractorRegex = /^part-\[(.+)]:/
```
chu121su12, line 6: I think the bracket capture should be the same.

https://github.com/antfu/unocss/blob/7c08eab2f016d3867f7820ea3250acfb11223b82/packages/preset-mini/src/variants/pseudo.ts#L58

packages/vite/src/modes/shadow-dom.ts
```diff
@@ -13,23 +13,43 @@ export function ShadowDomModuleModePlugin({ uno }: UnocssPluginContext): Plugin
...
+      name,
+      entries: applyParts.map(({ rule, part }) => {
+        const idx = idxResolver(name)
+        return [
+          `.${rule.replace(
+            /\[/g, '\\[',
+          ).replace(
+            /]/g, '\\]',
+          ).replace(
+            /:/g, '\\:')}::part(${part})`,
```
chu121su12, line 31: `rule.replace(/[:[\]]/g, '\\$&')`

As in this example:
https://github.com/antfu/unocss/blob/8c277dee9f411f6325332ccccbba6d1ddc16ac54/packages/core/src/utils/escape.ts#L2

packages/plugins-common/constants.ts
userquin: removed `@unocss-import`, now using only `@unocss-include`

test/fixtures/vite-lit/src/my-element.ts
```diff
@@ -0,0 +1,85 @@
...
+@customElement('my-element')
+export class MyElement extends LitElement {
+  static styles = css`
+    :host {
+      display: block;
+      border: solid 1px gray;
+      padding: 16px;
+      max-width: 800px;
+    }
+    @unocss-include
```
antfu, line 21: `@unocss-include` is a marker for the scanner to force include the file for scanning, while in this case, I think you are using it for another purpose. As it could confuse user, I'd suggest we rename it as `@unocss-placeholder`

userquin: can you check this https://github.com/antfu/unocss/pull/281#discussion_r772723063? I include explitly `@unocss-import` on `IMPORT_CSS` then change the context filter to also check if including it. If we rename it we've braking changes...

userquin: see this commit: https://github.com/antfu/unocss/pull/281/commits/490fe3f9984c79ed63c613dcd22ea98b2beb7e61

alexkarpovnow:
Hi, is there an example in documentation of how we can use UnoCSS with lit framework?

PR #292, opened 2021-12-22
feat(preset-mini): ignore color processing on bracket color (breaking)

chu121su12:
To try to convert to rgba, use `color-#abc` directly or `color-hex-x`.


test/__snapshots__/preset-mini.test.ts.snap
```diff
@@ -71,9 +74,10 @@ exports[`preset-mini > targets 1`] = `
 .border-b{border-bottom-width:1px;border-style:solid;}
 .border-t-2{border-top-width:2px;border-style:solid;}
 .border-x{border-left-width:1px;border-right-width:1px;border-style:solid;}
-.border-\\\\[\\\\#124\\\\]{--un-border-opacity:1;border-color:rgba(17,34,68,var(--un-border-opacity));}
+.border-\\\\[\\\\#124\\\\]{border-color:#124;}
```
antfu, line 77: When used without the opacity, we should keep the variable so it can be controlled by other utils.

test/__snapshots__/preset-mini.test.ts.snap
```diff
@@ -38,9 +38,10 @@ exports[`preset-mini > targets 1`] = `
 .opacity-\\\\$opa{opacity:var(--opa);}
 .opacity-0{opacity:0;}
 .placeholder-opacity-60::placeholder{opacity:0.6;}
-.bg-\\\\[\\\\#153\\\\]\\\\/10,.bg-\\\\[\\\\#1533\\\\]\\\\/10{background-color:rgba(17,85,51,0.1);}
-.bg-\\\\[\\\\#1533\\\\]{background-color:rgba(17,85,51,0.2);}
+.bg-\\\\[\\\\#153\\\\]\\\\/10{background-color:#153;}
```
antfu, line 41: Opacity is lost

chu121su12:
My intention is to just use the bracket value as is, just like the `[rgba()]`/[rgb()] example you don't get opacity via variable.

Options are:
- update color rules to not match bracket + opacity shorthand
- update color rules add opacity variable `--un-x-opacity` anyway if the shorthand is present

antfu:
I think it should work as the color you definite in the theme. If someone intentionally doesn't want the opacity var, they could use 4 digits or 8 digits hex.

PR #311, opened 2021-12-23
new preset `glyph` for embedding subset of font

chu121su12:
## What?
Based on #179 and #213 and what the FontMin package is, I think the purpose is to subset glyph from some font from particular font file and serve it as a base64 embedded font-face rule.

As for the use case, maybe for short text output, probably headings or quotes. You may just serve the whole font if you are using it for body text.

## How?

Rule: `/^glyph-(?<fontName>[\w]+)-(?<glyphs>.+)$/`

This can only make sense in a compiled environment as one needs to provide the font file to be subset(-ted). The fontmin package is used to load a `fontName`, extract only the `glyphs` selected and output an appropriate `@font-face` css. You can then use the specified glyphs in the class/attribute inside the body text with the updated font.

/cc @edwardnyc 

---

## Example
https://github.com/antfu/unocss/blob/eb6620da779d2cde3c539738fca9ca9617fdc2f0/test/preset-glyphs.test.ts#L8

Normal:
image

Hovered:
image

@antfu How do you think about this feature?

Edit:
Using postprocess, it's much cleaner now.


edwardnyc:
Well done. I'm very busy near the Chinese new year, so I don't have time to continue my related work. You have completed the function I want to achieve, thank you

But I think it's better to add union type to string to make TS user-friendly. After all, the input is fixed, what do you think? @chu121su12

antfu:
Maybe it's better to have it as a community preset?

chu121su12:
Got it!

chu121su12:
@edwardnyc The updated branch is on https://github.com/chu121su12/unocss/tree/preset-glyph with single commit. Feel free to use it.

antfu:
@chu121su12 I guess you can publish the preset to npm under your name and see a PR to add it to our community presets list :)

PR #316, opened 2021-12-24
feat(preset-mini): add new flex feature

xiaojieajie:
When flex was originally implemented, it was particularly inconvenient, The basis attribute often needs to be customized，Can't do the existing functions，So I am here to ！

image

image

image

> The picture above is the current problem

**The following is after I added new features**

image

image

**This is specifically compatible with other width types (such as the fit-content in the figure above)**

Can also support x/x writing

### DIFF
--- a/packages/preset-mini/src/rules/flex.ts
+++ b/packages/preset-mini/src/rules/flex.ts
@@ -10,7 +10,7 @@ export const flex: Rule[] = [
   ['flex-auto', { flex: '1 1 auto' }],
   ['flex-initial', { flex: '0 1 auto' }],
   ['flex-none', { flex: 'none' }],
-  [/^flex-(\[.+\])$/, ([, d]) => ({ flex: h.bracket(d) })],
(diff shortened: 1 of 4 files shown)


chu121su12:
Using the bracket helper, you can switch to `flex-[1_1_30%]` and it will replace the underscores with spaces.

xiaojieajie:
> Using the bracket helper, you can switch to `flex-[1_1_30%]` and it will replace the underscores with spaces.

Does it mean I want to replace-with _?

chu121su12:
For bracket values, usually yes. See https://github.com/antfu/unocss/search?q=bracket for example use.

xiaojieajie:
> 对于括号值，通常是。参见https://github.com/antfu/unocss/search?q=bracket例如使用。

Understand, it's a lot more convenient

chu121su12:
> Can also support x/x writing

This is now missing. Still want to support this?

xiaojieajie:
image

image

xiaojieajie:
> > Can also support x/x writing
> 
> This is now missing. Still want to support this?

support

PR #325, opened 2021-12-27
fix(preset-mini)!: grid's bracket syntax now use `_` instead of `,` as separator

### LINKED ISSUE #324: The "grid-cols-[repeat(3,auto)]" class generate error style
When I switch from windicss to unocss.
the "grid-cols-[repeat(3,auto)]" class generate style "grid-template-columns: repeat(3 auto);"
The correct style is "grid-template-columns: repeat(3,auto);"
[codesandbox.io](https://codesandbox.io/s/vite-project-forked-wnzi6?file=/src/App.vue)

image

chu121su12:
Closes #324.

Reorder & update rules; update test.


antfu:
This would be a breaking change, I think we should keep `,` for a while longer

antfu:
Oh ok I see the issue

PR #328, opened 2021-12-27
fix(core)!: fix sorting sheet before use

### LINKED ISSUE #323: Unstable print variant output
Adding print variant In #314, the generated output is unstable.

```css
    .will-change-unset{will-change:unset;}
  + @media print{
  + .print\:block{display:block;}
  + .print\:link\:\!underline:link{text-decoration:underline !important;}
  + }
    @media (prefers-color-scheme: dark){
    .dark\:text-xl{font-size:1.25rem;line-height:1.75rem;}
    .dark\:not-odd\:text-red:not(:nth-child(odd)){--un-text-opacity:1;color:rgba(248,113,113,var(--un-text-opacity));}
    }
    @media (prefers-color-scheme: light){
    .light\:text-sm{font-size:0.875rem;line-height:1.25rem;}
    }
  - @media print{
  - .print\:block{display:block;}
  - .print\:link\:\!underline:link{text-decoration:underline !important;}
  - }
    @media (max-width: 1024px){
    .lt-lg\:m2{margin:0.5rem;}
    }
```

When generated with multiple generator instance or tested the 2nd vs 3rd time, the test passes.

```js
await uno.generate(code)
expect(await uno.generate(code).css).toEqual(await uno.generate(code).css)
```

chu121su12:
Alternative to #329, Closes #323.

On implementing #314, the test couldn't pass due to issue above. The reason as I gather is due to the difference in token resolution speed, the timing when the `sheet`s are added vary, compared to the faster, more predictable resoultion from cache. However, in the `getLayer`, the existing sort usually does nothing to the sheet ordering:

https://github.com/antfu/unocss/blob/14c2f3f0aa9346fc7989106df3506ad1520d49a8/packages/core/src/generator/index.ts#L158

This PR adds additional sorting condition, based on the maxium rule index for each sheet, sorting in descending. The logic here, sorting and comparing, are added in 1 place in the sheet sort callback. However this logic necessitate sorting the rule inside ths sheet sort callback.

This PR also includes the #314 as separate commit for test purpose.


test/__snapshots__/scope.test.ts.snap
```diff
@@ -2,17 +2,17 @@
 
 exports[`scope 1`] = `
 "/* layer: default */
+@media (min-width: 640px){
```
antfu, line 5: This comes to top, which does not looks right to me

packages/core/src/generator/index.ts
```diff
@@ -157,7 +157,19 @@ export class UnoGenerator {
...
 
       let css = Array.from(sheet)
-        .sort((a, b) => (this.parentOrders.get(a[0]) || 0) - (this.parentOrders.get(b[0]) || 0))
+        .sort((a, b) => {
+          const parentOrderA = this.parentOrders.get(a[0])
+          const parentOrderB = this.parentOrders.get(b[0])
+          if (parentOrderA !== undefined && parentOrderB !== undefined)
+            return parentOrderA - parentOrderB
+
+          const ruleOrderA = a[1].map(x => x[0]).sort((a, b) => b - a)[0]
```
antfu, line 166: Use Math.max

chu121su12: Compared to #329, which would you prefer?

antfu: Not sure. If the outcome makes no difference ofc I'd prefer the shorter and performant one

PR #330, opened 2021-12-27
feat(preset-mini)!: use variable prefix for more rules

chu121su12:


antfu:
Wait. Instead of doing this, how about we add a post-process hook that replaces the prefix with the user confined one? So those static ones can keep static

chu121su12:
Do you want to revert #321 or only the css variable-one?

antfu:
I think the ctx change in #321 is a good one. We could only revert the vars part

PR #333, opened 2021-12-28
feat(core)!: callbacks for altering how css is generated

chu121su12:
Sample use on preset-mini:

```js
const uno = createGenerator({
  serializeEntry: (k, v) => `${k.replace(/^--un-/, '--css-uno-')}:${`${v}`.replace(/(\bvar\(--un-)/g, 'var(--css-uno-')};`,
  presets: [
```

image

Modifying raw rule needs to be done on the whole layer with `prepareLayer`. I'm not sure how easy it can be, so I'll take it out if you think it is not necessary.

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -72,6 +72,8 @@ export function resolveConfig(
     blocklist: [],
     safelist: [],
     sortLayers: layers => layers,
+    serializeEntry: (key, value) => `${key}:${value};`,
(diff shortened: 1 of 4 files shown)


antfu:
I have made another approach, please have a look and let me know if it fits your need for `print:`

chu121su12:
I'm sure this is unrelated to the print variant, but the pre- and post- process hooks are great.

PR #337, opened 2021-12-28
refactor(core)!: update `CSSValues` and `CSSEntries` types

chu121su12:
During some of the refactoring, I struggled to keep track of these two types:

https://github.com/antfu/unocss/blob/db28cfa2dd3aeaa0d2184cc756f7d570b1c54b66/packages/core/src/types.ts#L14
https://github.com/antfu/unocss/blob/db28cfa2dd3aeaa0d2184cc756f7d570b1c54b66/packages/core/src/types.ts#L120

I propose singularizing `CSSEntries` to become `CSSEntry[]` and changing `CSSValues` to:
```ts
export type CSSValues = CSSObject | (CSSObject | CSSEntry[])[]
```

So when you encounter `CSSValues`, it's easier to reason about


antfu:
The new type does not seem to be equivalent to the previous one. What's the issue you have faced and what's the benefit of this refactoring?

chu121su12:
Mostly keeping track of the nested array, especially with `CSSEntries` and `CSSEntries[]` and this function:

https://github.com/antfu/unocss/blob/db28cfa2dd3aeaa0d2184cc756f7d570b1c54b66/packages/core/src/utils/object.ts#L7-L18

The major refactor in this PR mostly due to the removal of `CSSEntries` as one of the union of `CSSValues` so rules that return `CSSEntries` needs to be wrapped again in array as `[CSSEntries]` to be valid.

If this is unnecessary, feel free to close this PR, since the core and presets seems solid now especially with the latest v20 release.

antfu:
Ok, I think it's too little to gain to introduce this breaking change.

PR #339, opened 2021-12-29
fix(core)!: resolve dynamic rules

chu121su12:
Resolve `rulesDynamic` the same time as `rulesStaticMap`. Also changed `rulesDynamic` to `[index, Rule]` since the index needs to be tracked.

Also fixes the dynamic rule counter in inspector.

https://github.com/antfu/unocss/blob/db28cfa2dd3aeaa0d2184cc756f7d570b1c54b66/packages/inspector/client/components/Overview.vue#L23


antfu:
Not sure why do you think this refactoring is better, I would avoid extra memory allocation if possible

chu121su12:
Yeah. The treadoff was this vs. loop count

PR #341, opened 2021-12-29
fix(inspector): fix dynamic rules counter

chu121su12:


packages/inspector/client/components/Overview.vue
```diff
@@ -20,7 +20,7 @@ overviewFetch.execute()
           <div op80>
             Rules
           </div>
-          {{ info?.config?.rulesDynamic?.length }} <span op50>dynamic</span><br>
```
antfu, line 23: How about we just do `.filter(Boolean)`?

PR #342, opened 2021-12-29
feat(core): variant ordering

chu121su12:
Addresses https://github.com/antfu/unocss/issues/70 for the feature, not for the preset.


antfu:
Can we have some tests?

PR #354, opened 2021-12-30
feat(preset-wind): add variant: reduced-motion & marker

chu121su12:
Edit: moved to #353 

Please advice if any/all of the variants added should go to `preset-mini` instead.


userquin (commented):
I think motion should be moved to `preset-mini` since the `transition` stuff is also there

chu121su12:
Good catch!

PR #360, opened 2022-01-01
feat(core): add newline between multiple selectors

chu121su12:


antfu:
Why?

chu121su12:
Mostly helping with tracking down long (variant) selector

Issue #364, opened 2022-01-01
Regression from `0.19.0` onwards while using attributify with custom prefix


bhvngt:
My code that was using `attributify` with custom prefix has stopped working since `v0.19.0` till the latest `v0.20.2`.

I have created a [test site](https://stackblitz.com/edit/sveltekit-21npxu?file=package.json) for your reference.

With `0.18.1` - `red` and `blue` backgrounds are displayed based on the screen width. The same stops working once you upgrade to `0.19.0` and beyond.

userquin:
@bhvngt can you resize the panel? it is working but since bg red and blue are scoped to media `sm` and `md` the color is gray: https://imgur.com/O8C0KyX

userquin:
I don't know the exact version where we fix variants, that were wrong, the PR fixing it was https://github.com/antfu/unocss/pull/263: we're doing some refactors, so between `0.18.1` and `0.20.2` versions maybe we have some regresion.

/cc @chu121su12

bhvngt:
> @bhvngt can you resize the panel? it is working but since bg red and blue are scoped to media `sm` and `md` the color is gray: https://imgur.com/O8C0KyX

Yes this version of my test works with 0.18.1. Gray is the default background color. If I upgrade the version to `0.20.2` it stops working.

https://user-images.githubusercontent.com/79074469/147851109-955fe74a-0e79-4f2a-b27a-a470f76e0eb6.mov

> I don't know the exact version where we fix variants, that were wrong, the PR fixing it was #263: we're doing some refactors, so between `0.18.1` and `0.20.2` versions maybe we had some regresion.
> 
> /cc @chu121su12

I tried using `at-` and `min-`. But that too does not seem to be working with `0.20.2`. They work with `0.18.1`. So not too sure if this #263 is the culprit to the regression.

userquin:
@bhvngt sorry, I have seen the 0.20.2 version on the reset but reviewing the package.json file the unocss version is the old one...

bhvngt:
Ohh. My bad. I had it earlier set it with `0.18.1`. Now I have set it as `0.20.2`.

userquin:
@bhvngt check this: https://stackblitz.com/edit/sveltekit-w558hl

You can remove the prefix on the attributify plugin and add a preprocessor, just check the `unocss.config.ts` config file.

bhvngt:
> @bhvngt check this: https://stackblitz.com/edit/sveltekit-w558hl
> 
[...]

PR #370, opened 2022-01-03
fix: revert using merged options from generator

chu121su12:
Fix regression in #364.

@antfu Do you think the options merging should be reverted too?


antfu:
The options in each preset shouldn't be merged together (as it's the preset's custom & specific options). For those meant to be shared should be configure top-level.

PR #376, opened 2022-01-03
feat(preset-mini): add aspect ratio rule keywords

chu121su12:
[Follows tailwind keywords.](https://tailwindcss.com/docs/aspect-ratio)

Should the keyword moved to `preset-wind`?


antfu:
Aspect ratio is now part of the standard and I think it's good to have it in preset-mini

PR #382, opened 2022-01-03
refactor(preset-wind): update list-x

chu121su12:


packages/preset-wind/src/rules/behaviors.ts
```diff
@@ -1,22 +1,22 @@
...
-const listStyleProps = ['disc', 'circle', 'square', 'decimal', 'zero-decimal', 'greek', 'roman', 'upper-roman', 'alpha', 'upper-alpha']
-
 export const listStyle: Rule[] = [
-  [new RegExp(`^list-((?:${listStyleProps.join('|')})(?:-outside|-inside)?)$`), ([, value]) => {
-    const style = value.split(/-outside|-inside/)[0]
-    const position = /outside|inside/.exec(value) ?? []
-
-    if (position.length) {
+  // base
+  [/^list-(disc|circle|square|decimal|zero-decimal|greek|roman|upper-roman|alpha|upper-alpha)(?:-(outside|inside))?$/, ([, style, position]) => {
```
antfu, line 6: Why do we do this refactoring? Static regex is more efficient then dynamic generated one

chu121su12: The constructor is removed and changed with inlined union `|`. Did I misuse the syntax?

antfu: Oh sorry I mis read the diff in wrong direction 😅😅

PR #390, opened 2022-01-04
feat(preset-wind): add `content-none`

chu121su12:


antfu:
We could have this in preset-mini

PR #395, opened 2022-01-04
fix(preset-mini): restore inset position, add inset tests

sondh0127:
fix https://github.com/antfu/unocss/issues/394


chu121su12:
I'm sorry the latest PR for the Issue. Could you please make the rule preserve the order of `inset-[unit]` (without x/y)?

sondh0127:
> I'm sorry the latest PR for the Issue. Could you please make the rule preserve the order of `inset-[unit]` (without x/y)?

No problem. Your work was fantastic.
I fixed the order. I add a map for zero: 0rem => 0. Is that look good?

chu121su12:
Maybe split the 0 handling to another PR since most sizing rules with handler may benefit from this.

packages/preset-mini/src/rules/position.ts
```diff
@@ -96,7 +96,7 @@ function handleInsetValues([, d, v]: string[]): CSSEntries | undefined {
 }
 
 export const insets: Rule[] = [
-  [/^(?:position-|pos-)?inset-()(.+)$/, handleInsetValues],
+  [/^(?:position-|pos-)?inset-()(.+)$/, ([, , v]) => ({ inset: handleInsetValue(v) })],
```
chu121su12, line 99: The empty capture `()` can be removed.

sondh0127: Alright, I'm new to this. Thanks for figuring it out

chu121su12:
Thanks!

PR #399, opened 2022-01-04
fix(preset-wind): update filters

chu121su12:
- add `backdrop-opacity`
- use `.` regex for `h.bracket`
- prioritize theme's value on `drop-shadow`
- update no-value matches


antfu (approved):
You can merge it after conflicts resolved

PR #408, opened 2022-01-04
feat(vite): detect app behind frontend server

userquin:
This PR allows enable the `UnoCSS` plugin for `global` model when the Vite application is behind a frontend server, also on dev mode: for example using `PHP + Laravel + Inertia`.

I added also a custom server configuration, I don't know if this makes sense, but @chuoke request it on the disccusion linked bellow.

For context see:
- https://github.com/antfu/unocss/issues/254
- https://github.com/antfu/unocss/discussions/387


antfu:
Could we somehow detect automatically (from vite config or client runtime) without introduce a separate option for that?

userquin:
I'm editing the file to inline the configuration using the vite types...

userquin:
> Could we somehow detect automatically (from vite config or client runtime) without introduce a separate option for that?

The problem is we don't know if the app is behind a front end, so, how we can detect that without any hist from the use land?

userquin:
For example, I use apache + dev server, the vite app cannot infer it apache is in frint of it

userquin:
we can include always the `http(s)://<host>:<port>` but the problem is the request `mode`, we can also include it always

userquin:
@chuoke with the new approach, you don't need to add nothing on the plugin configuration, we added the logic to detect automatically on the client side

hunteryun:
Uh, I've been watching (waiting) this, wondering what's going on right now? Are you preparing for a better plan? Can anyone reveal some plans?

呃，我一直在关注（等待）这个，不知道现在是怎么情况？是在准备更好的方案吗？谁能透露一些计划吗？

PR #411, opened 2022-01-04
feat(core): deduplicate shortcut instead of merging

chu121su12:
I haven't been following tailwind v2 internals closely, but at v3 its modifier classes (`transform`, `ring`, `filter`, etc.) does not provide the base css properties anymore (respectively: `transform`, `box-shadow`, `filter`, etc.) Instead, they define the variable versions (`--tw-transform`, `--tw-shadow`, `--tw-filter`, etc.), and then the base properties being moved to each modifiers.

Ex. previously: 
```css
.scale-2 {
  /* base transform styles */
  transform: var() /* combined vars. */;
}
.scale-2 {
  --tw-scale: 2;
}
```

now:
```css
.scale-2 {
  /* base transform styles */
  --tw-transform: var() /* combined vars. */;
}
.scale-2 {
  --tw-scale: 2;
  transform: var(--tw-transform);
}
```

While this is trivial to implement in normal rules, it becomes tricky to do in shortcut, as currently shortcuts get the merged properties of its definitions.

My proposed solution is to resolve shortcut definition individually like regular rules, so the orders are preserved. One drawback to this method is duplicated selector for each shortcut due to the css properties aren't merged.

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -1,6 +1,6 @@
 import type { CSSEntries, CSSObject, ExtractorContext, GenerateOptions, GenerateResult, ParsedUtil, RawUtil, ResolvedConfig, RuleContext, RuleMeta, StringifiedUtil, UserConfig, UserConfigDefaults, UtilObject, Variant, VariantContext, VariantHandler, VariantMatchedResult } from '../types'
 import { resolveConfig } from '../config'
-import { TwoKeyMap, e, entriesToCss, expandVariantGroup, isRawUtil, isStaticShortcut, normalizeCSSEntries, normalizeCSSValues, notNull, uniq, warnOnce } from '../utils'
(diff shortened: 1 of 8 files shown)


antfu:
I think shortcuts are supposed to be merged. We might need to somehow separate the base styles, but sperate all components in shortcuts is not an ideal solution to me.

chu121su12:
Two more options I can think off: 1) not supporting this behavior at all or 2) promote the handling of CONTROL_BYPASS_PSEUDO into core and make it as marker to tell that the entries should be separated (implementing twv3 behavior remove the need to use CONTROL_BYPASS_PSEUDO in the first place).

antfu:
There is no shortcut concept in TW, I am not sure how you mean by aligning its behavior.

I think we could have a `CONTROL_BYPASS_SHORTCUT_MERGING` in core

chu121su12:
What I meant was implementing this way of modifier handling removes the need to use CONTROL_BYPASS_PSEUDO

```css
.scale-2 {
  /* base transform styles */
  --tw-transform: var() /* combined vars. */;
}

antfu:
Oh ok, go for that. But it's not related to this PR right?

chu121su12:
This or the CONTROL_SHORTCUT_NO_MERGE is related for shortcut support for tw3 behavior.

PR #416, opened 2022-01-05
fix(inspector): update `codemirror` mode map for `jsx/typescript`

userquin:


antfu:
@userquin do you think at some point we should make a codemirror wrapper for Vue 3 (or even with codemirror@6) 👀

userquin:
@antfu we should ;), I'll take a look at version 6

PR #433, opened 2022-01-07
(preset-mini)!: remove x-width-y rule

chu121su12:
This PR removes `width-` in `width-|size-` rule simplifying rules to only use `size-`.

Alternatively, should the `width-` added to every `size-` rule? Especialy for those which return `x-width` properties.


antfu:
What's the problem of having `outline-width`? I think it's good as it matches with the CSS prop

chu121su12:
Alternative PR adding `width` in #437.

PR #436, opened 2022-01-08
Add instruction to add UnoCSS in Sveltekit

sharu725:


userquin:
@sharu725 this app fw is already on docs, just check vite package

sharu725:
> @sharu725 this app fw is already on docs, just check vite package

That one actually has better config. I should start reading the doc 🤦

PR #439, opened 2022-01-08
update Sveltekit config requirements

sharu725:
Apparently, ``svelte-preprocess`` is not required for UnoCSS. 
I have tested a build without it. Everything works as usual, even the class directives. Just the extractor is enough.

Also, ``target: '#svelte'`` is optional in Sveltekit. Sveltekit uses body of the app if no target is present. 
So I think it isn't necessary to show it in here.

### DIFF
--- a/packages/vite/README.md
+++ b/packages/vite/README.md
@@ -215,20 +215,12 @@ You can use simple rules with `class:`, for example `class:bg-red-500={foo}` or 
 
 ```ts
 // svelte.config.js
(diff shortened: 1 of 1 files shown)


userquin:
@sharu725 we use the default templates from the ui/app fw, removing some resources since we don't need to show how to configure UnoCSS (linter, icons, etc...). In this specific case, we work with some core team member: see https://github.com/antfu/unocss/pull/232 and https://github.com/antfu/unocss/pull/246

PR #443, opened 2022-01-09
feat(runtime): add simple css autoprefixer

chu121su12:
This PR adds basic css autoprefixer (from vue).

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -1,5 +1,6 @@
-import type { UnoGenerator, UserConfig, UserConfigDefaults } from '@unocss/core'
+import type { Postprocessor, UnoGenerator, UserConfig, UserConfigDefaults } from '@unocss/core'
 import { createGenerator } from '@unocss/core'
+import { capitalize } from '@unocss/preset-mini/utils'
 
 export interface RuntimeOptions {
   /**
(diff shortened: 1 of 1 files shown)


chu121su12:
Considering the rules are all kebab case, the `autoPrefix` function can be simplified without `hyphenate`:

```ts
  function autoPrefix(rawName: string): string {
    const cached = prefixCache[rawName]
    if (cached)
      return cached
    let name = camelize(rawName)
    if (name !== 'filter' && name in elementStyle)
      return (prefixCache[rawName] = rawName)
    name = capitalize(name)
    for (let i = 0; i < prefixes.length; i++) {
      if (`${prefixes[i]}${name}` in elementStyle)
        return (prefixCache[rawName] = `-${prefixes[i].toLowerCase()}-${rawName}`)
    }
    return rawName
  }
```

packages/runtime/src/index.ts
```diff
@@ -154,3 +164,32 @@ export default function init(options: RuntimeOptions = {}) {
   window.addEventListener('load', execute)
   window.addEventListener('DOMContentLoaded', execute)
 }
+
+function autoPrefixer(): Postprocessor {
```
antfu, line 168: That could be a very complex thing to do - are you really sure? 👀

If so, I think we should have a separate package like `@unocss/autoprefixer` and add unit tests for it

chu121su12: I can see the benefit of autoprefixer, but I only do this in runtime because I'm underqualified to do package & bundling.

I tracked the use of this function in vue and so far I can only find it and its usage inside `runtime-dom`.
https://github.com/vuejs/vue-next/blob/92f11d6740929f5b591740e30ae5fba50940ec82/packages/runtime-dom/src/modules/style.ts#L69-L70

Also the related commit by Evan himself
https://github.com/vuejs/vue-next/commit/2b2727e62cf791d301dc254628d965a69bf3398a

antfu: Ok, at least we need to have this function in a separate file and test it via direct imports

antfu:
- Would love to see more tests, and maybe compare to with the with/without
- We might need an option to disable this behaviour on runtime (or even disable by default to be consistent with build approach

chu121su12:
Reference
https://github.com/vuejs/core/blob/main/packages/runtime-dom/src/modules/style.ts

PR #444, opened 2022-01-10
feat(preset-mini,preset-wind): add logical shorthand

chu121su12:
Getting ahead of #438, this PR adds shorthands for logical spacings/sizes:
- `k`: bloc***k***-start
- `d`: block-en***d***
- `o`: bl***o***ck (block-start and block-end, block-size)
- `n`: i***n***line (inline-start and inline-end, inline-size)


antfu:
😂 I am not sure if this is clear enough, as you can't actually guess what is from reading the shorthand

chu121su12:
It was difficult to balance with only 1 letter. Especially with the sides sharing similar keyword for block-x or inline-x. Properly conveying the sides would need 2-letters api `bs`/`be`/`is`/`ie`. And using 2 letters for compound sides: `bl` for `y` replacement and `in` for `x` to not clash with 1-side shortcut.

I'll try this with other PR for clarity.

antfu:
I feel those might be rarely used, could we use have the full name without shorthand?

PR #445, opened 2022-01-10
fix(preset-wind): use space for standalone placeholder rule

chu121su12:


packages/preset-wind/src/rules/placeholder.ts
```diff
@@ -2,6 +2,6 @@ import type { Rule } from '@unocss/core'
 import { colorResolver, handler as h } from '@unocss/preset-mini/utils'
 
 export const placeholders: Rule[] = [
-  [/^\$-placeholder-(.+)$/, colorResolver('color', 'placeholder')],
-  [/^\$-placeholder-op(?:acity)?-?(.+)$/, ([, opacity]) => ({ '--un-placeholder-opacity': h.bracket.percent.cssvar(opacity) })],
+  [/^\$ placeholder-(.+)$/, colorResolver('color', 'placeholder')],
```
antfu, line 5: Could add a short comment here to state that it's handled by a variant

PR #446, opened 2022-01-10
feat(core): `noMerge` rule meta

chu121su12:
See https://www.w3.org/TR/selectors-3/#grouping

Mostly for safely using prefixed pseudo element (via variant).

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -110,7 +110,7 @@ export interface RuleMeta {
    * Option to not merge this selector even if the body are the same.
    * @default false
    */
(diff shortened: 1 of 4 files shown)


packages/core/src/types.ts
```diff
@@ -106,6 +106,11 @@ export interface RuleMeta {
    * @default 'default'
    */
   layer?: string
+  /**
+   * Option to not merge this selector even if the body are the same.
+   * @default false
+   */
+  splitSelector?: boolean
```
antfu, line 113:
```suggestion
  noMerge?: boolean
```

PR #456, opened 2022-01-11
refactor(preset-mini,preset-wind): simplify shadow

chu121su12:
The actual refactor is in https://github.com/antfu/unocss/compare/32214ebb5bf183ff665598109a19ce4a29d6bdd6...cba70e84c4e204876618038a0c623b7840daac7b

This PR removes the need for extra `shadow-x-colored` variable by directly assigning the shadow's color value to css-var's fallback value. Previously the css variables are separate and the `shadow-color-x` utilities switch the shadow value from theme to colored version.


antfu:
Can you resolve the conflicts? Thanks

PR #468, opened 2022-01-13
fix(svelte-scoped): try to prepend to existing style element

### LINKED ISSUE #447: Svelte-scoped mode does not take into consideration that a component already has a style tag
Svelte only accepts 1 style tag per component, so this behaviour forces the user to not use any style tags in their component.

Source: https://github.com/antfu/unocss/commit/0617733e8075ccc8953527c6a26409a600e3c3bf#diff-ac394d603804c8ed219c43572638b9f66088f440025a822da4e9fa4bdb4ae5cb

chu121su12:
Fixes #447

/cc @vipero07

### DIFF
--- a/packages/vite/src/modes/svelte-scoped.ts
+++ b/packages/vite/src/modes/svelte-scoped.ts
@@ -10,6 +10,8 @@ export function SvelteScopedPlugin({ uno, ready }: UnocssPluginContext): Plugin 
     const { css } = await uno.generate(code)
     if (!css)
       return null
+    if (code.match(/<style[^>]*>[\s\S]*?<\/style\s*>/))
+      return code.replace(/(<style[^>]*>)/, `$1${css}`)
     return `${code}\n<style>${css}</style>`
   }
 


packages/vite/src/modes/svelte-scoped.ts
```diff
@@ -10,6 +10,8 @@ export function SvelteScopedPlugin({ uno, ready }: UnocssPluginContext): Plugin
     const { css } = await uno.generate(code)
     if (!css)
       return null
+    if (code.match(/<style[^>]*>[\s\S]*?<\/style\s*>/))
+      return code.replace(/(<style[^>]*>)/, `$1${css}`)
```
antfu, line 14: Should we add it to the end?

chu121su12: Either way. My thought is adding at the beginning should allow the css in the style element override the unocss-generated ones, just as the recent PR https://github.com/antfu/unocss/pull/458

antfu: Anyway, let's have it until someone complain about it 👀

PR #484, opened 2022-01-15
feat(reset): update tailwind preflight

chu121su12:
https://github.com/tailwindlabs/tailwindcss/blob/1eb4127de233e9323c793878d492a4c3c30c2b3b/src/css/preflight.css

Should we update/remove this rule?
https://github.com/antfu/unocss/blob/3861968ea7eab6190b2930d4f29c3cc95b02f382/packages/reset/tailwind.css#L15-L18

### DIFF
--- a/packages/reset/tailwind.css
+++ b/packages/reset/tailwind.css
@@ -12,11 +12,6 @@
   border-color: currentColor; /* 2 */
 }
 
-::before,
-::after {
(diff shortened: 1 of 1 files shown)


antfu:
I don't think we need this

```css
 ::before, 
 ::after { 
   --tw-content: ''; 
 } 
```

PR #513, opened 2022-01-18
refactor(preset-mini)!: rework logical block/inline

chu121su12:
- Rename `block-x`|`inline-x` to `lb-x`/`li-x`
- Simplify corner radius shorthand


antfu:
Same https://github.com/antfu/unocss/pull/514#issuecomment-1018197724

PR #514, opened 2022-01-18
refactor(preset-wind)!: rework divide/space-block/inline

chu121su12:
Rename `block-x`/`inline-x` to `b-x`/`i-x` since block mappings should not have edge/corner.


antfu:
Not so sure about this, even spacing won't have `bottom`, having `b` might still be confusing for readers who are not aware of the context. I would suggest we keep the original to be more explict

PR #520, opened 2022-01-20
feat(runtime): add `extract` to process specified token(s)

chu121su12:


antfu:
What's the use case of this?

chu121su12:
My use case in vue is to preempt style update before render begin. This should ensure the css are available particularly before animating/showing overlay etc.

PR #528, opened 2022-01-23
feat(core): css color parser

chu121su12:
Add support for parsing css color function.
https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

If this can proceed, the next steps are:

1. To parse proper color component type as currentyl everything is string-based. However, with current presets logic, only the `alpha` is used for arithmatic while the components mostly being joined.
2. To switch the presets from using `RGBAColorValue` to the new color type.

My concerns are whether we should support css built-in color keywords https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#color_keywords. My take is just to support transparent/black/white & rebeccapurple


chu121su12:
Should be good for now, except for more test & name changes.

antfu:
I am a bit concerned about the complexity added to the core. Many at this point, we should extract those utilities in core and preset-mini to a more general `@unocss/utils` package?

chu121su12:
I'm ok eaither way. Although some that I can think of splitting mostly related to css value utilities.

PR #535, opened 2022-01-24
feat(runtime): add callback to modify default options

chu121su12:
My usecase is to insert additional preset at runtime, because currently passing runtime option override the existing preset.

```
window.__unocss = {
    runtime: { autoPrefix: true },
    prepare(e) {
        e.preflights = [
            tailwindReset,
            tailwindForms,
        ];
        e.presets.push(typography());
    },
};
```

Please advice better way to do this if I miss something.


packages/runtime/src/index.ts
```diff
@@ -12,6 +12,10 @@ export interface RuntimeOptions {
    * @default false
    */
   autoPrefix?: boolean
+  /**
+   * Callback to modify default settings
+   */
+  prepareDefaults?: (defaults: UserConfigDefaults) => void
```
antfu, line 18: How about `configResolved` to align with Vite's plugin api?

PR #550, opened 2022-01-27
feat(runtime): fire a callback when the runtime is ready

chu121su12:
Two uses:
- to make the default extraction (`extractAll`) preventable
- to be able to config the runtime from one place. ex:

```js
window.__unocss = { runtime: {
    ready({ extract }) {
        fetch(...)
            .then(r => r.text)
            .then(t => extract(t))
```

instead of 

```html
<script src="./uno.global.js"></script>
<script>
const { extract } = window.__unocss_runtime;
fetch(...)
    .then(r => r.text)
    .then(t => extract(t))
</script>

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -75,7 +75,7 @@ export interface RuntimeContext {
 declare global {
   interface Window {
     __unocss?: UserConfig & { runtime?: RuntimeOptions }
-    __unocss_runtime?: RuntimeObject
+    __unocss_runtime?: RuntimeContext
   }
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/runtime/src/index.ts [hunk 1]
```diff
@@ -16,60 +16,66 @@ export interface RuntimeOptions {
    * Callback to modify config
    */
   configResolved?: (config: UserConfig, defaults: UserConfigDefaults) => void
+  /**
+   * Callback when the runtime is ready. Returning false will prevent default extraction
+   */
+  ready?: (runtime: RuntimeObject) => false | any
```
antfu, line 22:
```suggestion
  ready?: (runtime: RuntimeContext) => false | any
```

PR #568, opened 2022-01-31
feat(preset-mini): update/add variants-[x] that target x as raw selector

chu121su12:
Used to target dom injected by external library. Now uses selector-based.

Clean commit is in [this comparison](https://github.com/antfu/unocss/compare/b296c19aa9038e0db1544cad82b6791da3582d45...57447bde1b130d561f6f3208fc3fc8aa6f2cd91c)

Updated variants:
- `all-[target]`: `[class name] [target]`
- `children-[target]`: `[class name] > [target]`
- `next-[target]`: `[class name] + [target]`
- `sibling-[target]`: `[class name] + [target]`
- `siblings-[target]`: `[class name] ~ [target]`
- `group-[target]`: `[target] [class name]`
- `parent-[target]`: `[target] > [class name]`
- `peer-[target]`: `[target] ~ [class name]`

New:
- `previous-[target]`: `[target] + [class name]`


antfu:
I like this idea, but how about extending the current modifiers, something like `children-[child-class]-xxx` and `parent-[parent-class]-xx`?

chu121su12:
The PR has now expanded into multiple variants:

`class-[target]`: `[target] [class name]` (same with `all`)
`class-all-[target]`: `[target] [class name]`
`class-children-[target]`: `[target] > [class name]`
`class-next-[target]`: `[target] + [class name]`
`class-sibling-[target]`: `[target] + [class name]`
`class-siblings-[target]`: `[target] ~ [class name]`
`class-group-[target]`: `[class name] [target]`
`class-parent-[target]`: `[class name] > [target]`
`class-previous-[target]`: `[class name] + [target]`
`class-peer-[target]`: `[class name] ~ [target]`

chu121su12:
> I like this idea, but how about extending the current modifiers, something like `children-[child-class]-xxx` and `parent-[parent-class]-xx`?

I tried without prefixing the variants (group/parent/peer) conflicts with attributify mode. Maybe this needs to be investigated further

PR #575, opened 2022-02-03
feat(preset-uno): sass mix color function

chu121su12:
Alternative for https://github.com/antfu/unocss/pull/570 by adding to preset-uno instead

It's breaking now since `presetUno` is no longer alias for `presetWind`. See https://github.com/antfu/unocss/pull/225#issuecomment-990499591

### DIFF
--- a/packages/preset-uno/src/variants/mix.ts
+++ b/packages/preset-uno/src/variants/mix.ts
@@ -3,6 +3,15 @@ import { parseCssColor } from '@unocss/preset-mini/utils'
 
 const mixComponent = (v1: string | number, v2: string | number, w: string | number) => `calc(${v2} + (${v1} - ${v2}) * ${w} / 100)`
 
+/**
+ * Returns RGB color that's a mixture of color1 and color2. Support RGB color values.
+ * https://sass-lang.com/documentation/modules/color#mix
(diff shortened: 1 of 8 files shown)


packages/preset-uno/src/variants/mix.ts
```diff
@@ -0,0 +1,54 @@
...
+    alpha: mixComponent(cssColors[0].alpha ?? 1, cssColors[1].alpha ?? 1, weight),
+  }
+}
+
+const tint = (color: string | CSSColorValue, weight: string | number) => mixColor('#fff', color, weight)
+const shade = (color: string | CSSColorValue, weight: string | number) => mixColor('#000', color, weight)
+const shift = (color: string | CSSColorValue, weight: string | number) => parseInt(`${weight}`, 10) > 0 ? shade(color, weight) : tint(color, `-${weight}`)
+const fns: Record<string, (color: string | CSSColorValue, weight: string | number) => CSSColorValue | undefined> = { tint, shade, shift }
+
+export const variantColorMix: Variant = (matcher) => {
```
antfu, line 32: Can you add a short JSDoc here to describe how this util work? This could help ppl to read the code and find the usage.

PR #579, opened 2022-02-03
fix(preset-wind):animate-[] returns animate -> animation

xmsz:
origin:
```
.animate-\[4s_linear_0s_test\] {
    animate: 4s linear 0s test;
}
```

animate is not a vaild value


chu121su12:
Please update the tests too

PR #596, opened 2022-02-10
feat(runtime): add new runtime that only registers the available presets

chu121su12:
Mainly to be able to use the typography without introducing permutation for the existing runtime presets


antfu:
Not so sure about it. I think we could have a full build for all presets, other than that, I would recommend using bundler to build custom presets

PR #597, opened 2022-02-10
feat(runtime): add `full` runtime

chu121su12:
Currently included presets are 
- attributify
- uno
- typography

I tried adding web-fonts but the file size jumped way high. Let me know if it should be added anyway


antfu:
for `web-fonts`, it's because of `axios` deps. We could use a lighter client, or just use fetch on CDN.

PR #609, opened 2022-02-12
fix(preset-icons): add svg xlink xml namespace

### LINKED ISSUE #608: [bug](@unocss/preset-icons) Losing the xmlns:xlink namespace when translating svg causes the icon to fail
Steps to reproduce:
1. Install @iconify-json/flat-ui
2. Configure unocss and @unocss/preset-icons
3. Use any icon that require the xlink attribute

image

After statistics, there are a total of 30 invalid icons that require the xlink namespace

- Valid icon (no require the xlink attribute): arrow, bag ...
- Invalid icon (require the xlink attribute): art ...

image

image

As far as I know, using the xlink attribute in svg needs to declare the xmlns:xlink namespace, like:

`<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"></svg>`

In the @unocss/preset-icons preset, the generated svg tags are generated via the iconToSVG function of @iconify/utils.

https://github.com/unocss/unocss/blob/b3e438c9c7c7000eb31c03ba29bc8ec5e9dac9a9/packages/preset-icons/src/index.ts#L43-L48

But the iconToSVG function only handles the four properties of width/height/preserveAspectRatio/viewBox.

https://github.com/iconify/iconify/blob/ded43ed8fc7bc4d1960cc6077310c0c7eed5df43/packages/utils/src/svg/build.ts#L221-L230

After I manually edit to add the xmlns:xlink namespace, the icon works.

image

Recently I have been using unocss in depth to build new modern applications, including but not limited to xlink namespaces are very useful for other beautiful svg illustrations, I hope this problem can be solved, thanks to the contributors for their efforts.

userquin:
closes #608


userquin:
I'll fix it also on `unplugin-icons`

PR #615, opened 2022-02-15
feat(preset-mini): add selector variant

chu121su12:


antfu:
I would feel this might obey the idea of having "atomic css" as it will populate the global styles. Not sure if want to have this pattern by default

chu121su12:
While I relate with the globals this rule can create, my main intention originally is for the rule to be able to target `:root` and its `@media` variants and other reset styles.

Feel free to close this PR. Maybe this could be revisited in the future (or not), but I'm sure this rule must come with big usage warning in the docs.

PR #619, opened 2022-02-16
BREAKING CHANGE: refactor(preset-mini,preset-wind): move variants to preset-wind

chu121su12:
- ~move `dark`/`light` variants~
- move `motion-reduce`/`motion-safe`
- move `portrait`/`landscape`

print variant is staying, but please advise if it should be moved too


antfu:
I would like to keep the dark/light variants to mini, other than that I am fine with moving to wind

antfu (approved):

PR #640, opened 2022-02-22
refactor(preset-mini,preset-wind): update themes

chu121su12:
- use 3px as default ring value


antfu:
I am still not sure. Other props like border are default to `1px`, having this as `3px` specially feel quite unintutive to me.

chu121su12:
I'll submit the refactor later with other theme updates.

PR #641, opened 2022-02-22
feat(preset-mini,preset-wind): add basic opacity theme

chu121su12:
- add in base `opacity-x`
- add in rules `x-opacity-y`
- theme opacity in `colorParser`

### DIFF
--- a/packages/preset-mini/src/rules/border.ts
+++ b/packages/preset-mini/src/rules/border.ts
@@ -108,8 +108,8 @@ function handlerBorderColor([, a = '', c]: string[], { theme }: RuleContext<Them
   }
 }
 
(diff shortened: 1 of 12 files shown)


antfu:
What would you expect the use case of this?

chu121su12:
I did this to add theme for opacity for consistency. I'm actually unsure for the use case, hence this is a separate PR.

One I can think of is maybe for pseudo styling. Otherwise, feel free to close this.

antfu:
I don't mind having it, just not sure if it's something that actually have ppl using it

chu121su12:
Maybe we can reintroduce this when there's a demand for it

PR #656, opened 2022-02-26
Doc: Updated status badges

rubiin:


antfu:
Thanks but I'd like to keep it clean and simple.

PR #660, opened 2022-02-27
Update README.md

rubiin:
Added `unocss-question-mark` plugin that provides a helpful ? dev time utility for unocss. The plugin is based on `tailwindcss-question-mark`


ydcjeff:
I think this is already built-in: https://github.com/unocss/unocss/blob/main/packages/preset-mini/src/rules/question-mark.ts

userquin:
> Added `unocss-question-mark` plugin that provides a helpful ? dev time utility for unocss. The plugin is based on `tailwindcss-question-mark`

Do you know that the question mark utility is here (only on dev)? https://streamable.com/ys457h

rubiin:
Closing this as it has already been implemented

PR #665, opened 2022-02-28
feat: source map support for transformers

hannoeru:


antfu:
I think the core (aim to be lightweight) should not deps on `magic-string` but only  the transformers.

hannoeru:
core package only uses types from `magic-string`, `magic-string` is only install as dependencies on vite package.

antfu:
@hannoeru @chu121su12 Can you help resolve this conflict? 😅

antfu:
Thanks!

PR #680, opened 2022-03-06
add cli under integrations

veritem:


antfu:
Thanks for your contribution. I have removed the docs/ dir and let's focus on the interactive docs first :)

PR #697, opened 2022-03-11
feat(directives): parse nested class

hannoeru:
- This will **not** transform nested class syntax, only parse and transform `@apply` directive.


antfu:
Awesome!

PR #709, opened 2022-03-14
fix(preset-mini): breakpoint in max-width issue

### LINKED ISSUE #707: RWD issue: lt-${breakpoint} should be `${breakpointValue - 0.1px}`
# Description

## TL;DR
- breakpoint: md
- breakpointValue: 768px

`lt-${breakpoint}` should be `${breakpointValue - 0.1px}` instead using ${breakpointValue} directly.

After I read the [source code](https://github.com/unocss/unocss/blob/main/packages/preset-mini/src/variants/breakpoints.ts#L37-L49), I think `at-${breakpoint}` do have the same issue. 

I could open the PR to fix it if needed.

## Situtation
I found the issue causes the user who uses screen which right `at` breakpoint will see nothing when I tried to use `lt-${breakpoint}:none` and `${breakpoint}:none` to control the different components like below sample code.

components
```jsx
const Desktop = () => <div display="flex lt-md:none">desktop</div>
const Mobile = () => <div display="flex md:none">mobile</div>
```` 
 output
```css
[display~="flex"]{display:flex;}

/* I expect to get `767.9px` instead of `768px` */
@media (max-width: 768px){
   [display~="lt-md\:none"]{display:none;}
}

@media (min-width: 768px){
   [display~="md\:none"]{display:none;}
}
```

# REF
[windicss-rwd-custom-range](https://windicss.org/features/responsive-design.html#custom-range)

achinchen:
# description
- Fix `lt:${breakpoint}` and `at:${breakpoint}` to close #707

# ref
- [windicss/tools#increaseWithUnit](https://github.com/windicss/windicss/blob/4ec1027483bf9e3c7e6bf4f3397912b8f7aec243/src/utils/tools.ts#L342-L360)

### DIFF
--- a/packages/preset-mini/src/variants/breakpoints.ts
+++ b/packages/preset-mini/src/variants/breakpoints.ts
@@ -3,6 +3,13 @@ import type { Theme } from '../theme'
 
 const regexCache: Record<string, RegExp> = {}
 
+const calcMaxWidthBySize = (size: string) => {
+  const value = size.match(/^-?[0-9]+\.?[0-9]*/)?.[0] || ''
+  const unit = size.slice(value.length)
+  const maxWidth = (parseFloat(value) - 0.1)
+  return Number.isNaN(maxWidth) ? size : `${maxWidth}${unit}`
+}
+
 export const variantBreakpoints: Variant<Theme> = (matcher, { theme }) => {
   const variantEntries: Array<[string, string, number]>
       = Object.entries(theme.breakpoints || {}).map(([point, size], idx) => [point, size, idx])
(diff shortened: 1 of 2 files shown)


chu121su12:
Note that in bootstrap, the delta is 0.02px

https://github.com/twbs/bootstrap/blob/8fe82c7176a577190f2d69f9cbdd990b7f3c8493/scss/mixins/_breakpoints.scss#L34-L46

achinchen:
> Note that in bootstrap, the delta is 0.02px
> 
> > Uses 0.02px rather than 0.01px to work around a current rounding bug in Safari.
> 
> https://github.com/twbs/bootstrap/blob/8fe82c7176a577190f2d69f9cbdd990b7f3c8493/scss/mixins/_breakpoints.scss#L34-L46

Due to I use the delta is `0.1` instead of `0.01`, I think it's okay. There's [demo for delta 0.1](https://jsbin.com/rapusep).

However it's quite interesting information for the issue with 0.01 vs 0.02 in Webkit. 
I did the experiment to confirm the "delta 0.01 bug" in different browsers by using the [demo](https://output.jsbin.com/tafogug/quiet) from bug report page, Firefox in MacOS is good for 0.01 but Chrome in MacOS isn't. 

Thanks `(:`

PR #716, opened 2022-03-15
feat: autocomplete support

QiroNT:
Currently just a proof of concept, only implemented `border` right now.

Added a `uno.suggest` method, passing the input in and will return an array of suggestions.

Sample for implementing suggestion for a rule:
```ts
export const borders: Rule[] = [
  // ...
  [/^(?:border|b)()(?:-(.+))?$/, handlerBorder, suggestBorder],
  // ...
]

function suggestBorder(i: string, ctx: SuggestionContext<Theme>): string[] | undefined {
  return new MiniSuggestionBuilder(i, ctx)
    // choose in
    // passing `true` to second argument means accept both with or without a trailing dash
    .withOptions(['border', 'b'], true)
    // follow either path
    .withEither([
      b => b
        // can have one of
        .withMaybeOptions(Object.keys(directionMap).filter(Boolean), true)
        .withEither([
          // first array for values without a trailing dash(eg. `border-b`),
          // second for values with a trailing dash(eg. `border-width-`)
          b => b.withMaybeOptions([], ['width', 'size'])
            .withOptions(Array.from(new Set([
              ...Object.keys(ctx.theme.lineWidth || {}).filter(i => i !== 'DEFAULT'),
              '0', '2', '4', '8',
            ]))),
          b => b.withMaybeOptions([], ['color'])
            // with a color
            .withColor(),
          b => b.withOptions([], ['op', 'opacity']),
        ]),
      b => b.withOptions([], ['rounded', 'rd'])
        .withMaybeOptions(Object.keys(cornerMap).filter(Boolean), true)
        .withOptions(Object.keys(ctx.theme.borderRadius || {}).filter(i => i !== 'DEFAULT')),
    ])
    .collect()
}

```

Result preview:

/cc @antfu


QiroNT:
Follow #723

PR #718, opened 2022-03-15
fixture(remix): add logo icon

userquin:
To test also preset icons: https://streamable.com/jy0dvh


userquin:
/cc @johannschopplich 

on local there is an error about `.cache` is blocked when running the cli test:

imagen

userquin:
just to test the preset icons is also working, all fw fixtures have it

PR #724, opened 2022-03-16
feat(autocomplete): support nested theme object interpolation

QiroNT:
`DEFAULT` is not supported since it requires nested groups.

### DIFF
--- a/packages/autocomplete/src/parse.ts
+++ b/packages/autocomplete/src/parse.ts
@@ -1,4 +1,4 @@
-import type { AutocompleteTemplatePart, ParsedAutocompleteTemplate } from './types'
+import type { ACTDeepGroupMember, AutocompleteTemplatePart, ParsedAutocompleteTemplate } from './types'
 
 const shorthands = {
   '#num': `(${[0, 1, 2, 3, 4, 5, 6, 8, 10, 12, 24, 36].join('|')})`,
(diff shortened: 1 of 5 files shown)


packages/autocomplete/src/parse.ts
```diff
@@ -23,10 +24,18 @@ export function parseAutocomplete(template: string, theme: any = {}): ParsedAuto
           })
         }
         if (key in theme) {
-          parts.push({
-            type: 'group',
-            values: Object.keys(theme[key]),
-          })
+          if (attrs.includes('deep')) {
+            parts.push({
+              type: 'deepgroup',
```
antfu, line 29: Don't we just enumerate the combinations add a group type so we don't need to have more logics to handle afterwards?

packages/preset-mini/src/autocomplete.ts
```diff
@@ -5,6 +5,6 @@ export const autocomplete: AutoCompleteTemplate[] = [
   '(border|b)-(rounded-|rd-)#num',
   '(border|b)-$colors',
   '(m|p)(x|y|t|b|l|r|s|e|)-#num',
-  'text-$colors',
+  'text-$colors:deep',
```
antfu, line 8: We probably don't need the `:deep`, we could always going deep by check if the child is an object or not

packages/autocomplete/src/parse.ts
QiroNT: idk since enumerating the combinations is sort of like doing a `listAll`.

packages/preset-mini/src/autocomplete.ts
QiroNT: Sure, could be done by default.

packages/autocomplete/src/parse.ts
QiroNT: `text-` generates ~700 color suggestions and it's the only one nesting so I'm still not sure, you can merge it and copy the `resolve` function in `listAll` part to get the group.

PR #731, opened 2022-03-18
fix(preset-mini): surrpot bg image & size prop

chris-zhu:
Align with tw usage.

`background-image` & `background-size` in `preset-wind` before.
I found this rule (`/^bg-(.+)$`)  in `preset-mini`, it will intercept all my input.

So, I move `image` & `size` to `preset-mini`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@unocss/monorepo",
-  "version": "0.29.1",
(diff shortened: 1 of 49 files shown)


antfu:
Let's have it in the preset wind, `[x: ]` is tailwind specific syntax

PR #734, opened 2022-03-18
feat(autocomplete)!: extractor support & attributify

QiroNT:
Needs more advices on this.

This is the only implementation that I could think of, but it's anything but pretty. Since transforms need to be bi-directional, It's somewhat hard to implement, still couldn't find a library to do this. And multi-transforms also needs to be able to implement which the current approach doesn't.


QiroNT:
Playground is now supported, feel free to try it out. Extension is not tested tho.

packages/core/src/types.ts
```diff
@@ -309,11 +309,68 @@ export interface ConfigBase<Theme extends {} = {}> {
    * Custom function for auto complete
    */
   autocomplete?: (AutoCompleteFunction | AutoCompleteTemplate)[]
+
+  /**
+   * Custom extractors for auto complete
+   */
+  autocompleteExtractors?: AutoCompleteExtractor[]
```
antfu, line 316: We could merge the interface with `autocomplete`

antfu (approved):

Issue #755, opened 2022-03-24
Support Arbitrary values for background image


winter-ice:
ref: https://tailwindcss.com/docs/background-image

```
<div class="bg-[url('/img/hero-pattern.svg')]">
  <!-- ... -->
</div>
```

PR #757, opened 2022-03-25
feat(preset-wind): suppport bg image & size prop

chris-zhu:


chris-zhu:
close #755

chris-zhu:
This rule generate right css, but test not pass. I don't why, maybe need some help.

PR #766, opened 2022-03-26
fix(spacing): fix DEFAULT spacing mismatch

PentaTea:
The padding and margin spacing are [written here](), but  the `+` is used in the second capture group of the rule,it can never be empty

This means that [the DEFAULT in spacing](https://github.com/unocss/unocss/blob/4981da7d9df42900a38db1fd18cf52ef4a17ad32/packages/preset-mini/src/theme/misc.ts#L18) cannot be matched

I've replaced the `+` with the `*`

before:

image

after:

image


PentaTea:
It seems that the first commit broke the parsing of `ma`

image

I made the second modification. It should be OK this time

image

PentaTea:
I broke this branch... I'll PR it again

PR #767, opened 2022-03-26
fix(spacing): fix DEFAULT spacing mismatch

PentaTea:
The padding and margin spacing are [written here](https://github.com/unocss/unocss/blob/main/packages/preset-mini/src/rules/spacing.ts), but  the `+` is used in the second capture group of the rule,it can never be empty

This means that [the DEFAULT in spacing](https://github.com/unocss/unocss/blob/main/packages/preset-mini/src/theme/misc.ts#L18) cannot be matched

I replaced the `+` with the `*`

before:

image

after:

image

### DIFF
--- a/packages/preset-mini/src/rules/spacing.ts
+++ b/packages/preset-mini/src/rules/spacing.ts
@@ -4,25 +4,17 @@ import { directionSize } from '../utils'
 export const paddings: Rule[] = [
   [/^pa?()-?(-?.+)$/, directionSize('padding'), { autocomplete: ['(m|p)<num>', '(m|p)-<num>'] }],
   [/^p-?xy()()$/, directionSize('padding'), { autocomplete: '(m|p)-(xy)' }],
-  [/^p-?([xy])-?(-?.+)$/, directionSize('padding')],
-  [/^p-?([xy])()$/, directionSize('padding')],
-  [/^p-?([rltbse])-?(-?.+)$/, directionSize('padding'), { autocomplete: '(m|p)<directions>-<num>' }],
-  [/^p-?([rltbse])()$/, directionSize('padding')],
-  [/^p-(block|inline)-(-?.+)$/, directionSize('padding'), { autocomplete: '(m|p)-(block|inline)-<num>' }],
-  [/^p-(block|inline)()$/, directionSize('padding')],
-  [/^p-?([bi][se])-?(-?.+)$/, directionSize('padding'), { autocomplete: '(m|p)-(bs|be|is|ie)-<num>' }],
-  [/^p-?([bi][se])()$/, directionSize('padding')],
+  [/^p-?([xy])(?:-?(-?.+))?$/, directionSize('padding')],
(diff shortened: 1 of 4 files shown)


antfu:
IMO `p-` shouldn't be matched. And `p` would be too general to me.

PentaTea:
> IMO `p-` shouldn't be matched. And `p` would be too general to me.

You are right. I'll fix `p-`

But `pa` can no longer be used as default padding, We need a better marker

Compared to `<div px py>`, How about `pxy`?

chris-zhu:
> > IMO `p-` shouldn't be matched. And `p` would be too general to me.
> 
> You are right. I'll fix `p-`
> 
> But `pa` can no longer be used as default padding, We need a better marker
> 
> Compared to `<div px py>`, How about `pxy`?
> 
> 
> 
> 
> 
> 
> 

pxy  ===  p？

PentaTea:
> > > IMO `p-` shouldn't be matched. And `p` would be too general to me.
> > 
> > 
> > You are right. I'll fix `p-`
> > But `pa` can no longer be used as default padding, We need a better marker
> > Compared to `<div px py>`, How about `pxy`?
> 
> pxy === p？

Yes, pxy === "padding: <DEFAULT>"

PentaTea:
@antfu I banned the use of `p`,`p-`,`pb-` and used `pxy` instead of `px py`

PS. Can I ask if force push will affect PR?

Before:
image
After:
image

antfu:
> if force push will affect PR?

No, feel free to. But actually no need if you mean to clean up the commits since it will be squashed anyway

packages/preset-mini/src/rules/spacing.ts
```diff
@@ -7,6 +7,11 @@ export const paddings: Rule[] = [
   [/^p-?([rltbse])-?(-?.+)$/, directionSize('padding'), { autocomplete: '(m|p)<directions>-<num>' }],
   [/^p-(block|inline)-(-?.+)$/, directionSize('padding'), { autocomplete: '(m|p)-(block|inline)-<num>' }],
   [/^p-?([bi][se])-?(-?.+)$/, directionSize('padding'), { autocomplete: '(m|p)-(bs|be|is|ie)-<num>' }],
+  [/^p-?([xy])()$/, directionSize("padding")],
+  [/^p-?([rltbse])()$/, directionSize("padding")],
+  [/^p-(block|inline)()$/, directionSize("padding")],
+  [/^p-?([bi][se])()$/, directionSize("padding")],
+  [/^p-?xy()()$/, directionSize("padding")]
 ]
```
chu121su12, line 15: Please move the added rules next to the original to allow better override

ex:
```ts
  [/^p-?([xy])-?(-?.+)$/, directionSize('padding')],
  [/^p-?([xy])()$/, directionSize("padding")],
  [/^p-?xy()()$/, directionSize("padding")],
  // then
  [/^p-?([rltbse])-?(-?.+)$/, directionSize('padding'), { autocomplete: '(m|p)<directions>-<num>' }],
  /// ...
```

Also, please provide some test.

antfu (approved):

PR #770, opened 2022-03-26
docs: update benchmark

Marr11317:
I was wondering how things evolved. The benchmard was from november, so I wondered if it had regressed. 

Turns out it has not. It is still faster than all others by a very very faire amount.

Note that I ran this update on a github codespace, and it seems to be slower than your machine. This means that the numbers in the benchmark are greater when comparing time. However, when comparing one with the other, the speed indexes stay about the same. These numbers are probably what would be obtained if this was periodically run on CI.


packages/inspector/client/auto-imports.d.ts
```diff
@@ -39,7 +39,6 @@ declare global {
   const makeDestructurable: typeof import('@vueuse/core')['makeDestructurable']
   const markRaw: typeof import('vue')['markRaw']
   const nextTick: typeof import('vue')['nextTick']
-  const note: typeof import('@vueuse/core')['note']
```
Marr11317, line 42: Note that I have no idea why this was modified. It was done automatically. I can remove it from the PR if you prefer.

Issue #796, opened 2022-04-04
Is there a feature in DevTools that allows me to modify a class directly and see how it behaves?


oekazuma:
I really like the ability in Windi CSS to change classes directly in DevTools to see how they work.

Does Uno CSS have this feature?

Reference: https://windicss.org/integrations/vite.html#design-in-devtools

antfu:
Not yet, since I don't design in that way. I am open for PR if anyone is interested to port it back to UnoCSS as a package.

PR #812, opened 2022-04-09
feat(vite): Design in DevTools

action-hong:
#796 

Just add the following lines to your main entry
```ts
import 'uno.css'
import 'virtual:unocss-devtools'
```

动画

---
- https://windicss.org/integrations/vite.html#design-in-devtools
- https://github.com/windicss/vite-plugin-windicss/blob/main/packages/vite-plugin-windicss/src/devtools.ts


antfu:
Can you also add a section to the readme of the vite package? Thanks

PR #824, opened 2022-04-13
fix(preset-mini): convert to static rules

chris-zhu:


chris-zhu:
Converting to static rules is friendly to autoComplete

Issue #826, opened 2022-04-13
@apply classes will not update when config updated


kaiser-9527:
// unicss.config.ts
```ts
{
    theme: {
        colors: {
            primary: '#333333'
        }
    }
}
```

// components

```css
.title {
    @apply text-primary;
}
```

when i update the primary color to `#666666`, the title text color is still `#333333`

TrickyPi:
I don't run into this issue when i config the theme in vite.config.ts.
It's my vite.config.ts
```ts
import { defineConfig } from 'vite'
import Unocss from 'unocss/vite'
import react from '@vitejs/plugin-react'
import tsconfigPaths from 'vite-tsconfig-paths'
import transformerDirectives from '@unocss/transformer-directives'
// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    react(),
    Unocss({
      transformers: [transformerDirectives()],
      theme: {
        colors: {
          primary: '#333333'
        }
      }
    }),
    tsconfigPaths(),
  ],
})
```

I guess the problem is that the change of uno.config.js didn't active vite/(other build tools) recompile, but I think if you need live updates when the theme is updated, it's more appropriate to config this in vite.config.ts.

antfu:
We temporarily close this due to the lack of enough information.
Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) to reopen the issue.
Thanks.

> [Why reproduction is required](https://antfu.me/posts/why-reproductions-are-required)

TrickyPi:
this #837 pr has been reverted in #917

PR #837, opened 2022-04-17
fix(vite): update @apply when config changed

action-hong:
#826


action-hong:
wait it still have some problem

PR #848, opened 2022-04-20
feat(preset-icons): add `unit` option

### LINKED ISSUE #842: How to use `rem` unit for preset icon scale option?
Hi, I want to use the `rem` unit for the icon preset's scale instead of the `em` unit.

Currently, I see no option to use `rem` instead of `em`.

chris-zhu:
Can I do this ? Maybe resolved #842


packages/preset-icons/src/index.ts
```diff
@@ -95,6 +96,11 @@ export const preset = (options: IconsOptions = {}): Preset => {
           return
         }
 
+        if (unit) {
```
userquin, line 99: this should be done with custom `iconCustomizer` like on `unplugin-icons` and only if the w/h on props are missing:
- the logic to apply will be something similar to this https://github.com/iconify/iconify/blob/master/packages/utils/src/loader/modern.ts#L42 only for width and height props
- you need to provide a custom `iconCustomizer` like this (with previous logic after provided `iconCustomizer`): https://github.com/antfu/unplugin-icons/blob/main/src/core/loader.ts#L78

userquin: we also need to provide a test to check the w/h aren't overridden

userquin: @antfu maybe it will be better send it upstream

packages/preset-icons/src/types.ts
```diff
@@ -61,4 +61,10 @@ export interface IconsOptions {
    * @default false
    */
   autoInstall?: boolean
+  /**
+   * Custom your Icon unit.
+   *
+   * @default `em`
+   */
+  unit?: string
```
userquin, line 69: cannot definea type with allowed values: `'em' | 'rem' ...`

packages/preset-icons/src/types.ts
```diff
@@ -52,7 +54,7 @@ export interface IconsOptions {
   /**
    * Custom icon customizations.
    */
-  customizations?: Omit<IconCustomizations, 'additionalProps' | 'trimCustomSvg'>
+  customizations?: Omit<IconCustomizations, 'additionalProps' | 'trimCustomSvg' | 'iconCustomizer'>
```
userquin, line 57: remove this, the type above, also `iconCustomizer` entry below, it must be done inside `customizations` option

userquin:
@chris-zhu 

I think you did not understand me or I explained myself badly, what you have to do is add the unit and then add a custom `iconCustomizer` to add the w/h if it is not configured, but for this we first call the one provided by the user and then we configure the width and height if the user has not configured them before in their customizations:

```ts
             async iconCustomizer(collection, icon, props) {
               await providedIconCustomizer?.(collection, icon, props)
               if (unit) {
                 if (props.width !== undefined && props.width !== null) {
                   props.width = `${scale}${unit}`
                 }
                 if (props.height !== undefined && props.height !== null) {
                   props.height = `${scale}${unit}`
                 }
               }
             },
```

packages/preset-icons/src/index.ts
```diff
@@ -70,6 +71,16 @@ export const preset = (options: IconsOptions = {}): Preset => {
             ...customizations,
             additionalProps: { ...extraProperties },
             trimCustomSvg: true,
+            async iconCustomizer(collection, icon, props) {
+              await customizations.iconCustomizer?.(collection, icon, props)
+              if (unit) {
+                if (props.width !== undefined && props.width !== null)
+                  props.width = `${scale}${unit}`
```
userquin, line 78: Should override value when undefined or null, also for height

test/preset-icons.test.ts
```diff
@@ -17,9 +17,33 @@ describe('preset-icons', () => {
...
   })
 
+  const unoWithUnit = createGenerator({
+    presets: [
+      presetIcons({
+        unit: 'rem',
+        scale: 2,
+        customizations: {
+          iconCustomizer(collection, icon, props) {
+            if (collection === 'carbon' && icon === 'sun') {
```
userquin, line 27: Update this to change the rest of icons and so this test allow us to check it is not overridden (once logic updated)

userquin: Just negate the if condition

chris-zhu:
@userquin Thanks so much!

packages/preset-icons/src/types.ts
```diff
@@ -61,4 +61,8 @@ export interface IconsOptions {
    * @default false
    */
   autoInstall?: boolean
+  /**
+   * Custom icon unit.
+   */
+  unit?: string
```
userquin, line 67: add the default value: 
```jsdoc
@default `em`
```

userquin (approved):

PR #852, opened 2022-04-20
docs: add `safelist` option

zguolee:


antfu:
I think it's better to move it after the `### Scanning` section as the safelist only make sense when we have the concept of static scanning.

zguolee:
Ok, I have done.

Issue #860, opened 2022-04-22
intelligence of  vscode extension doesn't work


jeffrey-bin:
v0.31.12 version of vscode extension not working 
![image](https://user-images.githubusercontent.com/23428596/164701831-156732b1-c2e6-4779-bac2-d5ad67c57b70.png)

but,

v0.31.10 version of vscode extension does work
![image](https://user-images.githubusercontent.com/23428596/164702003-8e277afa-7f01-4f41-b0f4-acd58259b7e1.png)

both screen captures are from the same repo.

I retrive the commit after v0.31.10 , maybe this commit destroy the detection:
https://github.com/unocss/unocss/commit/95a025d99e38bf9d2c5f48fc85e688eec619b94a#diff-13e315a8c1fa669f41b67be102e3860fdf8dc12bfff6d9792552f02862e18729L30

PR #861, opened 2022-04-22
fix(config): resolved cwd

lvjiaxuan:
#860


antfu:
Fixed in https://github.com/unocss/unocss/commit/12af59e398db30c6daee6add258d4655b92730f3. Thanks

PR #863, opened 2022-04-23
Fix/oblique and font synthesis

UltraCakeBakery:
These styles are on my list for my person preset. Maybe they better belong in preset-mini

https://www.w3.org/TR/css-fonts-3/#font-style-prop
https://www.w3.org/TR/css-fonts-3/#propdef-font-synthesis


UltraCakeBakery:
@antfu  the logs only show the prefix `error: ` without any actual message. Locally the test passed after updating the snapshot. I am wondering now though if the snapshot that is in my commit should actually be effected by my other changes.

chu121su12:
Please add tests for the features. Also you can fix the snapshots using `pnpm test:update`.

UltraCakeBakery:
Something out of my control is going wrong with the workflows. I reset this branch to `main`, added tests and updated the snapshots (see my latest commits). The tests should not be failing. I do not know what is going wrong.

Is it because I did a force push to the branch? If it helps, I happily create a new PR with the same changes. 
I'm experiencing dejavu from my previous PR attempt 😅

UltraCakeBakery:
@antfu Thank you!

PR #872, opened 2022-04-25
fix(cli): cli don't work

deane0:


johannschopplich:
My fault. Thanks for the PR, @deane0!

PR #881, opened 2022-04-27
feat(vite): support unocss can be processed by vite:css

action-hong:
in development mode, __uno.css can be processed by `vite:css`, but not in production

add `usePostcss` flag to make it works, so we can use `postcss` to process the generated css

```ts
// vite.config.ts
 Unocss({
     usePostcss: true
  }),
```

### DIFF
--- a/packages/vite/src/types.ts
+++ b/packages/vite/src/types.ts
@@ -28,4 +28,10 @@ export interface VitePluginConfig<Theme extends {} = {}> extends UserConfig<Them
    * @default false
    */
   transformCSS?: boolean | 'pre' | 'post'
+  /**
+   *
+   * make the generate css processed by postcss(https://vitejs.dev/guide/features.html#postcss)
+   * @default false
+   */
+  usePostcss?: boolean
 }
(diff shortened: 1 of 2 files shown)


packages/vite/src/types.ts
```diff
@@ -28,4 +28,10 @@ export interface VitePluginConfig<Theme extends {} = {}> extends UserConfig<Them
    * @default false
    */
   transformCSS?: boolean | 'pre' | 'post'
+  /**
+   *
+   * make the generate css processed by postcss(https://vitejs.dev/guide/features.html#postcss)
+   * @default false
```
antfu, line 34:
```suggestion
   * @default true
```

I guess we could enable this by default to align the dev and build

action-hong:
@antfu 

just find a problem with this PR
https://github.com/unocss/unocss/blob/4d5a18e3c35f64d041a27dfe9a3668d47a38e5ae/packages/vite/src/modes/global/build.ts#L130-L142

all css chunk expect unocss will be processed by postcss again(twice), 

haven't figured out how to better resolved generate css only

action-hong:
```ts
import replaceAsync from 'string-replace-async'

...
const css = chunk.source
  .replace(HASH_PLACEHOLDER_RE, '')

chunk.source = await replaceAsync(css, LAYER_PLACEHOLDER_RE, async(_, __, layer) => {
  replaced = true
  return await transformCSS(layer === LAYER_MARK_ALL
    ? result.getLayers(undefined, Array.from(vfsLayerMap.values()))
    : result.getLayer(layer) || '', `${chunk.fileName}.css`)
})
```

PR #897, opened 2022-04-29
feat(reset): stable scrollbar gutter

kecrily:


wheatjs:
Is this in the official tailwind reset? If not I'm wondering if we should add this? We probably want to keep the resets 1:1 since this reset is called the tailwind reset.

kecrily:
I don't think anyone would like a layout that changes because of the presence or absence of scrollbars. It doesn't change anything other than keeping the style consistent with or without scrollbars.

Of course if we really need to keep 1:1 with upstream, then I can also submit a PR to tailwind.

antfu:
Yes. The file is aim to align with Tailwind's. We could have it after Tailwind, or feel free to maintain your custom resets. Closing for now. Thanks.

PR #899, opened 2022-04-29
feat(preset-attributify): new `trueToNonValued` option

### LINKED ISSUE #876: [preset-uno & preset-attributify] weird behaviour of text-transform utilities
So, I'm using `preset-uno` and `preset-attributify` (with TypeScript) and want to make my text `uppercase`. The only way do it without errors is using `class="uppercase"`. When I try to specify just the `uppercase` attribute on the tag, it actually works on the page, but I get `TypeScript` error telling me that there's no such attribute (because it's not specified [here](https://github.com/unocss/unocss/blob/d515bc677b4ccd1d3b7a0fa1c8f764dc9015f28b/packages/preset-attributify/src/jsx.ts#L1) I guess). By the way, the behaviour is the same with `lowercase` and `capitalize`.

And that's not all. Somewhere here I've read that there's also `case-` attributes (`case-upper`, `case-lower`, etc.). When I try to use them, I receive no type errors, but they don't affect the page (e.g. don't work).

### LINKED ISSUE #878: [bug] Attribute mode doesn't work with SolidJS components
I have a component:
```ts
import { Component, JSX } from "solid-js";

export type CardProps = JSX.HTMLAttributes<HTMLDivElement>;

export const Card: Component<CardProps> = (props) => {
  return (
    <div p-4 bg-white rounded-2xl {...props}>
      {props.children}
    </div>
  );
};
```

It extends Solid's `JSX.HTMLAttributes`, so it can receive UnoCSS attributes. But when I try to use them with the component, they don't work (although they are accepted as valid props by TypeScript).

In other words, this works:
```ts
<Card class="h-full text-center group-hover:shadow-lg group-active:shadow-none">
  content
</Card>
```
But this doesn't:
```ts
<Card h-full text-center group-hover="shadow-lg" group-active="shadow-none">
  content
</Card>
```

In the latter case, the attributes are present in HTML, but the styles for them are not generated (note that the attributes specified in the root tag of the `Card` component (e.g. `p-4 bg-white rounded-2xl` do have generated styles).

A person in Solid's discord suggested the following:
> Most likely it's failing because uno is running too late, after solid's transform.
Classes work because it looks like uno just looks for strings, so it doesn't matter what solid does, the strings are still intact: https://github.com/unocss/unocss/blob/main/packages/core/src/extractors/split.ts
But attributify fails because it's looking for tags, which are already transformed into `createComponent` calls: https://github.com/unocss/unocss/blob/main/packages/preset-attributify/src/extractor.ts

I tried changing Vite config: put `solidPlugin` after `unocssPlugin` and set `enfore: "post"` for `solidPlugin`. It didn't help.

XiNiHa:
Closes #878
This slightly increases the size of CSS even when not using it, probably providing an option for it could be nice.


antfu:
This is probably not a the correct solution as the key will affects the content
https://codepen.io/antfu/pen/oNEvogp

XiNiHa:
Right, do you want to add an exception for `true` on `variantAttributify()` along with `~` then?

antfu:
Can you share some examples?

XiNiHa:
In https://github.com/XiNiHa/unocss/blob/a08618de72347a55b4f4eaf2338d2278ddcc0288/packages/preset-attributify/src/variants/attributify.ts#L26
`if (body === '~' || !body)` to `if (body === '~' || body === 'true' || !body)`
This will break rules that ends with `true`, but I think it's not that bad solution if we add an option for this.

antfu:
Ok I am fine to give it a try

XiNiHa:
Fixed

packages/preset-attributify/src/types.ts
```diff
@@ -36,4 +36,13 @@ export interface AttributifyOptions extends PresetOptions {
    * A list of attributes to be ignored from extracting.
    */
   ignoreAttributes?: string[]
+
+  /**
+   * Non-valued attributes will also match if the actual value represented in DOM is `true`.
+   * This option exists for supporting frameworks that encodes non-valued attributes as `true`.
+   * Enabling this option will break rules that ends with `true`.
+   *
+   * @default true
```
antfu, line 45:
```suggestion
   * @default false
```

I think it's better to default to false for a smaller bundle size. We could document this better to turn on when working with Solid.

XiNiHa:
I didn't add tests for the new option since the existing test script structure doesn't look suitable for this.

PR #902, opened 2022-04-29
feat(preset-wind): add @supports feature query

kecrily:


antfu:
Can you add some tests to them? Thanks

kecrily:
I've run into some problems and I may need to rethink and take a closer look at the entire code.

I closed it first, but of course it would be great if someone else could help implement it.

chu121su12:
@kecrily Maybe try to check out `@media` variant support

PR #904, opened 2022-04-29
docs: fix display width

Allen-1998:
###  The documentation was incredible, but I found a small bug .when there is an scroll-y, using a width of 100vw will produce an scroll-x 

image

### old:
image

### now:
image

### DIFF
--- a/interactive/app.vue
+++ b/interactive/app.vue
@@ -13,7 +13,7 @@ useHead({
 </script>
 
 <template>
-  <main class="pt2 text-center font-sans" w-100% h-screen of-hidden relative>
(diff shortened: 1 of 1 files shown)


interactive/app.vue
```diff
@@ -13,7 +13,7 @@ useHead({
 </script>
 
 <template>
-  <main class="pt2 text-center font-sans" w-screen h-screen of-hidden relative>
+  <main class="pt2 text-center font-sans" w-100% h-screen of-hidden relative>
```
antfu, line 16:
```suggestion
  <main class="pt2 text-center font-sans" w-full h-screen of-hidden relative>
```

PR #937, opened 2022-05-05
chore: add title for TheNav icon

Simon-He95:
it will better understand the role of icons

### DIFF
--- a/interactive/components/TheNav.vue
+++ b/interactive/components/TheNav.vue
@@ -40,9 +40,9 @@ import { currentTab, isCompact, toggleCompact, uno } from '~/composables/state'
         />
       </template>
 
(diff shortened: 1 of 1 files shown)


interactive/components/TheNav.vue
```diff
@@ -39,9 +40,9 @@ import { currentTab, isCompact, toggleCompact, uno } from '~/composables/state'
         />
       </template>
 
-      <a i-carbon-template aria-label="Playground" href="https://unocss.antfu.me/play/" target="_blank" />
+      <a i-carbon-template aria-label="Playground" href="https://unocss.antfu.me/play/" title="Playground" target="_blank" />
```
userquin, line 43: can you also add `rel="noopener noreferrer"` to both anchors (Playground and Github)

Simon-He95: of course

Issue #939, opened 2022-05-05
`,` escape breaks svelte scoping


alexanderniebuhr:
Using a class like `grid-cols-[repeat(6,1fr)]`, unocss will escape `,` in the selector with `\2c`. This works as a style, but it will not work with svelte's check for unused styles in scope mode.
Therefore the style is removed on build time, leading to missing style.

Just using the `\,` to escape the comma (as seen in the example) will make the style work and svelte actually understand that the style is used.

Actual Output:
https://stackblitz.com/edit/web-platform-1asmuc?file=styles.css

Expected Output:
https://stackblitz.com/edit/web-platform-akerk4?file=styles.css

PR #941, opened 2022-05-05
fix: escape `,` properly

alexanderniebuhr:
This PR updates the escape char of `,` to be `\,` instead of `\2c `. I suggest this change because using uno with svelte scoping breaks. Svelte removes unused styles (if scoped), and does not recognise the `\2c ` as the same class in the markup.

I tested the browser support (using online test grids) for the example of `grid-cols-[repeat(6,1fr)]`

**Desktop** (https://github.com/alexanderniebuhr/lambda-test/blob/main/TES100201621651772638589250.pdf)
- Chrome >59 ✅ 
- Firefox >58 ✅ 
- IE 11 ❌ 
- Safari >12 ✅ 

**Mobile** (https://github.com/alexanderniebuhr/lambda-test/blob/main/TES100201621651774781769360.pdf)
- iPhones
- multiple Android Phones

Everything seems to work, except IE. But the browser does not support grid, so I expect that. However IE should work with supported rules with `,` in them.

According to newest brower's dev tools, the generated css class is the same with both escape chars (see attached screenshot).

Screenshot 2022-05-05 at 20 03 43

closes https://github.com/unocss/unocss/issues/939

### DIFF
--- a/packages/core/src/utils/escape.ts
+++ b/packages/core/src/utils/escape.ts
@@ -27,7 +27,7 @@ export function escapeSelector(str: string): string {
 
     // Comma
     if (codeUnit === 44) {
-      result += '\,'
+      result += '\\,'
       continue
     }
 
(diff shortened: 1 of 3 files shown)


alexanderniebuhr:
I am more than happy to do some more test, if requested with some test items.

PR #945, opened 2022-05-07
feat(reset): update from tailwind reset

chu121su12:
I just noticed there's new flex reset here. Should the `.dark` block be removed in the tailwind reset?


antfu:
> Should the .dark block be removed in the tailwind reset?

Sure I am fine with it.

PR #946, opened 2022-05-07
docs(interactive): fix firefox

userquin:
On Chromium browsers there is no problem but on Firefox trailing commas throws error (not tested on Safari)

I need to fix it on playground, also failing


userquin:
maybe we can fix this here https://github.com/unocss/unocss/blob/main/packages/shared-docs/src/config.ts#L16

about playground we also need to fix it before using `JSON.parse`

antfu:
Maybe we could use prettier to format the code instead of using regex

PR #950, opened 2022-05-08
feat(compile-class): new transformer

### LINKED ISSUE #948: Feature request: A more useful compilation mode
Now that unocss can do transforms, I am wondering how easy it would be to add support for a more useful `compilation mode` compared to the one from [windicss](https://windicss.org/posts/modes.html). One where you can also manually pick what elements get their classes and attributes combined into one.

###### EXAMPLE:

**Before:**
```html
<!-- note the additional `compile` class at the beginning -->
<div class="compile bg-red-500 text-xl font-bold border border-gray-200 dark:hover:bg-green-500 transform scale-5">
``` 
**After:**
```html
<div class="uno-2xjw9">
```

antfu:
close #948

[rendered](https://github.com/unocss/unocss/blob/feat/transformer-compile-class/packages/transformer-compile-class/README.md)


jd-solanki:
Can this kind of compilation somehow help in generating style for dynamically classes like from prop `bg-${props.color}` without safelist?

PR #956, opened 2022-05-10
feat(vscode): support loading from iles.config file

YuukanOO:
Since I'm using the wonderful [iles](https://iles.pages.dev/) static site generator and unocss vscode extension did not support loading from the `iles.config.ts` (or js) file, this PR solves it since the format for the vite configuration is the same as other config files (a `vite` property).

For the extension to work correctly, I did have to explicitly defines the presets to use, without it, it did not find utilities, maybe the default preset is not applied 🤔

### DIFF
--- a/packages/vscode/README.md
+++ b/packages/vscode/README.md
@@ -17,7 +17,7 @@
 ## Features
 
 - Decoration and tooltip for matched utilities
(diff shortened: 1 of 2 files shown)


YuukanOO:
Hmmm, looks like the latest version `0.33.4` as per the `package.json` has not be deployed yet?

PR #961, opened 2022-05-12
fix(vite): emit better warning in build

Dunqing:
[unocss:global:build:generate] [unocss] does not found CSS placeholder in the generated chunks,
this is likely an internal bug of unocss vite plugin

If I don't import uno.css, I get this error during the build phase.

### DIFF
--- a/packages/vite/src/modes/global/build.ts
+++ b/packages/vite/src/modes/global/build.ts
@@ -119,8 +119,11 @@ export function GlobalModeBuildPlugin({ uno, ready, extract, tokens, modules, fi
       enforce: 'post',
       // rewrite the css placeholders
       async generateBundle(_, bundle) {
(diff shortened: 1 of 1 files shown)


antfu:
> If I don't import uno.css, I get this error during the build phase.

What's the point if you don't import uno.css but install the plugin?

Dunqing:
The problem is mainly in build watch mode, I probably deleted the imported "uno.css" and added it later, so I shouldn't get this error.

Maybe we should improve the error message?

antfu:
> Maybe we should improve the error message?

Sounds good to me

Dunqing:
I thought it would be more appropriate to add a warning, avoid user think `unocss` invalid and build failure

So I added the same warning as in development mode to remind the user to import `uno.css`.

What do you think?

PR #965, opened 2022-05-12
chore: capital letters

chris-zhu:


antfu:
Thanks but actually it's intentional as I treat them as a part of the whole sentence.

PR #966, opened 2022-05-12
docs(playground): missing icon preset

chris-zhu:
image
I thought it was intentional, but found that the `presetIcons` was missing

-----------------------

And: 
image

In my local, i don't understand why some properties are missing in the browser, although they are correct when I debug


userquin:
@chris-zhu IIRC the playground is only for browser, we don't need preset icons on node, the configuration for the browser is on this package [`packages/shared-docs`](https://github.com/unocss/unocss/blob/main/packages/shared-docs/src/defaultConfig.ts)

check on private window, we're using `cdn` to download the icon collections, so it can take a while

antfu (approved):

PR #969, opened 2022-05-13
feat(core): merge safelist presets from config

### LINKED ISSUE #968: Safelist within Presets not working
Try creating a custom preset and there is a `safelist` define within the preset.
When we implement the preset the safelist classes are not working.

# Reproduction

https://github.com/praburangki/reproduce-unocss

praburangki:
The main reason is if someone wants to build a UI library / presets on top of unocss we can define the classes that will always be generated

Close #968

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -30,7 +30,7 @@ export function resolveConfig(
 
   const layers = Object.assign(defaultLayers, ...rawPresets.map(i => i.layers), userConfig.layers)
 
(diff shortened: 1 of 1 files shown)


packages/core/src/config.ts
```diff
@@ -73,7 +73,7 @@ export function resolveConfig(
     mergeSelectors: true,
     warn: true,
     blocklist: [],
-    safelist: [],
+    safelist: mergePresets('safelist'),
```
antfu, line 76: You might need to move this under `...config`

praburangki: Ah right, I missed that part thanks.

PR #991, opened 2022-05-20
fix(core): revert handler insertion order

chu121su12:
The reverse order may be unnecessary since the flattened selector are forced to be reordered via the new helper. It also breaks the pseudo function usecase.

Another option is to hardcode pseudo elements to have higher `order` so as to be moved to the end of selector list (just like the `:active` pseudo [[diff]](https://github.com/unocss/unocss/compare/main...chu121su12:variant-ordering-1?expand=1#diff-72ab94343d47e2266a978d9981390f8bb9a6546dbdf17d0727b6a6687a769ca4))

/cc @sibbng Please correct me if I missed something

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -298,7 +298,7 @@ export class UnoGenerator {
         processed = handler.matcher
         if (Array.isArray(handler.parent))
           this.parentOrders.set(handler.parent[0], handler.parent[1])
-        handlers.unshift(handler)
+        handlers.push(handler)
         variants.add(v)
         applied = true
         break
--- a/test/__snapshots__/order.test.ts.snap
+++ b/test/__snapshots__/order.test.ts.snap
@@ -4,16 +4,16 @@ exports[`order > movePseudoElementsEnd 1`] = `".marker\\\\:file\\\\:hover\\\\:se
 
 exports[`order > multiple variant sorting 1`] = `
 "/* layer: default */
-.dark .group:hover:focus-within .dark\\\\:group-hover\\\\:group-focus-within\\\\:bg-blue-600{--un-bg-opacity:1;background-color:rgba(37,99,235,var(--un-bg-opacity));}
-.group:hover:focus-within .dark .group-hover\\\\:group-focus-within\\\\:dark\\\\:bg-red-600{--un-bg-opacity:1;background-color:rgba(220,38,38,var(--un-bg-opacity));}"
+.dark .group:focus-within:hover .group-hover\\\\:group-focus-within\\\\:dark\\\\:bg-red-600{--un-bg-opacity:1;background-color:rgba(220,38,38,var(--un-bg-opacity));}
+.group:focus-within:hover .dark .dark\\\\:group-hover\\\\:group-focus-within\\\\:bg-blue-600{--un-bg-opacity:1;background-color:rgba(37,99,235,var(--un-bg-opacity));}"
 `;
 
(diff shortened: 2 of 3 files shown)


sibbng:
> The reverse order may be unnecessary since the flattened selector are forced to be reordered via the new helper. It also breaks the pseudo function usecase.

Yes, it's changing pseudo functions' behaviour. But that's how it was supposed to work from day one. 

Let's say you have those elements on your Tailwind project:
```html
<div class="dark:group-hover:group-focus-within:bg-blue-600"></div>
<div class="hover:not-first:checked:bg-red-600"></div>
```
Tailwind doesn't support `not-` variants yet but there is open PR for it. If you migrate from Tailwind to UnoCSS, to get this code working you need to move `dark:` variant to the end of group utilities. In the second example, you need to swap `hover` and `checked`.

> Another option is to hardcode pseudo elements to have higher `order` so as to be moved to the end of selector list (just like the `:active` pseudo [[diff]](https://github.com/unocss/unocss/compare/main...chu121su12:variant-ordering-1?expand=1#diff-72ab94343d47e2266a978d9981390f8bb9a6546dbdf17d0727b6a6687a769ca4))

I'm totally okay with that but the diff you are linked to doesn't remove `movePseudoElementsEnd` from the code and it seems to order all pseudos instead of just pseudo elements.

sibbng:
Correction: in the second example you don't need to swap `hover` and `checked` to get it working. The order of those pseudo classes doesn't matter. I took them from tests and I thought you were talking about them by saying _"breaks the pseudo function usecase"_.

chu121su12:
The pseudo functions I was referring to are the pseudos that looks like function `:not()`. Please excuse my english.

The change in this commit is only on the `handlers.unshift`. What I understand in the original `push` implementation is so that the token composition will match the generated class, except when the variant specify the `order` key.

---

> Tailwind doesn't support `not-` variants yet but there is open PR for it. If you migrate from Tailwind to UnoCSS, to get this code working you need to move `dark:` variant to the end of group utilities.

The `dark/light` variants should be able to be moved around by the `order` key too.

https://github.com/unocss/unocss/blob/2717fcb725c3370b6062484e335a0c4c70feb616/test/order.test.ts#L31-L61

---

> I'm totally okay with that but the diff you are linked to doesn't remove `movePseudoElementsEnd` from the code 

I was tinkering with the diff branch first before I realized the change can be just with a single line as is in this PR. 

---

> it seems to order all pseudos instead of just pseudo elements.

The order condition is only applied to the direct pseudo element usage. I havent't explore their use with other combinations.

https://github.com/unocss/unocss/blob/2ebd299a6b94c3b141c54a7d81b9c3763f92a04b/packages/preset-mini/src/variants/pseudo.ts#L109-L125

sibbng:
> The pseudo functions I was referring to are the pseudos that looks like function :not(). Please excuse my english.

I know, we're talking about the same thing, I picked up this test because it's including `:not()`. I just haven't come up with any use case that breaks them.

> The dark/light variants should be able to be moved around by the order key too.

This forces them to be prioritized no matter they are defined in the class. This is not what we want. It might make for some variants, but it doesn't for `dark` variant.

image

I'm okay with the future refactors but snapshots shouldn't change.

sibbng:
@chu121su12 I guess you misunderstood my last words. I meant to say, I'm okay with changes in the codebase as long as snapshots don't change, including this changes.

chu121su12:
I now catch your changes. I was more fixated on the changes on the mini snapshot.

PR #992, opened 2022-05-20
refactor: deprecate `className` in favor of `selectorName`

ydcjeff:
Since the preset supports both class and attributify mode, it is better to change to `selectorName` instead of `className`.

About removal, is 1.0 a good timeline?


antfu:
Could we do soft deprecation for now and remove it in the next breaking?

Issue #918, opened 2022-05-01
backdrop merging


ameinhardt:
In WindiCSS, I would combine (light/dark) backdrop filter like `backdrop-blur-sm backdrop-brightness-75 dark:backdrop-brightness-125`. It works fine in @apply and inline.
In UnoCSS, I noticed that @apply doesn't merge but overwrites all `backdrop-*` filter of an element. Inline works better, but the `dark:backdrop-` still overwrites the other backdrop-filters.
Would you elaborate on how [the array response](https://github.com/unocss/unocss/blob/e7d2fa0e2979d2be8a309408a22a03cdf75aede1/packages/preset-wind/src/rules/filters.ts#L50-L57) with CONTROL_SHORTCUT_NO_MERGE for dynamic rules/shortcuts is (supposed to be) used in these cases?

[...]

PR #994, opened 2022-05-20
fix: backdrop merging (#918)

ameinhardt:
it seems that the * filterBase were duplicated with each backdrop- * filter and thus overwriting previous definitions.
Should * filterBase not be used once with lower order instead?


antfu:
I think it was by design. To allow using filters without explicitly set `filter` class.

chu121su12:
See the `transform` code and try to align with that

PR #1000, opened 2022-05-23
feat(preset-typography): inherit mono font from theme

equt:
I find no reason to hardcode it into `DEFAULT`. An optimization may forward the `Theme` to the `cssExtend` to allow more flexibility and maybe less CSS var, should be a breaking change though.


equt:
for sure

antfu (approved):

PR #1001, opened 2022-05-23
feat(preset-wind): support bg position

zyyv:


packages/preset-mini/src/utils/handlers/handlers.ts
```diff
@@ -97,7 +97,7 @@ function bracketWithType(str: string, type?: string) {
   if (str && str.startsWith('[') && str.endsWith(']')) {
     let base: string | undefined
 
-    const match = str.match(/^\[(color|length):/i)
+    const match = str.match(new RegExp(`^\\[\(${type}\):`, 'i'))
```
antfu, line 100: This will create a new regex in every match which will hurt the performance, I think it's better to leave it as literal for now (probably also better to hoist it)

Issue #982, opened 2022-05-18
TailwindCSS importing colors within scss


Hextar:
In tailwind I'm able to get colors into scss like:
```
theme('colors.primary.500')
```

In some cases I could use @apply bg-primary-500, but in other I need to apply colors to props like `fill` or within a scss map.

Is there something similar?

antfu:
PR welcome!

PR #1005, opened 2022-05-25
feat(directives): support `theme()` function

Dunqing:
close: #982 
Related docs: https://windicss.org/features/directives.html#theme

- Improved implementation of directives to support easier addition of directives
- Supports the `theme()` function directive, which allows multiple parameters.


Dunqing:
There seems to be some problems with ci testing, the last few ci jobs have failed.

test/transformer-directives.test.ts
```diff
@@ -395,4 +395,31 @@ describe('transformer-directives', () => {
...
+      )
+      expect(result)
+        .toMatchInlineSnapshot(`
+        ".btn {
+          background-color: #3b82f6;
+          padding: 0.75rem 0.875rem;
+          margin: 0.75rem 0.875rem 1.25rem 1.125rem;
+
+          color: theme(\\"color.none.500\\");
+          font-size: theme(\\"size.lg\\");
```
antfu, line 419: Should we throw on none-exist utils?

Dunqing: > Should we throw on none-exist utils?

I think it would be better to keep the non-existent utils to better troubleshoot the problem.

antfu:
The multiple parameters feature feels a bit confusing to me, since ppl might expect to use space or comma for the conjunction base on their needs. I think either Tailwind or Windi support that usage currently, I'd prefer to not have it for the initial implementation (we could discuss in other issue)

Dunqing:
> The multiple parameters feature feels a bit confusing to me, since ppl might expect to use space or comma for the conjunction base on their needs. I think either Tailwind or Windi support that usage currently, I'd prefer to not have it for the initial implementation (we could discuss in other issue)

Ok,  I didn't notice that they don't support multiple parameters.

antfu (approved):

PR #1010, opened 2022-05-27
feat(preset-tagify): new preset

### LINKED ISSUE #1004: [Proposal]: Extractor Specific `escape` Function & Specify Extractor Name(s) in `RuleMeta`
## Motivation

I was experimenting[^1] with the icons preset to allow icons to be declared as custom HTML tags. For example, the following HTML code

```html
<i-carbon-sun />
```

would generate CSS below

```css
i-carbon-sun {
  --un-icon:url("...");
  mask:var(--un-icon) no-repeat;
  mask-size:100% 100%;
  -webkit-mask:var(--un-icon) no-repeat;
  -webkit-mask-size:100% 100%;
  background-color:currentColor;
  width:1.2em;
  height:1.2em;
}
```

However, there were a number of issues I encountered during my experiment:

1. Tag names are not extracted by default (fine, I'll create my own extractor)
2. Extracted raw tag names will be automatically escaped[^2] as class selectors (no workaround)[^3]
3. CSS for icon classes will still be generated (no workaround)

With these limitations stated above, I find it impossible to implement this feature without changing the internal generation mechanism. Hence, this proposal.

### Extractor Specific `escape` Function

Extractors can define a custom escape function to allow other custom escaping logic for the extracted content. e.g.

```js
const extractorClassNames = {
  name: 'class-names',
  extract(ctx) { /* logic */ },
  escape: selector => `.${selector}`
}
```

Extractors that don't define this method can still fallback to the default behavior for backwards compatibility. This would allow custom escape behavior other than the default.

### Specify Extractor Name(s) in `RuleMeta`

Rules should be able to specify which extractor(s) to generate CSS from. e.g.

```js
export default defineConfig({
  rules: [
    // lets say these extractors exist and behave as expected
    ['foo', { color: 'red' }, { extractors: 'tag-name' }],
    ['bar', { color: 'green' }, { extractors: 'attributify' }],
    ['baz', { color: 'blue' }, { extractors: ['class-name', 'attributify'] }],
  ]
})
```

With this config, the following HTML

```html
<foo bar class="baz" baz />
```

should generate this CSS

```css
foo { color: red; }
[bar=""] { color: green; }
.baz, [baz=""] { color: blue; }
```

This feature combined with extractor specific escape function would allow powerful customization capabilities, and attributify selectors can now be implemented as extractor + rule instead of hard-coded internal support. Rules without extractor specified should default to `['class-name', 'attributify']` for backwards compatibility.

## Potential Problems

- Refactor complexity
- Escaping attributified selectors (see below)

### Escaping Attributified Selectors

With the above proposal, we should be able to define the attributify extractor with a custom escape function.

```js
const extractorAttributify = {
  name: 'attributify',
  extract(ctx) { /* extractor logic */ },
  escape(matcher) {
    const splitted = matcher.split(/-/)
    return `[${splitted[0]}="${splitted.slice(1).join('-')}"]`
  }
}
```

Therefore this HTML

```html
<div flex="~ col" border-b="1 red solid" />
```

Should extract `Set { 'flex', 'flex-col', 'border-b-1', 'border-b-red', 'border-b-solid' }`. And rules with attributify extractors specified can match these. However, there is a catch. Our code is unaware of whether to escape `'border-b-1'` as `[border-b-1=""]` or `[border-b="1"]` or `[border="b-1"]`. To this problem I don't have an elegant solution. The only thing I can think of is to provide more information from the extractor and pass that info in the escape function.

```js
const extractorAttributify = {
  name: 'attributify',
  extract(ctx) {
    const result = new Set()
    //          matcher      payload
    result.add(['border-b-1', { head: 'border-b', tail: '1' }])
    return result
  },
  escape(matcher, { head, tail }) {
    return `[${head}="${tail}"]`
  }
}
```

[^1]: [experiment](https://github.com/zojize/unocss-icon-tag)
[^2]: [toEscapedSelector](https://github.com/unocss/unocss/blob/49a20e54668ebdc6bbc321cb9ecfcd3c25619e29/packages/core/src/generator/index.ts#L557-L562)
[^3]: Technically I could use full control rules, but this brings other problems

zojize:
# Preset Tagify

closes #1004 

## Features

- Tagified unocss rules
- Prefixed HTML tags
- Extra CSS properties

### Tagify + Attributify

```html
<!-- instead of writing this -->
<flex class="flex-col"> </flex>
<!-- maybe we could do -->
<flex col> </flex>
```

which would generate

```css
flex { display: flex; }
flex [row=""] { flex-flow: column; }
```

not sure how to do this with a regex only extractor

## Todos

- [ ] mention a11y In readme, since this is technically not good sementic HTML practice.


packages/preset-tagify/src/extractor.ts
```diff
@@ -0,0 +1,16 @@
+import type { Extractor } from '@unocss/core'
+
+export const MARKER = '__TAGIFY__'
+export const htmlTagRE = /<([\w\d-]+)/g
```
sibbng, line 4: 👀
```suggestion
export const htmlTagRE = /<([\w\d-:]+)/g
```

zojize: I wasn't sure if `:` is allowed as a legal HTML tag name according to [W3C standards](https://www.w3.org/TR/2011/WD-html5-20110525/syntax.html#:~:text=Tags%20contain%20a%20tag%20name,005A%20LATIN%20CAPITAL%20LETTER%20Z.), but it seems to work on chrome so why not :p

packages/preset-tagify/README.md
```diff
@@ -0,0 +1,77 @@
...
+})
+presetTagify({
+  // extraProperties can also be a plain object
+  extraProperties: { display: 'block' }
+})
+```
+
+## License
+
+MIT License &copy; 2022-PRESENT [Jeff Zou](https://github.com/zojize)
```
antfu, line 76: @zojize I have added your credit here. I will merge this for now, feel free to send another PR if you'd like to adjust this part. Thanks!

antfu:
Merge and ship for now. I like the idea of using the tag name as the prefix, let's iterate that for later.

Issue #1021, opened 2022-05-29
theme() breaks @apply


ameinhardt:
Very cool that you are including theme() support!
It seems that @apply and theme() currently can't be used in the same rule. Even if it's only mentioned as a comment, see [Stackblitz](https://stackblitz.com/edit/vitejs-vite-qabktp?file=src%2Fstyle.css&terminal=dev)

Also, with that implementation after parsing to AST, it won't be applied to `childNode.type === 'Declaration' && childNode.value.type === "Raw"`, isn't it? This is how users might expect it from [tailwind](https://play.tailwindcss.com/Hnd4BmJbK9?file=css) or [windi](https://windicss.org/play.html?html=DwEwlgbgBAxgNgQwM5ILwCIAuBTJn0B8AFtnHAPbAD04EBQA&css=CYSwbgBA3gUBEAECGAHFAbAnhAZugpgB4QAWAtDgK7roQDuF1tAVpQM4AuIOmZAxvgB2HfACcIIEQFs2-ISNEBuOBDJkpvPgHt0W0QC4IHEvin4AFAHJtu0WwB0o-MHsBWAAzvLASmXwbeoZgSKLmahr8Onq%2BMAC%2BQA).
Would it be an option to use simple string replacement [before AST parsing](https://github.com/unocss/unocss/blob/3e2182460457da98bbd92943ccd999d86f10b08d/packages/transformer-directives/src/index.ts#L53) instead? As in something like the following:
```javascript
  const regTheme = /theme\((["']?)([^\1)]+)\1\)/g;
  let result = regTheme.exec(code.original);
  while (result != null) {
    const [match,, args] = result
    const keys = args.split(".")
    let value = uno.config.theme
    keys.every((key) => {
      if (!Reflect.has(value, key)) {
        value = null
        return false
      }
      value = value[key]
      return true
    })
    code.overwrite(result.index, result.index + match.length, value);
    result = regTheme.exec(code.original);
  }
```

antfu:
PR welcome!

PR #1025, opened 2022-05-29
fix(directives): `theme()` breaks `@apply`

Dunqing:
fix: #1021


Dunqing:
css-tree cannot parse Function in css variables, so theme() does not work.

Do we need to support it?

@antfu cc

hannoeru:
@Dunqing you can get raw data then throw it back to css-tree parser to parse that part as Function.

antfu:
We could just do a simple regex replace instead of relaying on the parse result.

PR #1029, opened 2022-05-30
feat(theme): safe area

okxiaoliang4:
safe area spacing


chu121su12:
Note that this is a tricky property to tackle as in https://github.com/tailwindlabs/tailwindcss/pull/6366

antfu:
I am not sure about it either. To me, they are too specific. I would imagine you only need it in the container, but I guess it might not make sense to use it in components?

antfu:
Closing for now. Currently, I'd suggest it be defined on userland.

PR #1036, opened 2022-05-31
feat(core)!: remove no merge control

chu121su12:
Previously, the `CONTROL_SHORTCUT_NO_MERGE` was added to address shortcut merging with rules that have "base" properties (#429) like transforms, filters, etc. Now that those rules are refactored to use separate layer (#1028, #1033) this control is not necessary inside the core anymore.

The breaking change is for developer/user that use the control in their codebase. This can closed too if the control is deemed necessary.


antfu:
I see there is no harm to keep supporting it, an escape hatch is better than not being able to do. Thanks.

PR #1045, opened 2022-05-31
fix: use absoultePath as fileCache key

TrickyPi:
## Motivation
exist different key of the same file
like this:
image


chu121su12:
For cli test & snapshot, please run `pnpm bulid` first

TrickyPi:
Thanks for your reply, i had run this command before committing.

antfu (approved):

test/__snapshots__/cli.test.ts.snap
```diff
@@ -19,5 +19,5 @@ exports[`cli > uno.css exclude initialized class after changing file 1`] = `
...
 /* layer: default */
-.bg-red{--un-bg-opacity:1;background-color:rgba(248,113,113,var(--un-bg-opacity));}"
+.bg-blue{--un-bg-opacity:1;background-color:rgba(96,165,250,var(--un-bg-opacity));}"
```
TrickyPi, line 22: @antfu the expected className is `.bg-red`,see here https://github.com/unocss/unocss/pull/1045/files#diff-1deee0a575599b2df117c280da319f7938aaf6fdb0c04bcdbde769dbf464be69R64

TrickyPi: Hmmm! that make sense

TrickyPi: Should I create a new PR to change this test?

PR #1047, opened 2022-06-01
feat(preset-tagify): add excludeTags option

chu121su12:
This PR adds option to ignore specified tags from being processed. Primarily used for h1-h6.

/cc @zojize If the default values are not ok on the main plugin maybe we'll try them only on runtime.

### DIFF
--- a/packages/preset-tagify/src/types.ts
+++ b/packages/preset-tagify/src/types.ts
@@ -5,6 +5,12 @@ export interface TagifyOptions {
   prefix?: string
 
   /**
+   * Tags excluded from processing.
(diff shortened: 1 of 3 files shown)


packages/preset-tagify/src/types.ts
```diff
@@ -4,6 +4,12 @@ export interface TagifyOptions {
    */
   prefix?: string
 
+  /**
+   * Tags excluded from processing.
+   * @default ['b', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'table']
+   */
+  excludedTags?: string[]
```
antfu, line 11: Should also be possible to accept a regex

antfu (approved):

PR #1049, opened 2022-06-01
feat: show generated info in watch mode

TrickyPi:
Seems cleaner than the current output

after changing

before changing


TrickyPi:
Hmm, I wonder whether the user care about which utilities are generated for uno.css in watch mode

johannschopplich:
He doesn't in my opinion. That's why excluded the information in watch mode in the first place. 🙂

TrickyPi:
well, then i close this pr

PR #1052, opened 2022-06-02
chore: replace `esno` with `tsx`

johannschopplich:
- Since `esno` is an alias by now
  - Thank you for the great work on `esno`!
- Also fixed `scripts/size.ts` (had top-level await)


antfu:
What's the benefit of it? While `esno` is now an alias of `tsx`, it's not deprecated. They are served to different preferences. I personally prefer to use `esno` as it reads more related to `node`.

johannschopplich:
As a developer, one less alias to know eases my life. That's why I prefer `tsx`. But I get your point. 🙂

PR #1056, opened 2022-06-03
feat: add defaultUnocssConfig to webpack plugin and unocss vscode extension

### LINKED ISSUE #1032: [Vscode extension] the preset config is prerequisite
I want to use Unocss Vscode extension in my unocss+vite+react project, it doesn't work until i config presets, and i didn't find this prerequisite for using Unocss Vscode extension. if it isn't a bug, maybe we should list this prerequisite in the README.md under packages/vscode?

# addtional info
it doesn't work
```js
//vite.config.ts
import { defineConfig } from 'vite'
import Unocss from 'unocss/vite'
import react from '@vitejs/plugin-react'
import tsconfigPaths from 'vite-tsconfig-paths'

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    react(),
    Unocss({}),
    tsconfigPaths(),
  ],
})

```
it works
```js
//vite.config.ts
import { defineConfig } from 'vite'
import Unocss from 'unocss/vite'
import { presetUno } from 'unocss'
import react from '@vitejs/plugin-react'
import tsconfigPaths from 'vite-tsconfig-paths'

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    react(),
    Unocss({
      presets: [presetUno()],
    }),
    tsconfigPaths(),
  ],
})

```
Please let me know if you need more information.

TrickyPi:
close #1032 

also make webpack plugin use [@unocss/preset-uno](https://github.com/unocss/unocss/tree/main/packages/preset-uno) as default preset.

### DIFF
--- a/packages/shared-integration/package.json
+++ b/packages/shared-integration/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@unocss/core": "workspace:*",
+    "@unocss/preset-uno": "workspace:*",
     "@unocss/scope": "workspace:*"
   }
 }
--- a/packages/shared-integration/src/defaults.ts
+++ b/packages/shared-integration/src/defaults.ts
@@ -1,4 +1,10 @@
+import type { UserConfigDefaults } from '@unocss/core'
 import { cssIdRE } from '@unocss/core'
+import presetUno from '@unocss/preset-uno'
 
 export const defaultExclude = [cssIdRE]
 export const defaultInclude = [/\.vue$/, /\.vue\?vue/, /\.svelte$/, /\.[jt]sx$/, /\.mdx?$/, /\.astro$/, /\.elm$/]
+
+export const defaultUnocssConfig: UserConfigDefaults = {
+  presets: [presetUno()],
+}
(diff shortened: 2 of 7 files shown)


antfu (changes requested):
Well, it's intentional to have `@unocss/vite` and `@unocss/webpack` preset irrelevant so higher integrations that provide their own presets could reuse them if they want. I am ok to have the default theme on the extension but not the webpack plugin. Thanks

TrickyPi:
The default presets will be overridden by custom configuration, like this describition https://github.com/unocss/unocss/blob/main/README.md?plain=1#L166, so i think there will be no issues about integration(maybe i'm wrong).
perhaps we can export a function like [UnocssVitePlugin](https://github.com/unocss/unocss/blob/main/packages/unocss/src/vite.ts#L8) for webpack user,

antfu:
I don't want to force download the presets for those who don't need them. Maybe we could have `unocss/webpack` which binds the default preset and ask `@unocss/webpack` as peerDeps

TrickyPi:
if we ask @unocss/webpack as peerDeps, maybe we need list this info in instruction.  Could we add instruction about how to use unocss on  webpack  to  [README.md](https://github.com/unocss/unocss/blob/main/README.md)?

antfu:
Sure, I would handle the docs part

TrickyPi:
Thanks for your commits.

PR #1082, opened 2022-06-09
Add `transformer-typography`

equt:
### TLDR

This transformer allows you to style your dynamic content like

```ts
{
  "prose": {
    "p": ["font-sans", "text-gray-700"],
    "p code": ["font-mono", "text-sm", "rounded", "bg-light-400", "px-2", "py-1"],
    "h1, h2": ["text-2xl", "lg:text-3xl"],
  }
}
```

It evolves from the [`preset-raw`](https://github.com/equt/preset-raw) I posted a few weeks ago in the discord channel.

### TODO

- [ ] Better tokens discovery, make it framework agnostic
- [ ] Pseudo selectors
- [ ] ? Runtime class injection


antfu:
I don't quite understand what it does, and what's the problem it's trying to solve. Can you describe it a bit more?

equt:
It basically lets you apply utils according to the selectors.

---

We could currently write something like

```html
<div>
  <p class="foo bar">
    <code class="buz">
    </code>
  </p>
<div>
```

and UnoCSS will generate utils classes `foo bar buz` for us **only if we have the above template in the compile-time**.

The `@unocss/preset-typography` allows us to style content generated at the runtime (or something we don't have control of), by introducing `prose-[name]`, but with one drawback, that **we're writing raw CSS again**, quoted from the current preset.

https://github.com/unocss/unocss/blob/e96aa039b43a37bd818e885af84a0150738103d2/packages/preset-typography/src/preflights/default.ts#L1-L24

This PR tries to make the following code work

```ts
export const DEFAULT = { 
   'h1, h2, h3, h4, h5, h6': ['text-gray-700', 'font-semibold', 'leading-relaxed'], 
   'a': ['text-gray-800', 'underline','font-semibold'],
   // ...
}
```

equt:
I'm going to close this PR since the initial goal (styling generated content using existing utility classes) has been achieved in the upstream rendering process. We believe that should be the correct way to handle such cases.

PR #1093, opened 2022-06-11
feat(core)!: support string in rule that returns array

### LINKED ISSUE #1092: Animate @keyframes genrate twice
I found `@keyframes` will genrate twice when use `presetAttributify` together, it's unnecessary.
How to avoid and handle it in the best way?

[playground](https://unocss.antfu.me/play/?html=DwEwlgbgBAhgdmAtjALgUwLQCMD2BXOAY0xgBsUA%2BAKCigJ0IGdGrgB6cCaoA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcEWWwCGuOlmggMUAGrr2mmAHEoELmCQBfOFksg4Aci4qJdOi6RIMAD0hYdGx1LgAbeEwcfCISYFIACmRUKAj6AC44AG0qVGyXCS46GAgQAFo08IwXckQ4KXDoLJdaNBc4WwBdKi7ZOgALaBgimDoslNRXIpKy8sHh0ZcWmACYcvDSOFX-dehNUgw4IYA3fQyd9dX1cJ9UW1kRBXGcvLlRJRVEgEpZYXlGCw2JweHx%2BD8-u8FJJpMk3qg6BIbhgsgBGAB0ACZIagJGhcC0BqwwOMAPSk%2BggdGDUl3Ka2X69NRQTTaXT6F65KbbFlaHRQPSGYzAUwWKxgCFMuzfJBAA&options=N4IgzgLgTglgxhEAuaBXApgXyA)

chu121su12:
Closes #1092.

This PR expanded support for rule to [return string](https://github.com/unocss/unocss#full-controlled-rules) inside an array.
The (somewhat) breaking change is that string-returned value will now be sorted among the same token.

### DIFF
--- a/test/__snapshots__/preset-attributify.test.ts.snap
+++ b/test/__snapshots__/preset-attributify.test.ts.snap
@@ -41,24 +41,6 @@ exports[`attributify > autocomplete extractor 5`] = `
 exports[`attributify > compatible with full controlled rules 1`] = `
 "/* layer: default */
 
-  [custom~=\\"\\\\31 \\"] {
-    font-size: 1px;
-  }
-  /* you can have multiple rules */
-  [custom~=\\"\\\\31 \\"]::after {
-    content: 'after';
-  }
-  .foo > [custom~=\\"\\\\31 \\"] {
-    color: red;
-  }
-  /* or media queries */
-  @media (min-width: 680px) {
-    [custom~=\\"\\\\31 \\"] {
-      font-size: 16px;
-    }
-  }
-  
-
   .custom-2 {
     font-size: 2px;
   }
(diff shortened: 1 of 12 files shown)


test/__snapshots__/preset-attributify.test.ts.snap
```diff
@@ -41,37 +41,37 @@ exports[`attributify > autocomplete extractor 5`] = `
...
 "/* layer: default */
 
-  [custom~=\\"\\\\31 \\"] {
-    font-size: 1px;
+  .custom-2 {
+    font-size: 2px;
   }
   /* you can have multiple rules */
-  [custom~=\\"\\\\31 \\"]::after {
+  .custom-2::after {
```
antfu, line 48: This seems not correct 🤔

chu121su12: This is the breaking change noted above. The result was swapped with its second result to its bottom.

That particular rule is based on string return, so it will not have `selector` in `StringifiedUtil` hence sorted by its body. That portion of rule start with the captured token used as selector:
https://github.com/unocss/unocss/blob/0c861ebb8cb1fb39770955a7ac31e7f026438054/test/preset-attributify.test.ts#L87-L91
The string to extract was `<div custom="1" class="custom-2">` yielding `[custom~="1"]` and `.custom-2`. Sorting them leads to `.` > `[` hence the swapped result.

antfu (approved):

PR #1113, opened 2022-06-18
Update vite.config.ts

Scouterdev:
fix path for in @unocss/vite


userquin:
the `UnoCSS Vite Plugin` is included on `unocss` package (https://github.com/unocss/unocss/tree/main/packages), you don't need to change the import

antfu:
It's intentional. Both works, but `unocss/vite` provides built-in presets as a fallback. Thanks

Issue #985, opened 2022-05-18
[VSCode extension] Provide autocompletion for style file, eg `.css` file


0x-jerry:
WindiCSS provides `@apply` autocompletion for style file, eg. `.css` file. It could be very nice if this feature includes in UnoCSS extension.

![image](https://user-images.githubusercontent.com/14226737/169098229-eb5f2975-42b5-46a8-87b5-7b84db7cedbe.png)

btw, I can contribute if someone can point me in the right direction.

NicolaSpadari:
+1

For now I just install the windicss intellisense extension and disable almost everything.
Would be nice to have some of its features in the unocss extension too

PR #1118, opened 2022-06-19
fix(vscode): autocomplete for css

TrickyPi:
### motivation
support prompt of autocomplete and annotation for css file as before.

cc @QiroNT

### DIFF
--- a/packages/vscode/src/annotation.ts
+++ b/packages/vscode/src/annotation.ts
@@ -60,9 +60,9 @@ export async function registerAnnotations(
         return reset()
 
       let ctx = await contextLoader.resolveContext(code, id)
-      if (!ctx && (code.includes(INCLUDE_COMMENT_IDE) || isCssId(id)))
+      if (!ctx)
         ctx = await contextLoader.resolveClosestContext(code, id)
-      else if (!ctx?.filter(code, id))
+      else if (!ctx.filter(code, id) && !code.includes(INCLUDE_COMMENT_IDE) && !isCssId(id))
         return null
 
       const result = await ctx.uno.generate(code, { id, preflights: false, minify: true })
--- a/packages/vscode/src/autocomplete.ts
+++ b/packages/vscode/src/autocomplete.ts
@@ -71,9 +71,9 @@ export async function registerAutoComplete(
         return null
 
       let ctx = await contextLoader.resolveContext(code, id)
-      if (!ctx && isCssId(id))
+      if (!ctx)
         ctx = await contextLoader.resolveClosestContext(code, id)
-      else if (!ctx?.filter(code, id))
+      else if (!ctx.filter(code, id) && !isCssId(id))
         return null
 
       try {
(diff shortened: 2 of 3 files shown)


QiroNT:
Hi, thanks for the contribution.

I think the changed condition is still the same, and css-like files actually have context cache.

https://github.com/unocss/unocss/blob/1097c5bd682294a48f32a3ddf2ffaa965a269291/packages/vscode/src/contextLoader.ts#L235-L238

What's the problem with the current implementation? After fixing #1119 I don't see any issue on my side.

TrickyPi:
The changed Conditional isn't totally same as before. 
https://github.com/unocss/unocss/pull/1118/files#diff-3241d63de07ad5b12662fc4d9f3559f3e4c2836772e8b921612675c05b9f6f87L76
the result of executing `context.filter(code, id)` is false if the `id` param is a path of css file，so the prompt of autoComplete for css file does not work.

QiroNT:
I think... AutoComplete will still work since `resolveClosestContext` will 100% resolve a context for it.

The current logic before change basically is:
- If `resolveContext` resolved a context and `filter` passed, then use the context.
- If it's a css-like file, use the context from `resolveClosestContext`.
- If none of the above matches, the prompt will not be activated.

There for, css-like files will always get a context either from `resolveContext` or `resolveClosestContext`, it should never fail.

TrickyPi:
it will cause another issue.
Existed the following dir construct
```
├── index.html
├── package.json
├── pnpm-lock.yaml
├── src
│   ├── App.css
│   ├── App.tsx
│   ├── favicon.svg
│   ├── index.css
│   ├── logo.svg
│   ├── main.tsx
│   └── vite-env.d.ts
├── tsconfig.json
├── tsconfig.node.json
└── vite.config.ts
```
Actually, the context of css-like file is `null`
Caused by this code 

https://github.com/unocss/unocss/blob/1097c5bd682294a48f32a3ddf2ffaa965a269291/packages/vscode/src/contextLoader.ts#L131toL136

TrickyPi:
> Oh I know, css-like files which have a context but don't pass the filter will not be resolved.
> Is that a real issue? If the file is excluded, I think the autocomplete prompt shouldn't be activated.

yeah, it's a real issue(see https://github.com/unocss/unocss/issues/985), by the way i don't know why the `defaultExclude` contains css-like files.
https://github.com/unocss/unocss/blob/main/packages/shared-integration/src/defaults.ts#L3

QiroNT (approved):
I see, it's ok.

If `resolveClosestContext` could be called only when needed then it would be better. It's not that expensive tho.

antfu:
Can you help resolve the conflicts? Thanks

TrickyPi:
Ok, just a minute.

TrickyPi:
Done.

ckvv:
@TrickyPi 
autocompletion seems to only take effect in css files, can it take effect elsewhere? For example, the vue file unocss.config.ts, etc.

Issue #1104, opened 2022-06-15
`checked:next:bg-red` styles are generated wrongly


jd-solanki:
Previously I was able to apply style to next children via `checked:next:bg-red`. Unfortunately from now on, I can't.

~~Docs link: https://uno.antfu.me/?s=checked:next:bg-primary~~
Docs link: https://uno.antfu.me/?s=checked:next:bg-red

jd-solanki:
In my existing project:
![image](https://user-images.githubusercontent.com/47495003/173807302-45636a50-6bc6-4c48-81b7-ea082031c4b1.png)

In 0.39.0:
https://uno.antfu.me/?s=checked:next:bg-red

chris-zhu:
I remember a PR that changed the order of the variants， try to use `next:checked:bg-red`

jd-solanki:
Upon inspect I found that usage of `checked:next:bg-red` is now `next:checked:bg-red`. But, I think previous `checked:next:bg-red` was making much more sense:

- `checked:` => Current state
- `next:` => Aim for all children
- `bg-red` => Add bg

jd-solanki:
@chris-zhu Please let me know your thoughts on the above comment.

If you think the order is fine and not confusing you can close this issue.

Regards :)

PR #1126, opened 2022-06-21
feat(core): use recursive callback to resolve variant

chu121su12:
- update `applyVariants`
- ~deprecate most of `VariantHandler` properties, moving them to `handler()`~
- add `VariantHandler.handle()`

The main goal for this is to enable usecase of #1104 without plenty of selector replace hack, while also still supports #987

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -220,7 +220,7 @@ export interface VariantHandler {
   /**
    * Callback to process the handler.
    */
-  handler?: (input: VariantHandlerContext, next: (input: VariantHandlerContext) => VariantHandlerContext) => VariantHandlerContext
+  handle?: (input: VariantHandlerContext, next: (input: VariantHandlerContext) => VariantHandlerContext) => VariantHandlerContext
   /**
    * The result rewritten selector for the next round of matching
    */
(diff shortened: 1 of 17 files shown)


packages/core/src/types.ts
```diff
@@ -189,32 +189,73 @@ export interface Preflight<Theme extends {} = {}> {
...
-   */
-  order?: number
-  /**
+   * @deprecated use `handler` instead. It will be removed in 1.0.
+   *
    * Order in which the variant is sorted within single rule.
    */
   sort?: number
   /**
+   * @deprecated use `handler` instead. It will be removed in 1.0.
```
antfu, line 257: I don't feel it's worth it to deprecate them. If we really need to be strict, we could have `type VariantHandler = VariantHandlerObject | VariantHandlerFunction` to say you can't use the object style when `handle` is provided.

packages/core/src/types.ts
```diff
@@ -189,32 +189,73 @@ export interface Preflight<Theme extends {} = {}> {
...
+   * Order in which the variant is sorted within single rule.
+   */
+  sort?: number
+}
+
 export interface VariantHandler {
+  /**
+   * Callback to process the handler.
+   */
+  handler?: (input: VariantHandlerContext, next: (input: VariantHandlerContext) => VariantHandlerContext) => VariantHandlerContext
```
antfu, line 223:
```suggestion
  handle?: (input: VariantHandlerContext, next: (input: VariantHandlerContext) => VariantHandlerContext) => VariantHandlerContext
```

test/variant-handler.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+describe('variants', () => {
+  test('variant context is propagated', async () => {
+    const uno = createGenerator({
+      rules: [
+        ['foo', { name: 'bar' }],
+      ],
+      variants: [
+        {
+          match: (input: string) => {
```
antfu, line 12:
```suggestion
          match(input) {
```

I guess the type can be inferred

antfu (approved):

PR #1132, opened 2022-06-22
fix(preset-wind): gradient stop via higher than stop to in rules

tammai-bigin:
When using from-* via-* to-*, to stop be override by via because the order in rules.


antfu:
Can you update the snapshot? Thanks

tammai-bigin:
My mistake. This is a first time I make a PR. Snapshot updated!

PR #1133, opened 2022-06-22
feat(webpack): support `pre` transformer

### LINKED ISSUE #1096: Transformer support for Webpack
image

ChpShy:
- fix #1096


ChpShy:
Unfortunately, only support transformer that set `enforce: pre` currently. I don't have a good idea for this at the moment.

ChpShy:
@antfu Is there have some problem? I want to commit another PR about `pre-scan` and want to use this PR's  content .😂

PR #1146, opened 2022-06-23
fix(playground): remove unocss options in storage

chris-zhu:


antfu:
I don't want to write defaultOptions to storage as it would prevent future changes to take effect.

antfu (approved):

PR #1147, opened 2022-06-23
fix(playground): remove `strict` option

chris-zhu:
Currently, `defaultConfig` will override `userConfig` in playground.  It was invalid when `userConfig` no `strict` option.
image

Now, this option will control `userConfig` to toggle `strict`

image


antfu:
This is a bit hacky to me 😅. While the strict option was introduced when we apply default config to users', with the current approach of exposing the entire config, maybe we could just remove the option and let users define the option in their configure manually?

chris-zhu:
> This is a bit hacky to me 😅. While the strict option was introduced when we apply default config to users', with the current approach of exposing the entire config, maybe we could just remove the option and let users define the option in their configure manually?

OK, I have already removed it 😄.

PR #1158, opened 2022-06-25
fix: add config type

xiaojieajie:


antfu:
It should be auto inferred, you may run dev first.

xiaojieajie:
After I run dev, I still have any

image

PR #1162, opened 2022-06-25
fix(core)!: revert variant apply order in #1127

antfu:
- #987
- #1127


antfu:
from my understanding, variant should be apply like:

`foo:bar:p1` -> `foo:(bar:p1)`

sibbng (approved):
LGTM

chu121su12:
Needs more revert on the variants definitions

PR #1169, opened 2022-06-26
fix(vscode): respect `unocss.root`

### LINKED ISSUE #1183: [vscode-intelisence] incorrectly detects typescript class as CSS
In a project I have code similar to the following, `static` is being detected by `UnoCSS`.

```ts
export class UserService {
    static async findById(id: string) {
        console.log(id);
    }
}
```

image
image

holazz:
After v0.40.0, configuring `unocss.root` does not work, decorations and tooltips are displayed in all files.

close #1183


TrickyPi:
I guess the real issue in here
https://github.com/unocss/unocss/blob/main/packages/vscode/src/index.ts#L22toL28

the `registerAutoComplete` and `registerAnnotations` function always be called , even though the project does not contain any unocss config files, and the `ContextLoader` would create a default context.

holazz:
Hi, @TrickyPi  thanks for your comment. I updated a commit. The extension shouldn't be activated when the config file is not found under `unocss.root`, right?

This is my first PR on `unocss` and my changes may be ill-considered. Hope someone more experienced can fix it if I'm in the wrong direction.

packages/vscode/src/annotation.ts
```diff
@@ -53,9 +53,11 @@ export async function registerAnnotations(
       if (!doc)
         return reset()
 
-      const code = doc.getText()
       const id = doc.uri.fsPath
+      if (!id.startsWith(cwd))
```
TrickyPi, line 57: I guess it's unnecessary.

packages/vscode/src/autocomplete.ts
```diff
@@ -64,9 +65,11 @@ export async function registerAutoComplete(
 
   const provider: CompletionItemProvider<UnoCompletionItem> = {
     async provideCompletionItems(doc, position) {
-      const code = doc.getText()
       const id = doc.uri.fsPath
+      if (!id.startsWith(cwd))
```
TrickyPi, line 69: As above.

packages/vscode/src/index.ts
```diff
@@ -19,12 +19,17 @@ export async function activate(ext: ExtensionContext) {
   log.appendLine(`UnoCSS for VS Code  v${version} ${process.cwd()}`)
 
   const contextLoader = new ContextLoader(cwd)
+
+  const hasConfig = await contextLoader.loadContextInDirectory(cwd)
+  if (!hasConfig)
```
TrickyPi, line 24: The `load` function in `loadContextInDirectory` may be called twice, because `new ContextLoader` calls loadContextInDirectory by default, so it's better to call `contextLoader.loadContextInDirectory` after `contextLoader.ready`, and this call will hit the `contextsMap` cache (the cache value is `null`)

TrickyPi (commented):
These are my ideas.

TrickyPi:
> The extension shouldn't be activated when the config file is not found under `unocss.root`, right?

Yes

antfu (approved):

antfu (approved):

PR #1170, opened 2022-06-26
feat(preset-mini,preset-wind): add global keywords

chu121su12:
Basically, adding `handler.global` almost everywhere. For static rules, `makeGlobalStaticRules` were introduced to help with this.

PR #1171, opened 2022-06-26
feat(preset-mini,preset-wind)!: add more global keywords

chu121su12:
This PR adds on top of #1170, but instead introduces fix that can be considered breaking change.
- use border style list from in `divide-x` style
- update `font-synthesis` to use static rules with bracket rules for compund values (See https://developer.mozilla.org/en-US/docs/Web/CSS/font-synthesis#syntax)
- fix: support global value for `animate-x`. previously `animate-<global>` would produce `animation-play-state: <global>`
- fix: properly support `list-<style>-<outside|inside>`
- fix: `line-clamp-none` was set to `line-clamp: unset`
- add `h.global` to the related utilites


antfu:
What would be the breaking changes?

chu121su12:
I guess due to slight behavior change especially in these:
- changes to font-synthesis rule order (converted to static)
- the animate-<global>
- line-clamp-none

Otherwise it can be more of fix update

Issue #1168, opened 2022-06-25
Feature request: support for opacity on color related utilities that use a `custom property` suffixed with `-rgb` as their value


UltraCakeBakery:
The combination of a custom property value like `bg-$custom-color` and opacity utility like `bg-opacity-50` does not work.

I want to propose that when a utility using a custom property is suffixed with `-rgb`, unocss generates values for the utility that are compatible with the related opacity utility.

# example

###### Input:
```html
<div style="--color-rgb: 255, 0, 0">
<div class="bg-$color-rgb bg-opacity-50">I'm half red!</div>
</div>
```

###### output:
```css
.bg-\$color-rgb { background: rgba( var(--color-rgb), var(--un-bg-opacity) ); }
```

chu121su12:
For now, I'd recommend to define the opacity manually inside another variable:
```html
<div style="--color-rgb: 255, 0, 0">
<div style="--_color-rgb: rgb(var(--color-rgb), var(--un-bg-opacity, 1))"
     class="bg-$_color-rgb bg-opacity-50">
I'm half red!</div>
</div>
```

CSS color syntax is quite handy as it is especially pairing it with custom properties

UltraCakeBakery:
> For now, I'd recommend to define the opacity manually inside another variable:
> 
> ```
> <div style="--color-rgb: 255, 0, 0">
> <div style="--_color-rgb: rgb(var(--color-rgb), var(--un-bg-opacity, 1))"
>      class="bg-$_color-rgb bg-opacity-50">
> I'm half red!</div>
> </div>
> ```
> 
> CSS color syntax is quite handy as it is especially pairing it with custom properties

The solution you provided requires me to redefine the custom property on every element that uses the custom property, which will become very messy due to how many times I use that variable.

I basically need to get it working like this below:

```
<main style="--color-rgb: 255, 0, 0;--color: rgba(var(--color-rgb), var(--un-bg-opacity, 1))">
   <div class="bg-$color bg-opacity-50">
     I'm half red!
   </main>
</div>
```

chu121su12:
[...]

Issue #623, opened 2022-02-18
bg-opacity-<value> doesn't work with extended colors


jd-0001:
Hi,

I added some of my own colors as per [this](https://github.com/unocss/unocss#extend-theme) guide.

However, I found that if I use `bg-primary bg-opacity-50` opacity isn't get applied.

I guess, we need to convert those colors in our own form and add css var for bg opacity.

Regards.

jd-0001:
Yes, I do. Uno preset.

Btw, I have colors as:

```js
colors: {
  primary: "hsl(var(--primary))",
  success: "hsl(var(--success))",
  info: "hsl(var(--info))",
  warning: "hsl(var(--warning))",
  danger: "hsl(var(--danger))",
},
```

```css
:root {
  --primary: 261.4, 97.7%, 66.3%;
  --success: 94.5, 100%, 39.6%;
  --info: 200.1, 100%, 54.3%;
  --warning: 42.4, 100%, 50%;
  --danger: 358.3, 100%, 64.9%;
}
```

PR #1172, opened 2022-06-26
feat: support for `var(*` values in `theme.colors`

UltraCakeBakery:
Solves https://github.com/unocss/unocss/issues/1168

It allows for direct usage of css variables inside theme colors.


chu121su12:
There's [support](https://github.com/unocss/unocss#extend-theme) if you use proper css color. See also #623.

UltraCakeBakery:
@chu121su12 I see. I'm going to close my PR here. Thanks for pointing it out. I completely overlooked that in the code haha.

PR #1188, opened 2022-06-29
feat: ignore .d.ts by default

TrickyPi:
related #1183


antfu:
It's not even in `include`, why do we need to exclude it? 👀

TrickyPi:
Oh it's my fault, i mistaken `/\.[jt]sx$/` for `/\.[jt]s$/`.

PR #1189, opened 2022-06-29
docs(playground): add format buttons and improve design

chris-zhu:


packages/inspector/client/components.d.ts
```diff
@@ -1,6 +1,6 @@
 // generated by unplugin-vue-components
 // We suggest you to commit this file into source control
-// Read more: https://github.com/vuejs/core/pull/3399
+// Read more: https://github.com/vuejs/vue-next/pull/3399
```
antfu, line 3: You are using an old local dep

chris-zhu: This file was auto generated, i don't know the reason about this change. I'm changed it manually now. I will check upstream code about `unplugin-vue-components`.

antfu:
It's a bit weird to have checkboxes of "prettify" on editable content. I would suggest using buttons for "HTML" and "Config" tabs. And better to make then right-aligned (need to move the github and links to top bar)

chris-zhu:
> It's a bit weird to have checkboxes of "prettify" on editable content. I would suggest using buttons for "HTML" and "Config" tabs. And better to make them right-aligned (need to move the github and links to top bar)

I had adjust this buttons position layout, they are right-aligned both.
And i fount `transform` button was invalid, i will fix it in a new PR

antfu:
I think for "HTML" and "Config", they should be "Format" buttons instead of checkboxes (we could use icon buttons), to serve one-time formatting instead of keeping two sources.

antfu:
You can pick any icon as a placeholder for now. Later I'd like to revise all icons to make them consistent.

Issue #1178, opened 2022-06-27
Shortcut referencing another shortcut with a variant


ghalle:
Referencing a shortcut within another shortcut works, but with a variant it doesn't work.

I'm not sure if this is supposed to work or not but it seems a bit inconsistent that it works without a variant and doesn't work with one.

In the example below the `accent` class doesn't work on hover.

[Example](https://uno.antfu.me/play/?html=DwEwlgbgBAFgtAMwK4BsVQC4FMAeG4DGWAdtgE5QIq5QDOW1B%2BxA9sVlAIZoBcGZnYrTAYwbOACYADFIB8AKChRQkKAFtOCpUpXQCKTrVoBeAEQAjDMUy58ANhzoADnAAsprdqWXii7cAB6cAhPHWCofUMTCysJGzw4B2c3Dz8vHwk05SDIT0DghXzcoA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6NKD2UPx6eTBWMInQ9rikGIkAHAAME3AcpIkA7lXAvemR%2BvYSEhi4MN36vZb9EvxD41MzcxxOGwDWqy36HDC4e3ptIB38cDUAbhhQwR8Gy2OzghkwUCceAwAEp7i09E9cAAmV7vT7fCB-AE%2BA5HE64M7TWaJK63MG4CFQ-BwqhqWQiBSNcJrRmMZQQHww2TCeSMFhsTg8Pj8Lk8uSiGCSaR%2BNaoOgSeypUIARgAdMjxagJGhcHkqqwwI0APTG%2BggNXVY1mB5qbnlJB2pBAA&options=N4IgzgLgTglgxhEAuaBXApgGhNAhgOzADMB7KAW2SNwBswsQp0wAHEwmAN3StvoF8gA)

PR #1192, opened 2022-06-30
feat(core): expand shortcuts with variants

edwardnyc:
fix: [1178](https://github.com/unocss/unocss/issues/1178)


test/assets/preset-mini-targets.ts
```diff
@@ -962,6 +962,9 @@ export const presetMiniTargets: string[] = [
 
   // variants - multiple parents
   'sm:lt-lg:p-10',
+
+  // shortcuts
+  'btn',
```
antfu, line 967: Can we move to another test suite so it could be clear it's testing the shortcuts instead of the preset?

I'd also want to test `hover:btn` and `focus:btn`

antfu:
Side note, you might want to change your local git email, I see you are committing as `ByteDance` 👀

PR #1205, opened 2022-07-01
perf(preset-mini): apply preflight vars to `:root`

antfu:
I don't know why we aren't doing it from the beginning 🤣

/cc @chu121su12 in case you are aware of any caveat of this approach.


sibbng:
This change turns per element variables into some kind of global variables. As a result, nested elements will have unwanted styles. For example:

```html
<div class="p-10 bg-black text-white">
  <div class="ring-8 ring-offset-8 p-4">
    should have offset
    <div class="ring transform">shouldnt have offset</div>
  </div>
</div>
```

[UnoCSS Playground](https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIAOBaAjABigCMBzbIxGAaygBcBTAD1uwHcALMB9APgCgoUUJFiIUGAE5gAdmQAcUKbOwB7AGZqk9FgpwAWXgMFQk7FQFc4IKOwQR6UdZu1HBw6PGRp0SknQkI0khqKhIAtrymFlbStDZ2Dk5atMAA9OAQ-G7pkPxpGfxAA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6sfFJ1bX1enkwVjCJTqRwfZYD0Pa4pBhwNQBuGFDBowN99k5mqGqyIgqN4ZFyokoqPgCUssLyjCxsnDx8-OeXRwqS0n6HqHQS6xihAEYAHQAJheqAkaFweSqrDAjQA9Aj6CAgdUEZsWmoLuUkNikEA&options=N4IgzgLgTglgxhEAuaBXApgGhFdYAOA9gHZgwBu6yAZgIYA2YWI0tp1hUAtjQ0wL5A) - [Tailwind Playground](https://play.tailwindcss.com/wqHIZpB86T)

Expected (Tailwind Playground): Nested div shouldn't have ring offset
image

Actual (UnoCSS): Nested div have ring offset
image

This issue will appear for all CSS variables that are used in the output but do not have a redefinition, such as ring-color, ring offset color, etc.

antfu:
Oh I see. Thanks for the explanation and I will revert it

PR #1214, opened 2022-07-02
docs(preset-mini): theme usage docs

### LINKED ISSUE #1211: breakpoints deep merge not working
Hi there, 

I just ran into an issue when trying to configure custom breakpoints in a `Nuxt3` project.
If you configure your breakpoints like this : 

```ts
export default defineNuxtConfig({
    ...,
   unocss: {
     theme: {
        ...,
        breakpoints: {
          xs: '320px',
        },
     }
   }
})
```

All other breakpoints seem to be removed. You can check this commit that i added to my project : 

https://github.com/Shyrro/personnal-website/commit/36d6e24665966869718d9309d1f876bd6132df4a

If i uncomment the breakpoints config with the `xs` only, it doesn't work unless if i also uncomment the `md` breakpoint.

chris-zhu:
close #1211

### DIFF
--- a/packages/preset-mini/README.md
+++ b/packages/preset-mini/README.md
@@ -88,15 +88,17 @@ You can fully customize your theme property in your config, and UnoCSS will even
 <!--eslint-skip-->
 
 ```ts
-theme: {
-  // ...
-  colors: {
-    'veryCool': '#0000ff', // class="text-very-cool"
-    'brand': {
-      'primary': 'hsla(var(--hue, 217), 78%, 51%)', //class="bg-brand-primary"
-    }
-  },
-}
+presetMini({
+  theme: {
+    // ...
+    colors: {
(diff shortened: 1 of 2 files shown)


packages/preset-mini/README.md
```diff
@@ -82,6 +82,38 @@ will generate:
...
+  theme: {
+    // ...
+    colors: {
+      primary: '#ff0000',
+    },
+  },
+})
+```
+
+Notably, unless `breakpoint` properties. UnoCSS gives `breakpoint` control to the user, you need to fully customize the `breakpoint` properties that you want to use. For example:
```
antfu:
```suggestion
One exception is that UnoCSS gives full control of `breakpoints` to users. When a custom `breakpoints` is provided, the default will be overridden instead of merging. For example:
```

PR #1243, opened 2022-07-11
feat(reset): add tailwind.ce.css reset

Zhang-Wei-666:
基于 tailwind.css, 给 web components 使用, 移除了 html, body 样式


antfu:
Thanks, but we are probably not going to maintain two resets. Feel free to publish your own packages.

Issue #1248, opened 2022-07-13
[Feature Request] Customizing the class name


winter-ice:
ref: https://tailwindcss.com/docs/dark-mode#customizing-the-class-name

PR #1250, opened 2022-07-13
feat(preset-mini): customizing dark mode selector

Dunqing:
close: #1248

### DIFF
--- a/packages/preset-mini/src/index.ts
+++ b/packages/preset-mini/src/index.ts
@@ -11,7 +11,7 @@ export { parseColor } from './utils'
 
 export type { ThemeAnimation, Theme }
 
-interface DarkSelector {
+type DarkModeSelectors = ['class', {
   /**
    * @default '.light'
    */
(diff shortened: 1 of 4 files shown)


packages/preset-mini/src/index.ts
```diff
@@ -11,11 +11,23 @@ export { parseColor } from './utils'
...
+   */
+  dark?: string
+}
+
 export interface PresetMiniOptions extends PresetOptions {
   /**
    * @default 'class'
    */
-  dark?: 'class' | 'media'
+  dark?: 'class' | 'media' | ['class', DarkSelector]
```
antfu, line 30:
```suggestion
  dark?: 'class' | 'media' | DarkModeSelectors
```

packages/preset-mini/src/index.ts
```diff
@@ -11,7 +11,7 @@ export { parseColor } from './utils'
 
 export type { ThemeAnimation, Theme }
 
-interface DarkSelector {
+type DarkModeSelectors = ['class', {
```
antfu, line 14: I think a simple object is enough, the array feels redundant to me (plus they are not always `class` anyway)

Dunqing: Same as I thought, not sure why `tailwindicss` needs to set a `class`

Issue #1246, opened 2022-07-12
unocss does not generate a proper selector for `children:space-y-2` 


UltraCakeBakery:
See the playground/html below. I expected all red squares to have space between them, but they do not. 
This did work in version v0.30.1. Somewhere between that version and v0.44.0 it broke

playground: https://uno.antfu.me/play/?html=DwEwlgbgBAxgFmANiATgUwHYC4DOAHAQxjQFoBPEgJgD4AoKKUSOhhp6AIwHMT0QSArAAYhUAO4kAjKLhSh1YAHpwEFq3ZRuvNP2GiJ0qLOkLlzeozOqL7NWxWaefQSPFyjc0yruMHW53puhsbySt42VnRhzEA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAKq4ISAL5wsUCCDgByLpIDGdOsqRIMAD0ix02AIZcANvEw58REsFIAKZMNoMYdAFxwA2lTdRGAkIJwBKIVRqd0YWNk4ePn5wqgBdGTCkIA&options=N4IgzgLgTglgxhEAuaBXApgGhNAhgOzADMB7KAW2SNwBswsQp0wAHEwmAN3StvoF8gA

###### HTML
```html
<div children:space-y-2>
  <div>
    <div bg-red-500 w-10 h-10></div>
    <div bg-red-500 w-10 h-10></div>
  </div>
  <div>
    <div bg-red-500 w-10 h-10></div>
    <div bg-red-500 w-10 h-10></div>
  </div>
</div>
```
###### config
```js
import {
  defineConfig,
  presetAttributify,
  presetUno
} from 'unocss'

export default defineConfig({
  presets: [
    presetUno(),
    presetAttributify()
  ]
})

```

###### generated css:
```css
/* layer: default */
.w-10,
[w-10=""]{width:2.5rem;}
[h-10=""]{height:2.5rem;}
[children\:space-y-2=""]>:not([hidden])~:not([hidden])>*{--un-space-y-reverse:0;margin-top:calc(0.5rem * calc(1 - var(--un-space-y-reverse)));margin-bottom:calc(0.5rem * var(--un-space-y-reverse));}
.bg-red-500,
[bg-red-500=""]{--un-bg-opacity:1;background-color:rgba(239,68,68,var(--un-bg-opacity));}
```

[...]

PR #1252, opened 2022-07-13
fix(preset-mini): selector match variantSpaceAndDivide

edwardnyc:
fix [1246](https://github.com/unocss/unocss/issues/1246)

I'm not sure if the current solution is correct because I don't understand why the `~` selector was used before

### DIFF
--- a/packages/preset-wind/src/variants/misc.ts
+++ b/packages/preset-wind/src/variants/misc.ts
@@ -5,7 +5,7 @@ export const variantSpaceAndDivide: Variant = (matcher) => {
     return {
       matcher,
       selector: (input) => {
-        return `${input}>:not([hidden])~:not([hidden])`
+        return `${input}>:not([hidden])`
       },
     }
   }
--- a/test/__snapshots__/preset-wind.test.ts.snap
+++ b/test/__snapshots__/preset-wind.test.ts.snap
@@ -191,36 +191,36 @@ exports[`preset-wind > targets 1`] = `
 .break-inside-revert{break-inside:revert;}
 .break-after-column{break-after:column;}
 .break-after-unset{break-after:unset;}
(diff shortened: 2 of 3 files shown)


chris-zhu:
> fix [1246](https://github.com/unocss/unocss/issues/1246)
> 
> I'm not sure if the current solution is correct because I don't understand why the `~` selector was used before

I'm guessing it's because of the order in which the variants are resolved when multiple variants are used. Changes to variant order in previous PRs.
Although there are resolved,  but `space` is special, it has both rule and variant. 

Maybe this isn't a better way to resolved it.

chu121su12:
The `~` is so that the selector wont target the first element in the list.

packages/preset-wind/src/variants/misc.ts
```diff
@@ -5,7 +5,7 @@ export const variantSpaceAndDivide: Variant = (matcher) => {
     return {
       matcher,
       selector: (input) => {
-        return `${input}>:not([hidden])~:not([hidden])`
+        return `${input}>:not([hidden])`
```
chu121su12, line 8: Without `~`, `space-x` will set margin-left to the left-most element while `space-y` will set margin-top to the top-most element

chu121su12 (commented):
As chris-zhu said, there were changes in variant processing order that makes the linked issue tricky to solve for that particular combination

UltraCakeBakery:
@chris-zhu why was this PR closed?

chris-zhu:
> @chris-zhu why was this PR closed?

This isn't a right way to resolve it. And the PR haven't active for a long time, If you need, i will reopen it.

UltraCakeBakery:
> > @chris-zhu why was this PR closed?
> 
> This isn't a right way to resolve it. And the PR haven't active for a long time, If you need, i will reopen it.

This PR was made for my ticket #1246. I kept that ticket open by bumping it (to prevent stale bot), because this PR was still open at the time. I'll close my issue now so I don't have to keep bumping it anymore. If anyone needs a proper solution to #1246, feel free to request re-opening of that issue. You can work around this bug using custom rules.

PR #1265, opened 2022-07-14
fix(vscode): add html to the default includes

adamburgess:
As mentioned in the [readme](https://github.com/unocss/unocss#scanning):

> By default UnoCSS will scan for components files like: .jsx, .tsx, .vue, .md, **.html**, .svelte, .astro.

But it wasn't in the array. Feel free to close if this isn't necessary.

It was a bit confusing to me how index.html _was_ included when I was using the [vite plugin](https://github.com/unocss/unocss/blob/163be1b4a6a20020af2305e437405ceb8be99a98/packages/vite/src/transformers.ts#L13), but when I was using the vscode extension, it _wasn't_ included.

### DIFF
--- a/packages/shared-integration/src/defaults.ts
+++ b/packages/shared-integration/src/defaults.ts
@@ -1,4 +1,4 @@
 import { cssIdRE } from '@unocss/core'
 
 export const defaultExclude = [cssIdRE]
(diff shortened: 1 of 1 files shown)


antfu:
Vite handles HTML in a different pipeline. And yeah that make sense for the extension since the list is reused.

antfu (approved):

PR #1304, opened 2022-07-20
fix(preset-mini): compatible `bracket` parse

### LINKED ISSUE #1303: Preset Wind Missing Functionality `transition-[...]`
example
`transition-[color_background-color_border-color]`

should generate
```css
.transition-\[color_background-color_border-color\] {
    transition-property: color background-color border-color;
    transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
    transition-duration: 150ms;
}
```

chris-zhu:
This PR change this: 
- compatible `bracket` parse
- support css variable
close #1303


chu121su12:
How do you think about using `transition-[bracket]` and `transition-$variable` to do just raw transition [shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/transition#syntax)?

chris-zhu:
> How do you think about using `transition-[bracket]` and `transition-$variable` to do just raw transition [shorthand](https://developer.mozilla.org/en-US/docs/Web/CSS/transition#syntax)?

Of course this is better, but it would be a breaking change and not aligned with tw, wind.

PR #1307, opened 2022-07-20
more detailed config instructions for bundler

gotjoshua:


antfu:
Thanks but I think this is a bit too specific to your custom setup. While I think the TypeScript definition already serves as part of the documentation. If we do want to document it, I would expect a more general way.

gotjoshua:
> bit too specific to your custom setup.

understandable. 

Even though you are totally correct about the ts defs, I do think that some words about the difference between assigning to window.__unocss and calling initUnocssRuntime are in order.

PR #1316, opened 2022-07-22
fix:change button selector in reset tailwind.css

### LINKED ISSUE #1315: [type=button]  style preset in tailwind.css is not very friendly in some components
The property selector takes precedence over class, resulting in a strange look in some component libraries, which may be changed to the same as `@ unocss\ reset\ normalize.css`
```css
:where(button, input:is([type="button"], [type="reset"], [type="submit"])) {
    -webkit-appearance: button;
    background-color: transparent;
    background-image: none;
}
```

Or does not use`: where` and `:is`
```css
button, input[type="button"], input[type="reset"], input[type="submit"] {
    -webkit-appearance: button;
    background-color: transparent;
    background-image: none;
}
```

amihhs:
This PR change this:
- `reset/tailwind.css`
- When `< button type= "button" / >` is excluded, prevent the property selector from being higher than the class selector on button
fix #1315

### DIFF
--- a/packages/reset/tailwind.css
+++ b/packages/reset/tailwind.css
@@ -177,7 +177,7 @@ select {
 */
 
 button,
-[type='button'],
+:not(button)[type='button'],
 [type='reset'],
 [type='submit'] {
   -webkit-appearance: button; /* 1 */


antfu:
I think this is how Tailwind handles it,
image

Thanks but we probably should align with Tailwind at this moment.

PR #1353, opened 2022-07-30
fix(reset): set default tailwind.css border color

sibbng:


privatenumber:
1. This was a breaking change for me
2. I think it was better off as `currentColor`.
  Tailwind's preflight defaults to [`borderColor.DEFAULT` and falls back to `currentColor`](https://github.com/tailwindlabs/tailwindcss/blob/99b53b416085413a0b0459a37d57daac53cb9d4d/src/css/preflight.css#L12), where [`borderColor.DEFAULT` is `#e5e7eb` in the default theme](https://github.com/tailwindlabs/tailwindcss/blob/8d6e68228e0862ee7223932be681f19217881fdb/stubs/defaultConfig.stub.js#L184). Seems strange that this CSS reset is prioritizing a color from a specific theme in Tailwind. I think it should behave as if there is no theme (which is true), and use `currentColor`.

sibbng:
I'm strongly against using `currentColor` here. I knew this was going to be a breaking change for some people. But in my opinion,  if you are importing a CSS that is called `tailwind.css`, this should be what you expect to see.

> I think it should behave as if there is no theme (which is true), and use currentColor.

I wouldn't think that way if I'm not overriding the default `borderColor`. tailwind.css is not just a reset file. It's a compatibility layer.

It would be nice to handle this situation dynamically like Tailwind CSS and it seems straightforwardly easy with `preflights`. But this should be considered as a feature addition. I created [another issue ](https://github.com/unocss/unocss/issues/1385) to track this.

privatenumber:
Sorry my first point wasn't very clear. I was trying to say it probably should've warranted a major version bump (feature bump since it's still in major-zero) or a warning message in the release notes.

Thanks for the follow-up issue.

PR #1364, opened 2022-08-02
feat(web-fonts): add `bunny` web-fonts provider

chrisbbreuer:
I just swiftly whipped this up, and it's working well on my end so far.

Heading to bed now and will check this once more tomorrow. 

@antfu, if you see anything to be adjusted, would appreciate a comment 💪🏼

### DIFF
--- a/packages/preset-web-fonts/README.md
+++ b/packages/preset-web-fonts/README.md
@@ -70,12 +70,13 @@ Currently supported Providers:
 
 - `none` - do nothing, treat the font as system font
 - `google` - [Google Fonts](https://fonts.google.com/)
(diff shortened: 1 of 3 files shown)


Julien-R44:
Ahaha I was also about to do it yesterday before going to bed, but I finally preferred to sleep. And this morning I get up, ready to do it and I see your PR! So cool 😁

The only thing I have to say is that maybe we could have made a generic function that handles both Bunny and Google, because Bunny is a drop-in replacement for google fonts, so we could reuse the same code instead of duplicating the function. Not a big deal at all however

antfu (approved):

antfu (approved):

PR #1365, opened 2022-08-03
feat(runtime): try to decode page's html on default `extractAll` using regex

### LINKED ISSUE #1323: `@unocss/runtime`: arbitrary variant doesn't work on page load
Arbitrary variant doesn't work on page load, only takes effect (generates css) if were added in dynamically via hot reload (dev env) or Console HTML edit
Easily reproduced with basic .html file / online editor:
image
image

(After edited using Console:)
image

chu121su12:
This PR uses simple regex to replace only 3 encoded entities: `&`, `<` and `>`
Alternative to #1325
Closes #1323

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -1,6 +1,6 @@
 import type { UnoGenerator, UserConfig, UserConfigDefaults } from '@unocss/core'
 import { createGenerator } from '@unocss/core'
-import { autoPrefixer } from './utils'
+import { autoPrefixer, decodeHtml } from './utils'
 
 export interface RuntimeOptions {
   /**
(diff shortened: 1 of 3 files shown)


packages/runtime/src/index.ts
```diff
@@ -148,12 +150,13 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
       await scheduleUpdate()
   }
 
-  async function extractAll() {
+  async function extractAll(decode?: boolean) {
     const html = document.body && document.body.outerHTML
     if (html) {
-      await extract(html)
-      removeCloak()
+      await extract(decode ? decodeHtml(html) : html)
```
antfu, line 156: Can we just do it in one go without the options? like `extract(html + decodeHtml(html))`?

antfu (commented):
I prefer this solution 👍

chu121su12:
Marked as no longer breaking since the signature was reverted

antfu (approved):

PR #1371, opened 2022-08-04
feat(extractor): split backslash

XiNiHa:
Before this PR, given the string `"class=\"bg-white\""`, the default split extractor extracted `bg-white\` instead of `bg-white`.
This becomes problematic when extracting from SolidJS SSR output, which becomes an issue when using SolidJS with Astro.
This PR fixes the issue by including `\"` in the split token list.


XiNiHa:
I agree that this can be somewhat opinionated. I fully respect your opinion and if you don't like it, I'll just use it as a custom extractor in my projects :D

packages/core/src/extractors/split.ts
```diff
@@ -1,7 +1,7 @@
 import type { Extractor } from '../types'
 import { isValidSelector } from '../utils'
 
-export const splitCode = (code: string) => code.split(/[\s'"`;=]+/g).filter(isValidSelector)
+export const splitCode = (code: string) => code.split(/[\s'"`;=]|(?:\")+/g).filter(isValidSelector)
```
antfu, line 4:
```suggestion
export const splitCode = (code: string) => code.split(/[\s'"`;=]|(?:\\["'`])+/g).filter(isValidSelector)
```

XiNiHa:
Okay I guess `\` was not allowed in the rules name anyways XD

antfu:
Ah, we do have some cases that need backslash 🤔

XiNiHa:
```
 ❯ test/preset-mini.test.ts  (7 tests | 1 failed) 272ms
   ❯ test/preset-mini.test.ts > preset-mini > targets
     → expected [ '[content:attr(attr\_content)]' ] to deeply equal []
```
Looks like we should strictly check for `\\["']`?

antfu:
I guess we could do `/\\?[\s'"`;=]+/g

antfu:
Can you add a test for your case `"class=\"bg-white\""` so we won't break the usage accidentally in the future

PR #1394, opened 2022-08-09
feat: expose theme type var

### LINKED ISSUE #1270: Need useful types from the theme object
Is there any suported way to get access to the theme with useful types? When I import `presetUno` to try and use a `fontFamily` preset for instance, I might try this:
`const theme = presetUno().theme?.fontFamily` — but the type for `fontFamily` is `Record<string, string> | undefined`. One, it's a little inconvenient that `undefined` is one of the values, but two, it is inconvenient that I can't see exactly what's on that object.

I found that if I directly import the theme from the `dist` directory (which feels a little hacky), I can get much more user-friendly typings for the `theme` object:

```
import * as theme from "@unocss/preset-uno/dist/theme"
const {mono} = theme.fontFamily // works great because nothing is undefined and the type of the object seems to be inferred
```

I would really love to be able to have these more useful types available from the named exports rather than digging into the `dist` folder.

Also, as an aside, it'd also be useful to have more helpful types on the `colors` object than `Record<string, string>`

equt:
This PR takes a try to type `Theme` in all official packages correctly, instead of omitting it or setting it to `any`.

c.c. This should close #1270, and previous related PR #1002, #1003 

## Attentions Required

1. Currently, I type all themes as `T`, neither `T extends {}` nor `T extends Record<string, unknown>`, as this kind of restriction doesn't make any sense to me.
2. CI is broken, as there're several decisions that have to be made
  1. Currently, `@unocss/preset-uno` is the default preset, this means we cannot type the them e as a generic. The below code is theoretically not correct.
  https://github.com/unocss/unocss/blob/2180281ba36fab75b395330bf76c89f845c4e8df/packages/unocss/src/vite.ts#L8-L19

### DIFF
--- a/packages/autocomplete/src/create.ts
+++ b/packages/autocomplete/src/create.ts
@@ -5,7 +5,7 @@ import { parseAutocomplete } from './parse'
 import type { ParsedAutocompleteTemplate, UnocssAutocomplete } from './types'
 import { searchUsageBoundary } from './utils'
 
-export function createAutocomplete(uno: UnoGenerator): UnocssAutocomplete {
+export function createAutocomplete<T>(uno: UnoGenerator<T>): UnocssAutocomplete {
   const templateCache = new Map<string, ParsedAutocompleteTemplate>()
   const cache = new LRU<string, string[]>({ max: 5000 })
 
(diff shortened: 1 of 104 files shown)


equt:
I'm open to that. Without restriction provides more flexibility, and maybe some people are more comfortable with that. I would suggest extracting the `Theme` interface to a standalone package then.

smashercosmo:
Any plans to resurrect this PR?

Issue #1390, opened 2022-08-08
Is there some thought for dark/light theme doesn't only by `dark:`


Xwudao:
Now, not only `unocss` but also `wind`, `tailwind` support dark mode only with `dark` variant.

It means We should write more class name for supporting dark mode.

A better method for this that use `color palette` (refer to mui: https://mui.com/material-ui/customization/dark-mode/#main-content):

We design two color tokens, and automaticly change those tokens with compute mode changed or by user manual.

unocss has any plans to implement this?

antfu:
You can write your own variant to support that:

https://github.com/unocss/unocss/blob/2180281ba36fab75b395330bf76c89f845c4e8df/packages/preset-mini/src/variants/dark.ts#L1-L22

Xwudao:
@antfu Hi,

I mean I define two sets of colors in the configuration file. Then, at some point in time, uno automatically switches which set of colors to use.


----

Now I have a preliminary idea to use CSS variable, for example:

```html
<head>
    <style>
        body[data-theme="dark"] {
            --bg-main: rgb(167, 28, 28);
        }

        body {
            --bg-main: rgb(199, 189, 189);
        }

        .bg-gray {
            width: 40px;
            height: 40px;
            background-color: var(--bg-main);
        }
    </style>
</head>

<!-- <body data-theme="dark"> -->

<body>
    <div class="bg-gray">hello this</div>
</body>
```

---

I try to use in unocss:

```css
// index.css imported by main.ts
body{
  --bg-gray: red
}

```

```ts
//vite.config.ts
  Unocss({
      theme: {
          color: {
              main: 'var(--bg-gray)',
          }
      },
  })
```

```tsx
//App.tsx
<div className="bg-main">hello uno</div>
```

That's not work.

However, even if it works, it is not the best implement in my mind.

In my mind the best ways to implement is that we define two sets of colors in unocss config file.

example:

```ts
//vite.config.ts
  Unocss({
      theme: {
        dark: {
            color: {
                main: '#fff',
            }
        },
        light: {
            color: {
                main: '#f1f1f1',
            }
        }
      },
  })
```

I only use `bg-main`，the uno will automatic to switch which real color would be used.

antfu:
I see. I like your idea and I see the usage. However, it requires an overhaul of all the utilities and might break existing apps. I would recommend starting a new preset with this idea in mind. If you do, I am happy to list it as a community preset in the readme.

[...]

PR #1407, opened 2022-08-12
feat(prest-theme): support automatic theme switching

Dunqing:
close: #1390

Currently I have a problem, I want to collect the used `theme` in `postprocess` and return it in `preflights`. But `preflights` will be executed first.

What should I do?


antfu:
Thank you, but could we start with a community preset first? Deps on how it goes, we could see if we may include it officially.

Issue #1391, opened 2022-08-08
How to add autoprefixer with unocss?


zouhangwithsweet:

chris-zhu:
https://github.com/unocss/unocss/blob/2180281ba36fab75b395330bf76c89f845c4e8df/packages/preset-mini/src/index.ts#L43-L54

Do these two configuration items help you?

antfu:
Isn't it work already?

antfu:
Please provide a reproduction

antfu:
😅 minimal, please

antfu:
I would suggest: 

- Create a new project using just UnoCSS and PostCSS
- Adding a plain CSS and ensure PostCSS prefixed it
- Adding a corresponding UnoCSS util that doesn't work

to make the repro.

PR #1411, opened 2022-08-13
`user-select` Compatibility issues

lihe07:
> 'user-select' is not supported by Safari, Safari on iOS. Add '-webkit-user-select' to support Safari 3+, Safari on iOS 3+

Added two rules for static.ts


chris-zhu:
> > 'user-select' is not supported by Safari, Safari on iOS. Add '-webkit-user-select' to support Safari 3+, Safari on iOS 3+
> 
> Added two rules for static.ts

I guess you want use autoprefixer. There is a similar case #1391. [unocss-preset-autoprefixer](https://github.com/zouhangwithsweet/unocss-preset-autoprefixer) maybe help you.

Issue #1402, opened 2022-08-11
[unocss] does not found CSS placeholder in the generated chunks


Jelenkee:
I get this error message when I try to build my [Astro](https://github.com/withastro/astro) project where I use unocss.

Version: 0.45.6

There was a fix for #1341 but it didn't work for me.
I am not sure if this problem is caused by unocss, astro, svelte or even vite. Maybe you can help me.

I started debugging into unocss and found out that at this [line](https://github.com/unocss/unocss/blob/c2cdbd2fa75fb2bc21734b99805229eb7d52abb0/packages/vite/src/modes/global/build.ts#L154) the bundle contains a CSS-asset with css generated from my [svelte components](https://svelte.dev/docs#component-format-style). But this CSS doesn't contain the placeholder which unocss is looking for. So it cannot be replaced and the variable `replaced` remains `false`.
![image](https://user-images.githubusercontent.com/59470612/184192340-fe6b1b74-9689-44c1-bfdc-758033899674.png)

I tried to set the svelte option `emitCss` to `false` but it didn't work. Maybe it is overridden by the astro-svelte-plugin.

Dunqing:
Can you provide repo? I don't know astro.

Jelenkee:
There you go
https://github.com/CPUEater/bug-astro
Just run `npm run build`

Dunqing:
I tested it and found the same problem before version 0.43, I guess `unocss` doesn't support `astro`.

Dunqing:
See #1423.

Jelenkee:
Sorry, I don't understand your 'solution'.
Using `import uno.css` in a script tag should not work, because JS in script tags is executed in the browser only (according to the [docs](https://docs.astro.build/en/core-concepts/astro-components/#client-side-scripts)).

Just to clarify: I can only reproduce the problem with astro and svelte. When I comment out the css in the style tag in my svelte component, it works fine.

Dunqing:
Vite will transform `script`  refer [Using Hoisted Scripts](https://docs.astro.build/en/core-concepts/astro-components/#using-hoisted-scripts)

Dunqing:
> Just to clarify: I can only reproduce the problem with astro and svelte. When I comment out the css in the style tag in my svelte component, it works fine.

Because `svelte` is processed through `vite`

jolo-dev:
@Jelenkee it works what @Dunqing said.

[...]

PR #1423, opened 2022-08-16
feat(astro): introduce `@unocss/astro` integration 

Dunqing:
close: #1402

Just insert the following code into the *.astro page or layout component.

```html
<script>import 'uno.css'</script>.
```

### Note: 

The imported 'uno.css' must be included on all pages or layouts that use `unocss`, otherwise it will not take effect.

I'm not sure if this is the best solution, but it should be the simplest one that requires no changes at this time.


examples/astro/src/components/Card.astro
```diff
@@ -0,0 +1,76 @@
...
+	.link-card {
+		list-style: none;
+		display: flex;
+		padding: 0.15rem;
+		background-image: var(--link-gradient);
+		background-size: 400%;
+		border-radius: 0.5rem;
+		background-position: 100%;
+		transition: background-position 0.6s cubic-bezier(0.22, 1, 0.36, 1);
+	}
```
antfu, line 36: Could we remove all these inline styles to use UnoCSS exclusively?

examples/astro/README.md
```diff
@@ -0,0 +1,44 @@
+# Welcome to [Astro](https://astro.build)
```
antfu, line 1: We might remove the default readme, but instead, explain about how to integrate UnoCSS intro Astro and with the caveats you mentioned.

examples/astro/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    // Enable top-level await, and other modern ESM features.
+    "target": "ESNext",
+    "module": "ESNext",
+    // Enable node-style module resolution, for things like npm package imports.
+    "moduleResolution": "node",
+    // Enable JSON imports.
+    "resolveJsonModule": true,
+    // Enable stricter transpilation for better output.
+    "isolatedModules": true,
+    // Astro will directly run your TypeScript code, no transpilation needed.
```
antfu, line 12: We could remove these comments

antfu (changes requested):

antfu:
Thanks for the initiative. I have extended this PR to also introduce a new Astro integration so the requirement of importing `uno.css` is no longer a case.

antfu (approved):

Dunqing:
> Thanks for the initiative. I have extended this PR to also introduce a new Astro integration so the requirement of importing `uno.css` is no longer a case.

Thank you, this is the better solution!

aspiiire:
With the integration it doesn't build for me, it gives:
```
[unocss:global:build:generate] [unocss] does not found CSS placeholder in the generated chunks
This is likely an internal bug of unocss vite plugin
```
If I remove it from the integration and import unocss/vite on astro.config.mjs under vite.plugins and add import 'uno.css' everything works fine why?

Issue #1401, opened 2022-08-11
[BUG] unocss cannot generate css with Wepack@4.5.19 + Vue@2.x + VueRouter in production


zouhangwithsweet:
### systeminfo
- macOS 12.0.1 (21A559)
- node v14.18.0
- vue@^2.6.11
- vue-router@^3.2.0
- @vue/cli-* @~4.5.17
- unocss@
- @unocss/webpack@^0.45.6

### expect
- can generate css in dev & pro 

### actually
- works well in development
- **unocss only can generate the css utils referenced in `App.vue` in production.** The other css in components willnot generate

### minireproduction
https://stackblitz.com/edit/vitejs-vite-uperoz
You may need to download the project to build it locally

Issue #856, opened 2022-04-21
[vite] hot updated: /__uno.css


shiwanjun1995:
## main.js
import 'uno.css'
## result
chrome console tab
25285client.ts:352 [vite] hot updated: /__uno.css
## why?
<img width="566" alt="截屏2022-04-21 下午10 54 17" src="https://user-images.githubusercontent.com/38515714/164486018-cce829e8-276c-4529-aa45-c2cbe4360392.png">

antfu:
We temporarily close this due to the lack of enough information. Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) to reopen the issue. Thanks.

izhaoqing:
控制台一直输出 [vite] hot updated: /__uno.css，我这里也遇到了这个问题。

coderzhaoziwei:
我也遇到了这个问题，热更新好像是陷入了某个死循环停不下来，CPU开始升温。具体原因暂时没找到，公司项目无法提供出来。如果我有办法复现这个问题了，我会提交过来。

---

I also encountered this problem. The hot update seems to be stuck in an infinite loop and cannot stop, and the CPU starts to heat up.
The specific reason has not been found for the time being, and the company's project cannot be provided.
If I have a way to reproduce this problem, I will submit it.
(translate by google)

zgsgs:
我遇到的场景是: 在 `npm run dev` 运行环境下, 当使用 `git pull` 命令更新代码时会出现这类问题.

astronautchen:
I also encountered this problem

Dragonlord1005:
[Astro-boi](https://github.com/Dragonlord1005/astro-boi/tree/revamp-projects) happening on revamp project branch

kunjinW:
我也遇到了这个问题

PR #1428, opened 2022-08-17
fix(webpack): webpack4 optimize filter *

ChpShy:
fixes:  #1401  #1417

It is because `mini-css-extract-plugin` create a childCompiler that trigger the unocss's `optimizeAssets` before the main compiler, then unocss change the content. So, when the main trigger `optimizeAssets`, this haven't `LAYER_PLACEHOLDER_RE` anymore, and there is no way to update the latest tokens content to final asserts.

And `mini-css-extract-plugin`'s file name is `*`, so it works by filter `*` file. Or is there  more better solution?


ChpShy:
This actually could solve this two issues.
But I found that, it will be infinite compile when I run `pnpm run dev` in project `example/vue-cli`, simile #856.Then I create new project with the latest `vue-cli`, test the different `unocss`'s version. It get the same result.
Although it does not seem to be caused by this change, I need to look again.Or create a new issue about that.

ChpShy:
yes. I cloned your [minimal reproduction](https://stackblitz.com/edit/vitejs-vite-uperoz), and it works by my test.

antfu (approved):

PR #1430, opened 2022-08-17
fix(webpack): applyTransformers result maybe undefined

ChpShy:


antfu:
`result == null` already covers `undefined`.

PR #1454, opened 2022-08-22
fix: parsing html tag correctly

azaleta:
some css are missing during the parsing phase.
`<div m-1>` will be splitted into `<div` and `m-1>`
the further process will not recognize  `m-1>`  because of the html tag
(such as function  [counting css util](https://github.com/unocss/unocss/blob/main/packages/vscode/src/annotation.ts#L145]))

adding split logic for '>' and '/>' (it seems `<>` are having special meaning and could not included in regex?)

### DIFF
--- a/packages/core/src/extractors/split.ts
+++ b/packages/core/src/extractors/split.ts
@@ -1,7 +1,19 @@
 import type { Extractor } from '../types'
 import { isValidSelector } from '../utils'
 
-export const splitCode = (code: string) => code.split(/\\?[\s'"`;={}]+/g).filter(isValidSelector)
+export const splitCode = (code: string) => {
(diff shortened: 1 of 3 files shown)


antfu:
Emm. `<div m-1>` is handled by attributify extractor, the split extractor is for class usage.

azaleta:
I write the test case with a config contains presetUno only. 
I suppose attributify extractor will not be actived in this pattern. (without presetAttributify)
if so,  `m-1` should not be paring as matched css util 

Am I right?

[reproduction](https://github.com/azaleta/unotemp)

if you change App.vue line 6 to `<div ma >`  the unocss util count in status bar will change.

antfu:
Without attributify, `<div m-1 >` simply won't work anyway as it generate `.m-1 { ... }`

azaleta:
understood.Thanks!

PR #1463, opened 2022-08-24
docs: fix import in example

dominikg:


ydcjeff:
I don't think we need to scope as the installation above is using `unocss` and `astro` is exported as [sub-module in `unocss`](https://github.com/unocss/unocss/blob/47a0905daf042b0f56bf10eeaf71df95cba7f971/packages/unocss/package.json#L79).

dominikg:
oh, I got confused by the similar looking import in the example below. Sorry

PR #1468, opened 2022-08-26
feat(editor): editor highlight for variant groups

itpropro:
- Updated extractor regex
- Updated `getMatchedPositions`
- Added and updated tests

Currently, only the classes inside of the variant group are highlighted, not the variant itself (e.g. with `hover:(bg-red-400 h-4)`, `hover:` will still not be highlighted).


antfu (approved):

itpropro:
Thanks for improving @antfu, I'm just getting started with the codebase 🙌

PR #1484, opened 2022-08-28
feat(unocss): improve unocss pkg types

chris-zhu:


userquin:
@chris-zhu dts shouldn't be added also to the package exports entry?

chris-zhu:
> @chris-zhu dts shouldn't be added also to the package exports entry?

Can you guide me on what to do next ?

userquin:
> > @chris-zhu dts shouldn't be added also to the package exports entry?
> 
> Can you guide me on what to do next ?

```json
"exports": {
  ".": {
    "types": "./dist/index.d.ts", // <== this entry
    "main": "./dist/index.cjs",
    "module": "./dist/index.mjs"
  }
}
```
You will need also add the "files" entry if missing or just add the dts entry if missing  (i.e: https://github.com/antfu/vite-plugin-inspect/blob/main/package.json#L31):
```json
"files": { // <== this one
  "dist",
  "*.d.ts" // <== or this one
}
```

@antfu how about "typesVersion"? some packages here haven't the exports types entry nor files including the dts files (for example, inspector missing entries while preset-icons have the exports types but not the files types). We need to review all package.json files..

antfu (approved):

Issue #1489, opened 2022-08-29
Can i upgrade vite version for unocss to 3.0.9 now ?


sudongyuer:
![image](https://user-images.githubusercontent.com/76603360/187146596-6f93feb5-fecc-4c1f-b17b-9ad88686ebce.png)

PR #1490, opened 2022-08-29
chore: upgrade vite version

sudongyuer:
#1489

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2,6 +2,7 @@ lockfileVersion: 5.4
 
 overrides:
   '@nuxt/kit-edge': npm:@nuxt/kit@latest
+  vite: 3.0.4
 
 importers:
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -1,8 +1,7 @@
-lockfileVersion: 5.4
```
antfu, line 1: Please enable corepack

antfu (changes requested):

pnpm-lock.yaml
sudongyuer: Do i solve it tony ?

antfu (approved):

antfu:
The test is failing. I guess the override is still needed to make sure a single version of Vite installed.

sudongyuer:
> 

#1505

PR #1491, opened 2022-08-29
chore(alias): add some alias & order by file directory order

sudongyuer:


alias.ts
```diff
@@ -3,29 +3,35 @@ import { resolve } from 'path'
...
+  '@unocss/shared-integration': r('./packages/shared-integration/src/'),
+  '@unocss/transformer-attributify-jsx': r('./packages/transformer-attributify-jsx/src/'),
+  '@unocss/transformer-compile-class': r('./packages/transformer-compile-class/src/'),
   '@unocss/transformer-directives': r('./packages/transformer-directives/src/'),
   '@unocss/transformer-variant-group': r('./packages/transformer-variant-group/src/'),
-  '@unocss/transformer-compile-class': r('./packages/transformer-compile-class/src/'),
-  '@unocss/transformer-attributify-jsx': r('./packages/transformer-attributify-jsx/src/'),
+  'unocss': r('./packages/unocss/src/'),
   '@unocss/vite': r('./packages/vite/src/'),
+  '@vscode/unocss': r('./packages/vscode/src/'),
```
antfu, line 35: This won't be imported, shouldn't include it

alias.ts
```diff
@@ -3,29 +3,35 @@ import { resolve } from 'path'
...
+  '@unocss/shared-common': r('./packages/shared-common/src/'),
+  '@unocss/shared-docs': r('./packages/shared-docs/src/'),
+  '@unocss/shared-integration': r('./packages/shared-integration/src/'),
+  '@unocss/transformer-attributify-jsx': r('./packages/transformer-attributify-jsx/src/'),
+  '@unocss/transformer-compile-class': r('./packages/transformer-compile-class/src/'),
   '@unocss/transformer-directives': r('./packages/transformer-directives/src/'),
   '@unocss/transformer-variant-group': r('./packages/transformer-variant-group/src/'),
-  '@unocss/transformer-compile-class': r('./packages/transformer-compile-class/src/'),
-  '@unocss/transformer-attributify-jsx': r('./packages/transformer-attributify-jsx/src/'),
+  'unocss': r('./packages/unocss/src/'),
```
antfu, line 33: Why don't we just sort alphabetically?

sudongyuer: 🐻‍❄️ I think it's easy for contributors to know where the alias should go
image

antfu (approved):

alias.ts
sudongyuer: > 

Before that, I'd also like to implement a Node script that implements it 😂, If you want to do that, call me. I would like to do it

antfu:
Test failing

Issue #1509, opened 2022-09-01
theme.colors extend bug for base-only colors


NamesMT:
On theme.colors extend, base-only colors like `black` and `white` will be unable to use it's base style (ie: `bg-white`)

[Playground Reproduce](https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIAuBTAHpgWgCNEYBrKIgc2LgFdt0B6APgCgALMKTD7AJ2xA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAqrhJQL5xZQQhwDkAV1YBjOnX5IkGAB6RY6bAENBAG3iYc%2BIiWCkAFMlQwAFhhAYAXIiqoREVdDrWjqNwCNVSkQGsXtt1QaUCUoAE9rfgBiLFi4-kpA1HZEpM9BKxskt2CQUIiBKIAGAFYSgBYAMUqEgOTU%2BqoUqhp6Rmc4AG0A1oZmVn0ASlSAXSR2QaQgA&options=N4XyA)

Image from Playground Reproduce:
![image](https://user-images.githubusercontent.com/23612546/187857333-2541d948-d5ab-4161-8e3e-ffccb201a7c1.png)

(Old) Image:
![image](https://user-images.githubusercontent.com/23612546/187856421-45ea517a-e2fb-4f6c-b43c-af93001454aa.png)

NamesMT:
Would be fixed with a modification to [preset-mini/colors](https://github.com/unocss/unocss/blob/main/packages/preset-mini/src/_theme/colors.ts).
But I can't of the most efficient of doing it rn.

chris-zhu:
You rewrite this color, add `DEFAULT` key to make it work.

[...]

PR #1511, opened 2022-09-01
fix(preset-mini): extending `black` and `white` causing default color to lost (#1509)

NamesMT:


chris-zhu:
You can check this code. If `color` parsed was a `string` type, this color equal this value, or if was a `object`, get this `DEFAULT` key to value by default.

https://github.com/unocss/unocss/blob/805c29137510676524a69bd85aba0628b799ed95/packages/preset-mini/src/_utils/utilities.ts#L108-L116

PR #1513, opened 2022-09-01
feat(preset-mini): support `align-start` and `align-end`

sudongyuer:


sudongyuer:
> Would be better to have some explicit tests for them.

sudongyuer:
> Would be better to have some explicit tests for them.

Some tests have been added, please review

antfu (approved):

PR #1533, opened 2022-09-03
feat(playground): add reset option

sudongyuer:
Just open the link below for more details👇

https://jam.dev/c/965392e2-5e57-45ce-ae0c-11940a6a894e

image

### DIFF
--- a/playground/src/components/Editor.vue
+++ b/playground/src/components/Editor.vue
@@ -33,6 +33,11 @@ const panelSizes = useLocalStorage<number[]>(
   { listenToStorageChanges: false },
 )
 
+function handleReset() {
+  inputHTML.value = defaultHTML
+  customConfigRaw.value = defaultConfigRaw
+}
+
 function handleResize(event: ({ size: number })[]) {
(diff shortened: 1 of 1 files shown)


sudongyuer:
> Just open the link below for more details👇
> 
> https://jam.dev/c/965392e2-5e57-45ce-ae0c-11940a6a894e
> 
> image

By the way,I found that no one in UNOCSS/CLI is maintaining it. Can I join the team to maintain @unocss/cli ? Please give me this opportunity ~

antfu (approved):
It would be better to:

- Add confirmation before doing so
- Move the button to the left-most, and use `i-carbon-erase` icon otherwise it's could be a bit confusing to recognize as "refresh" or "undo"

sudongyuer:
> It would be better to:
> 
> * Add confirmation before doing so
> * Move the button to the left-most, and use `i-carbon-erase` icon otherwise it's could be a bit confusing to recognize as "refresh" or "undo"

Ok got it

sudongyuer:
Now it looks like this
image

PR #1534, opened 2022-09-03
perf(core): filter `rulesDynamic` in `resolveConfig`

QiroNT:
#### Changes to `@unocss/core`

- `rulesDynamic` is being filtered and reversed in `resolveConfig`, this yields a ~10ms improvement on the new benchmark.
- `extractorSplit` now first de-dupe via set then test the string. The impact is minimal, but faster on dedicated benchmark.
- Extractor results are merged via a for-loop instead of a `forEach`. The impact is again minimal, but it's nice to have.

#### Changes to the benchmark

- The new benchmark measures the full build time, from `buildStart` to `closeBundle`, instead of `buildEnd`.
- The new benchmark runs for 200 times instead of 50 times to give a more consistent result.
- Metric can also be set to percentiles, and it's now set to show the 75th percentile result which is more representative of the real-world performance.
- The source now includes a bit of HTML to test the splitter. Tho I think it should also have some garbage, I'll leave that to a future person.

The old benchmark is... flawed at best.
Since both TailwindCSS and WindiCSS generate their CSS before `buildEnd`, and UnoCSS generates it's CSS in `generateBundle` which is after `buildEnd`. The benchmark doesn't actually measure the full build time for UnoCSS, but only part of the extraction time.
To make the benchmark result more realistic, and better for future comparisons, the new one measures the full build time, which now includes CSS generation time.


packages/core/src/config.ts
```diff
@@ -66,15 +66,19 @@ export function resolveConfig(
...
+        const prefix = rule[2]?.prefix || ''
+        rulesStaticMap[prefix + rule[0]] = [i, rule[1], rule[2], rule]
+        // delete static rules so we can't skip them in matching
+        // but keep the order
+        return undefined
+      }
+      return [i, ...rule]
+    })
+    .filter(Boolean)
+    .reverse() as ResolvedConfig['rulesDynamic']
```
chris-zhu, line 81: Wouldn't there be a performance overhead here?

packages/core/src/extractors/split.ts
```diff
@@ -1,12 +1,12 @@
 import type { Extractor } from '../types'
 import { isValidSelector } from '../utils'
 
-export const splitCode = (code: string) => code.split(/\\?[\s'"`;={}]+/g).filter(isValidSelector)
+export const splitCode = (code: string) => [...new Set(code.split(/\\?[\s'"`;={}]+/g))].filter(isValidSelector)
```
chris-zhu, line 4: I think it does the same job as his name, just to simply split the code. Deduplication should be handed over to the next step. (Just my opinion 😄)

packages/core/src/generator/index.ts
```diff
@@ -45,7 +45,10 @@ export class UnoGenerator {
 
     for (const extractor of this.config.extractors) {
       const result = await extractor.extract(context)
-      result?.forEach(t => set.add(t))
+      if (result) {
+        for (const token of result)
+          set.add(token)
+      }
```
chris-zhu, line 51: It seems to be no different from the original code, and the original number of lines is less.

packages/core/src/generator/index.ts
```diff
@@ -434,25 +435,24 @@ export class UnoGenerator {
...
-    for (let i = rulesSize - 1; i >= 0; i--) {
-      const rule = rulesDynamic[i]
-
-      // static rules are omitted as undefined
-      if (!rule)
-        continue
+    const { rulesDynamic } = this.config
 
+    // match rules
+    for (const [i, matcher, handler, meta] of rulesDynamic) {
```
chris-zhu, line 441: UnoCSS parse rules is from last to first, right?

packages/core/src/generator/index.ts
```diff
@@ -645,7 +646,8 @@ function applyScope(css: string, scope?: string) {
 export function movePseudoElementsEnd(selector: string) {
   const pseudoElements = selector.match(/::[\w-]+(\([\w-]+\))?/g)
   if (pseudoElements) {
-    pseudoElements.forEach(e => (selector = selector.replace(e, '')))
+    for (const e of pseudoElements)
```
chris-zhu, line 649: Did you use some formatting tools? (It's not related to the question. I prefer es6 syntax 😅)

chris-zhu (commented):
Some suggest

antfu:
Nice catch! Thanks

antfu (approved):

Issue #1539, opened 2022-09-04
Bug: any modification to root files triggers VSCode extension reload


kidonng:
This paired with #1540 is pure nightmare fuel.

OS: Linux (GitHub Codespaces), but I can replicate this behavior on macOS as well
VSCode:  1.71.0
UnoCSS extension: v0.45.15

Vite config:

```js
import { defineConfig } from 'vite'
import unocss from 'unocss/vite'

export default defineConfig({
  // Must pass an option or else the extension won't load
  //               ↓
  plugins: [unocss({})]
})
```

Demo:

https://user-images.githubusercontent.com/44045911/188325144-35d645e5-ba61-4b90-af66-078950c8de5d.mov

TrickyPi:
fixed by #1546, this issue could be closed.

PR #1546, opened 2022-09-05
fix(vscode): improve trigger reload config

azaleta:
#1539


antfu:
Instead exposing a function, I think it's better to directly expose the array of config file list

azaleta:
fixed

PR #1555, opened 2022-09-06
feat(runtime): add preset-icon

mokeyish:


packages/preset-icons/src/cdn.ts
```diff
@@ -2,7 +2,19 @@ import type { IconifyJSON } from '@iconify/types'
 import { loadIcon } from '@iconify/utils/lib/loader/loader'
 import { searchForIcon } from '@iconify/utils/lib/loader/modern'
 import type { UniversalIconLoader } from '@iconify/utils/lib/loader/types'
-import { $fetch } from 'ohmyfetch'
+
+const myFetch = async function myFetch(url: string) {
+  if (typeof window !== 'undefined' && window.fetch) {
+    // browser
+    const res = await window.fetch(url)
+    return await res.json()
```
userquin, line 10: shouldn't check for res and also for status? if the cdn is down/missing or the response is not 20X you will get weird errors:
```ts
return res && res.ok 
  ? await res.json()
  : throw new Error(`unexpected/missing response from ${url}: ${res?.status ?? 'no response'}`)
```
```

Comment (mokeyish) [reply packages/preset-icons/src/cdn.ts]: @userquin I have update it.

**Comment by userquin** [comment]
@mokeyish why this PR? what are you trying to solve?

**Comment by mokeyish** [comment]
> @mokeyish why this PR? what are you trying to solve?

I want to use preset-icon with cdn import in browser directly.

**Comment by jd-solanki** [comment]
IG this will also help in cases where icons' list come from API and render those unknown icons.

Basically, this will allow generating icons that arent' generated at build time.

**Comment by userquin** [comment]
`ohmyfetch` will work in `node` and in `browser`: we've it working in the interactive docs, check it in the interactive folder:

imagen

**Comment by userquin** [comment]
@mokeyish I was thinking that this PR will solve this another one issue: https://github.com/unocss/unocss/issues/1242

@jd-solanki this PR is just using `fetch` instead `ohmyfetch` if in browser...

**Comment by mokeyish** [comment]
图片


@userquin I used it here, not that playground.

**Comment by mokeyish** [comment]
I build `runtime` with script `build:cdn` in `https://github.com/unocss/unocss/blob/main/packages/runtime/package.json`, and import `full.global.js`, it throws `required http` error.

**Comment by mokeyish** [comment]
> > @userquin I used it here, not that playground.
> 
> then, you should solve the comment here: https://github.com/unocss/unocss/pull/1555/files#diff-0b6b72858957081dc7f339d0b71a592b58fff569e3a337d58bfab4fe1b76e1cfR23

This line will throw error. `ohmyfetch` didn't detect browser environment automatically.

https://github.com/unocss/unocss/blob/f5255f523eb109a6fbb00423185911bde0a4877a/packages/preset-icons/src/cdn.ts#L5

**Comment by userquin** [comment]
https://github.com/unjs/ohmyfetch/blob/main/src/index.ts

**Comment by mokeyish** [comment]
@userquin 
Maybe you should try this yourself.

图片

**Comment by mokeyish** [comment]
I don't know why `build:cdn` compiled with node -> https://github.com/unjs/ohmyfetch/blob/03016df975aa945658e07f4d5dd59d59f76205ad/package.json#L11-L14

Do yout known how to configure it ?

**Comment by userquin** [comment]
@mokeyish can you try to build from root folder? the build script will build both, node and cdn

**Comment by mokeyish** [comment]
@userquin No, it not work, still throw error.

**Comment by mokeyish** [comment]
> > @userquin I used it here, not that playground.
> 
> then, you should solve the comment here: https://github.com/unocss/unocss/pull/1555/files#diff-0b6b72858957081dc7f339d0b71a592b58fff569e3a337d58bfab4fe1b76e1cfR23

I have solve the comment above in another PR #1559

**Comment by mokeyish** [comment]
It seems that vite building(unocss playground) use  [ohmyfetch/index.ts](https://github.com/unjs/ohmyfetch/blob/main/src/index.ts), So [https://uno.antfu.me/?s=i-twemoji-frog](https://uno.antfu.me/?s=i-twemoji-frog) no problem.  but `build:cdn` use [ohmyfetch/node.ts](https://github.com/unjs/ohmyfetch/blob/main/src/node.ts) , it not work in browser.

**Comment by userquin** [comment]
@mokeyish there are 2 distinct problems: to include the runtime in the browser you will need to exclude/externalize a few modules (node ones from iconify: check the nuxt.config.ts module in the interactive folder) and the error building this repo, I have never get any error.

The playground is using Vite/Rollup to build it, check the vite.config.ts (`optimizedDeps.exclude` and rollup options): also excluding/externalizing iconify node stuff.

**Comment by mokeyish** [comment]
> @mokeyish there are 2 distinct problems: to include the runtime in the browser you will need to exclude/externalize a few modules (node ones from iconify: check the nuxt.config.ts module in the interactive folder) and the error building this repo, I have never get any error.
> 
> The playground is using Vite/Rollup to build it, check the vite.config.ts (`optimizedDeps.exclude` and rollup options): also excluding/externalizing iconify node stuff.

The package `runtime` use `tsup` to build, replace it with `vite`?

**Comment by mokeyish** [comment]
@userquin I add  option `--platform browser` to tsup, it throws a lot of errors 😢
图片

File: packages/preset-icons/src/cdn.ts [hunk 2]
```diff
@@ -2,7 +2,23 @@ import type { IconifyJSON } from '@iconify/types'
 import { loadIcon } from '@iconify/utils/lib/loader/loader'
 import { searchForIcon } from '@iconify/utils/lib/loader/modern'
 import type { UniversalIconLoader } from '@iconify/utils/lib/loader/types'
-import { $fetch } from 'ohmyfetch'
+
+const myFetch = async function myFetch(url: string) {
```
antfu, line 6: IIRC, `ohmyfetch` already fallback to native fetch automatically.

packages/runtime/src/cdn/full.ts
```diff
@@ -11,6 +12,9 @@ init({
       presetUno(),
       presetTypography(),
       presetTagify(),
+      presetIcons({
+        cdn: 'https://esm.sh/',
```
antfu, line 16: I am not sure to hard-code this. Since the CDN mode of preset-icons is not self-included, I would recommend asking users to use the bundler approach for their custom needs. That's why we didn't include it by default at the beginning.

antfu:
As the review comments. We might better document that the full build does not include the icons preset and provide an example setup of including icons instead. Thanks for bringing this up.

PR #1559, opened 2022-09-06
feat(preset-icons): load collection  dynamically

mokeyish:


antfu:
This will slow down the loading (collections.json contains many info that we don't need), adding more possibility of network failure. I'd prefer to get the list at build time similar to https://github.com/antfu/unplugin-icons/blob/main/scripts/prebuild.ts

Issue #1566, opened 2022-09-07
IDEA & Webstorm Eslint ^8.23.0 error


sudongyuer:
<img width="1387" alt="image" src="https://user-images.githubusercontent.com/76603360/188917581-47054388-6877-4b1b-84ed-596f4b6f06ba.png">

## References
stackoverflow: https://stackoverflow.com/questions/73509984/eslint-typeerror-this-liboptions-parse-is-not-a-function

Eslint: https://github.com/eslint/eslint/issues/16250

jetbrains: https://youtrack.jetbrains.com/issue/WEB-57089/ESLint823-TypeError-thislibOptionsparse-is-not-a-function

gaojiangtao-git:
这个问题  webstorm  问题,解决方案:1.降级eslint  版本即可 8.21, 2.webstorm  下一个版本更新

userquin:
@sudongyuer use the lint and lint-fix scripts from root (`pnpm run lint` or `pnpm run lint-fix`)

PR #1567, opened 2022-09-07
chore: IDEA eslint compatibility

sudongyuer:
#1566

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "bumpp": "^8.2.1",
     "codemirror": "^5.65.6",
     "codemirror-theme-vars": "^0.1.1",
-    "eslint": "^8.23.0",
+    "eslint": "8.22.0",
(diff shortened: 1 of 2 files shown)


antfu:
I am not sure. I think it's better to have IDEA fix it instead of reverting the version on our side

userquin:
@antfu will be fixed in 2022.2.3 so there is no need to update here: https://youtrack.jetbrains.com/issue/WEB-57089

sudongyuer:
> @antfu will be fixed in 2022.2.3 so there is no need to update here: https://youtrack.jetbrains.com/issue/WEB-57089
> 
> EDIT: it is 2 versions from current, this will take time... and it is a bad experience since we lose dx

+1

userquin:
@antfu IntelliJ 2022.2.2 released and seems the error is gone, so we can revert this.

PR #1568, opened 2022-09-07
chore: add missing `types` exports field

kidonng:
Some plugins are missing `types` field in `package.json`'s `exports`, which is required for [TypeScript](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing) to work under `"module": "node16"`.

~~Note: I removed types added in #1484 because they are never actually referenced (or used).~~

### DIFF
--- a/packages/astro/package.json
+++ b/packages/astro/package.json
@@ -25,6 +25,7 @@
   "sideEffects": false,
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "require": "./dist/index.cjs",
       "import": "./dist/index.mjs"
     }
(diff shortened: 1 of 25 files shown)


ydcjeff:
Can you revert the types that you have removed? They are required to work under `moduleResolution: "node"`.

kidonng:
I was confused by the way TypeScript deal with `exports` LOL. (To begin with, I'm unfamiliar with packages with multiple entry points).

`unocss` types are restored as requested.

ydcjeff (approved):

PR #1572, opened 2022-09-08
chore: switch default web font provider

PatrickHeneise:
Google Fonts tracks >20% of the web traffic. No need to add to that.

Screenshot 2022-09-08 at 10 00 54 AM

Source: https://whotracks.me/trackers/google_fonts.html

Signed-off-by: Patrick Heneise <patrick@zentered.co>

### DIFF
--- a/packages/preset-web-fonts/README.md
+++ b/packages/preset-web-fonts/README.md
@@ -16,7 +16,7 @@ Unocss({
   presets: [
     presetUno(),
     presetWebFonts({
-      provider: 'google', // default provider
+      provider: 'bunny', // default provider
       fonts: {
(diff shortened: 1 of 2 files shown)


antfu:
Thanks but I would need to reject this change. As the providers already provide the capability of using bunny if you want to, I don't think it's safe to flip this behavior for every user.

PatrickHeneise:
I didn't know of bunny before. Would be nice to see this in a major change, so it doesn't break behavior.

antfu:
I didn't know it either, so I can't make promise to count on it's availability. It would be used with users' explicit awareness

Issue #839, opened 2022-04-19
separator config


Youdaman:
Is there a way to set the separator character?

e.g. https://tailwindcss.com/docs/configuration#separator has:
```js
module.exports = {
  separator: '_',
}
```

However trying to add this to unocss.config.ts warns:
> Object literal may only specify known properties, and 'separator' does not exist in type 'UserConfig<{}>'.ts(2345)

I'm wanting to use Pug in my templates which does not accept `:` in class names.

My current workaround is instead of:
```pug
    .shrink-0.w-4.sm:w-48
```
I instead write:
```pug
    .shrink-0.w-4(class="sm:w-48")
```

antfu:
In variants, `:` and `-` are interchangeable. Hope that fits what you need.

Youdaman:
Perfect! Thank you.

PR #1573, opened 2022-09-08
Config separator

tjk:
Need support for custom separator to migrate from tailwind (https://github.com/unocss/unocss/issues/839#issuecomment-1102437795 is not good enough)

If we expect just one separator char instead of the default `[:-]`, it should be *faster*.

TODO:
- [ ] fix https://github.com/unocss/unocss/pull/1573#issuecomment-1240438066
- [ ] there are more places where variants just *implicitly* use `:` (eg. https://github.com/unocss/unocss/blob/341a70da81d5fd72e66250113fa9b64ae0091b61/packages/preset-attributify/src/variant.ts#L5)
- [ ] fix `regexClassGroup` to use config.separator
- [ ] fix tests
- [ ] add test(s) customizing config.separator

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -115,5 +115,6 @@ export function resolveConfig(
     shortcuts: resolveShortcuts(mergePresets('shortcuts')),
     extractors,
     safelist: mergePresets('safelist'),
+    separator: config.separator || '[:-]',
   }
 }
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -308,6 +308,13 @@ export interface ConfigBase<Theme extends {} = {}> {
   rules?: Rule<Theme>[]
 
   /**
+   * Variant separator (regex substring, eg. [:-] or _)
+   *
+   * @default '[:-]'
+   */
+  separator?: string
+
+  /**
    * Variants that preprocess the selectors,
    * having the ability to rewrite the CSS object.
    */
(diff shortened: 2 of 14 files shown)


chu121su12:
Before you continue, I'd point out there's special case for backdrop pseudo that only uses `:` because using `-` clashes with the backdrop rule
https://github.com/unocss/unocss/blob/b9fb172419608098af5db9cd72932790c6f06c6b/packages/preset-mini/src/_variants/pseudo.ts#L87

tjk:
Thank you for pointing that out and providing quick feedback!

I think based on that, we should essentially try to use the configured separator, excluding `-` -- is that right?

tjk:
Also, I'd rather just make this config option a 1 character thing but not sure of the history and if anyone would ever count on having both : and - at same time.

chu121su12:
My personal take for now is for the `separator` to accept `string[]` thus the regex is or-ed (ex. `(-|:)`), and modify the exception on pseudo to exclude `-` instead now hardcoding `:`.

tjk:
Yea I actually already have a change to make it a string that just gets [] wrapped in regex interpolation to mostly satisfy the main case where it is just one char (but leaving the default as ":-" with the exclusion happening as you suggest)... a bit torn on whether to just make it an array though. Once again really not sure what use cases there are for it to be more than one char.

chu121su12:
Use case of string separator (as in multiple chars) may just be as compatibility with tailwind
https://github.com/tailwindlabs/tailwindcss/blob/0b5bfc8065e928caa7cf11a7bac37fda10411b6f/tests/arbitrary-variants.test.js#L322-L324
https://github.com/tailwindlabs/tailwindcss/blob/0b5bfc8065e928caa7cf11a7bac37fda10411b6f/tests/arbitrary-variants.test.js#L348-L350

Otherwise a nice addition in term of flexibility.

chu121su12:
@tjk If you are yet working on this PR, can I work on the updates?

tjk:
> @tjk If you are yet working on this PR, can I work on the updates?

Oh yes sorry I got distracted with work!

PR #1582, opened 2022-09-12
fix(transformer-attributify-jsx): ignore the className part

Flower-F:
When I tried to use the UnoCSS in a React project today, I found a bug of the `transformer-attributify-jsx`. The transformer-attributify-jsx will change the attribute to `xxx=""`, such as `text-red=""`, to avoid the limitation of React JSX. But it also changes the className part by mistake. For example, I might write some code snippet like this:
```tsx
<div className={red ? 'text-red' : 'text-green' }></div>
```
But it would be transformed to something like this:
```tsx
<div className={red ? 'text-red=""' : 'text-green=""' }></div>
```
So I can't get the correct result. And this PR is to solve this problem, as it will ignore the className part when transforming jsx.


SnowingFox:
nice

chris-zhu (approved):

PR #1596, opened 2022-09-14
fix(web-fonts): remove duplicate fonts in font-family string

everyx:


antfu:
What's that issue is this trying to solve? I am not sure if it's worth the complexity here.

everyx:
> What's that issue is this trying to solve? I am not sure if it's worth the complexity here.

Yes, it's just follow the [`font-family-no-duplicate-names` rule from stylelint](https://stylelint.io/user-guide/rules/list/font-family-no-duplicate-names/), so, it's up to you, you can close this PR if you don't need it :)

antfu:
I think it's a bit overkill, and technically `"Robot"` is different from plain `Robot`. Thank you but I am closing this for now.

PR #1592, opened 2022-09-14
feat(cli): support for transformers

wkeylin:
fix: #1369

PR #1602, opened 2022-09-15
fix(cli): remove filename of input

lzt1008:
When using `[...fileCache]`to convert Map to an array, the key of `fileCache` as filename, will also be passed. That's not what we want.


chris-zhu:
@lzt1008  #1592 also handle a part of the this code, you can reveiew it. I think it would be better to merge the commits together

PR #1610, opened 2022-09-18
chore: refactor vite plugins integration

sudongyuer:


antfu:
How would this help?

sudongyuer:
> How would this help?
image

Oh~ I just think this will be more readable,this maintains consistency. If don't like this , just ignore it. sorry

antfu:
`Extra` would sound too ambiguous. I am not against this change, but the same argument could apply backward. Thanks but I don't think this is necessary.

PR #1617, opened 2022-09-20
feat(core)!: generate preflights after utilties

Dunqing:
### Motivation

PR: #1407

I want to collect the used theme in postprocess and return it in preflights. But preflights will be executed faster.

### Purpose

Add theme variables collected by `postprocess` to preflight

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -174,7 +174,8 @@ export class UnoGenerator {
       }
     })
 
-    const preflightPromise = (async () => {
+    await Promise.all(tokenPromises)
+    await (async () => {
       if (!preflights)
         return
 
@@ -205,11 +206,6 @@ export class UnoGenerator {
         )),
       )
     })()
-
-    await Promise.all([
-      ...tokenPromises,
-      preflightPromise,
-    ])
 
     const layers = this.config.sortLayers(Array
       .from(layerSet)
--- a/test/generate-async.test.ts
+++ b/test/generate-async.test.ts
@@ -31,7 +31,7 @@ describe('generate-async', () => {
         [/^rule$/, () => new Promise(resolve => setTimeout(() => {
           order.push(1)
           resolve('/* rule */')
-        }, 1000))],
+        }, 20))],
       ],
       preflights: [
         {
--- a/test/generate-async.test.ts
+++ b/test/generate-async.test.ts
@@ -24,21 +24,21 @@ describe('generate-async', () => {
     expect(order).eql([1, 2])
   })
 
-  test('preflight-first', async () => {
+  test('preflight at the end', async () => {
     const order: number[] = []
     const uno = createGenerator({
       rules: [
         [/^rule$/, () => new Promise(resolve => setTimeout(() => {
-          order.push(2)
+          order.push(1)
           resolve('/* rule */')
-        }, 20))],
+        }, 1000))],
       ],
       preflights: [
         {
-          getCSS: () => new Promise(resolve => setTimeout(() => {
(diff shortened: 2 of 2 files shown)


test/generate-async.test.ts
```diff
@@ -23,26 +23,4 @@ describe('generate-async', () => {
     await uno.generate('rule')
     expect(order).eql([1, 2])
   })
-
-  test('preflight-first', async () => {
```
antfu, line 27: Why?

Dunqing: Now parseToken first

antfu: This will be a breaking change - Can we still keep the test but instead update the order?

Dunqing: > This will be a breaking change - Can we still keep the test but instead update the order?

Yes, updated

test/generate-async.test.ts
```diff
@@ -24,21 +24,21 @@ describe('generate-async', () => {
...
+  test('preflight at the end', async () => {
     const order: number[] = []
     const uno = createGenerator({
       rules: [
         [/^rule$/, () => new Promise(resolve => setTimeout(() => {
-          order.push(2)
+          order.push(1)
           resolve('/* rule */')
-        }, 20))],
+        }, 1000))],
```
antfu, line 34: Do we need the timeout to be this long?

Dunqing: I just wanted to make it look like it was at the end.

packages/core/src/generator/index.ts
```diff
@@ -174,7 +174,8 @@ export class UnoGenerator {
       }
     })
 
-    const preflightPromise = (async () => {
+    await Promise.all(tokenPromises)
+    await (async () => {
```
chu121su12, line 178: Why not use out of bound promise to wait for the variables from the postprocess'? This defeats the purpose of racing tokens vs preflights as preflights might often be slower especially due to use of disk or network; compared to mostly cpu-bound token generation.

Dunqing: Thank for you review.
Is there any way let us know the postprocess is all finished？

chu121su12: Good point, since waiting for all postprocess cannot easily be detected due to the destructure
https://github.com/unocss/unocss/blob/a8d4db6c4c7c6e1fdbb00d98f7dd282063a4cec6/packages/core/src/generator/index.ts#L200-L201

I'd propose to upgrade the existing event system to add communications from token/preflight generation. Something [like this](https://github.com/chu121su12/unocss/commit/27838248b317b8b946b7322488e6ae2198c31929).

Dunqing: Agreed, this would be useful by the way

chu121su12:
As @chris-zhu said in #1695, there are another way to wait until either tokens or preflights are all processed before doing the other. Try to get access to the `UnoGenerator` then you can do:

- token only: `UnoGenerator.generate($tokenVar, { preflights: false })`
- preflights only: `UnoGenerator.generate('', { preflights: true })` (generate on an empty string)

You can combine both calls. Token generation uses cache until you update the config with `UnoGenerator.setConfig()` and preflight generation is without caching.

antfu:
I am actually fine with this PR since it doesn't add extra complexity. @chu121su12 What's your opinion on that? Do you mean you prefer a more general solution?

chu121su12:
@antfu Just with the async behavior as I commented above.

With this changes the whole test file could be removed since it was added to ensure the racing behavior.

chu121su12 (approved):

PR #1630, opened 2022-09-24
chore: no webpack and vite peerDependencies warning for nuxt

kecrily:


antfu:
Probably no longer needed with https://github.com/unjs/unplugin/commit/6d3d5af6b00eef367ab03182229a9efa125bd61b

kecrily:
But `@unocss/nuxt` depends on `@unocss/webpack` and `@unocss/vite`, and they each require peerDependencies (vite and webpack).

PR #1632, opened 2022-09-25
fix: use `pathe`

sudongyuer:


packages/vscode/package.json
```diff
@@ -69,6 +69,9 @@
     "dev": "tsup --watch src",
     "publish": "esno ./scripts/publish.ts"
   },
+  "dependencies": {
+    "pathe": "^0.3.7"
```
antfu, line 73: This should go devDeps as VS Code is not a npm package.

antfu:
Is there any real issue to resolve?

packages/vscode/package.json
sudongyuer: resolved

sudongyuer:
> Is there any real issue to be resolved by this?
When i write some tests on windows to contribute, be possible to occur. 
image

antfu (approved):

Issue #566, opened 2022-01-30
unocss config for setting  arrays of globs/pattern to watch with outfile name?


goldcoders:
I'm using unocss cli

and for me to generate css from html i need to do this

```
 "unocss": "unocss \"layouts/**/**/*.html\" --watch -o ./assets/css/uno.css",
```

is there a way i can passed multiple instance of 
paths to watch and output file name?

is there a unocss.config for this?

i dont wanna end up with my solution like this

```
 "uno-home": "unocss \"layouts/index.html\" --watch -o ./assets/css/home.css",
"uno-blog": "unocss \"layouts/blog/*.html\" --watch -o ./assets/css/blog.css",
"css-watch": "run-p uno-home uno-blog"
```

just to generate a css for home.css, blog.css

and watch 2 instance in parallel to create 2 files....

i think it would be great if we can pass on the unocss.config.ts/js

an array of path patterns to watch and paths to output the file css

I think it has good usecase for jamstacks , espcially hugo framework...

but the config now dont have that features

sudongyuer:
I am working on this

PR #1633, opened 2022-09-25
feat: support unocss config for cli

### LINKED ISSUE #543: [feature] recursive file scan from specific folders
Hi. I am using unocss cli & This issue is all about adding some useful cli options.

-  add `-d --directory` option with following value formats `"folder1:folder2:..."` or `"folder1 folder2 ..."` or `[folder1,folder2,...]` .
-  add `-r --recursive` option to scan for file recursively down to the folder tree.
-  add `-e --extension` options to scan for files with specific extensions with following value formats `"*.html:*.js:*.jsx:..."` or `"*.html *.js *.jsx ..."` or `[*.html,*.js,*.jsx,...]`.

Note: I also should mention that unocss fails to create required folder recursively when it is trying to save output to a folder that does not exist.

sudongyuer:
resolve #543 #566

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -512,6 +512,13 @@ export interface UserOnlyOptions<Theme extends {} = {}> {
   envMode?: 'dev' | 'build'
 }
 
+/**
+ * For unocss-cli config
+ */
+export interface CliOptions {
+  cliOptions?: CliOptionItem[]
+}
+
 export interface UnocssPluginContext<Config extends UserConfig = UserConfig> {
   ready: Promise<LoadConfigResult<Config>>
   uno: UnoGenerator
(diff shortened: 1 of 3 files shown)


packages/core/src/types.ts
```diff
@@ -512,6 +512,13 @@ export interface UserOnlyOptions<Theme extends {} = {}> {
   envMode?: 'dev' | 'build'
 }
 
+/**
+ * For unocss-cli config
+ */
+export interface CliOptions {
+  cliOptions?: CliOptionItem[]
```
antfu, line 519:
```suggestion
  cli?: {
    entry?: CliEntryItem | CliEntryItem[]
  }
```

sudongyuer: I changed the api , tony have a look🌈~

Issue #1644, opened 2022-09-26
vscode extension doesn't work after upgrating to 0.45.23


Flower-F:
After upgrating the extension to version 0.45.23, the extension doesn't work anymore with TSX.
![屏幕截图 2022-09-26 220246](https://user-images.githubusercontent.com/78215016/192297517-2e120ecc-2b8f-44ee-91f5-889f584445b0.png)
Revert to version 0.45.21, there isn't any problem.
![屏幕截图 2022-09-26 220126](https://user-images.githubusercontent.com/78215016/192297743-7fb07e2a-b2ac-493e-9b2c-c8e85472ac2b.png)
Here is the [reproduction](https://github.com/Flower-F/uno-vscode-reproduction). There isn't anything special, just a normal React template created by Vite. And I think it's because of this [commit](https://github.com/unocss/unocss/commit/b8c8844a).

joseehilton147:
same here

chris-zhu:
I'm not sure the reason about that commit, but it works well in my vue project.

Flower-F:
> I'm not sure the reason about that commit, but it works well in my vue project.

I tried a Vue3 project, and the issue still exsits when upgrating to version 0.45.23. And I use the Windows system, are you using the Mac OS?

sudongyuer:
This [commit](https://github.com/unocss/unocss/commit/b8c8844a) seems not resovle this issue , the issue #1565 still exist

Flower-F:
> And this [commit](https://github.com/unocss/unocss/commit/b8c8844a) seems not resovle this issue , the issue #1565 still exist

I think it might need to make a revert just as sxzz said.(

azaleta:
The problem is not in #1565 
PR is coming #1645

Flower-F:
> @Flower-F you are on Windows, am I right?

Yes

sudongyuer:
> The problem is not in #1565 PR is coming #1645

Yes, you are right , thx

PR #1645, opened 2022-09-26
fix(vscode): windows pathe format

azaleta:
Linked issue: #1644

Linked Commit: #1632 

function `pathe:dirname` will change the slash of path
https://github.com/unjs/pathe/blob/main/test/index.spec.ts#L86

This cause unequal on Windows and could not load config file.
image


sudongyuer (approved):

packages/vscode/src/contextLoader.ts
```diff
@@ -250,3 +251,11 @@ export class ContextLoader {
     return this.defaultContext
   }
 }
+
+// Util to normalize windows paths to posix
+function normalizeWindowsPath(input = '') {
```
chris-zhu, line 256: How about move it to `utils` file?

sudongyuer:
@chris-zhu do you think it's necessary to remove pathe.

PR #1649, opened 2022-09-26
feat: use turbo improve build speed

sudongyuer:


sudongyuer:
Sometimes, when rerun local build script can  see how fast it's going while the cache is hit. 
image

antfu:
For some reason it never hit cache on my machine:
image

It only does for `build-post`
image

Issue #1646, opened 2022-09-26
text-size totally ignored if defined inside rules or shortcuts


aspiiire:
I don't understand the logic, I have defined many rules and shortcuts, but the ones with text-size-[Xpx] get ignored totally, if I apply inline classes with text-size-8 for example it works perfectly.

This are rule examples that I tried to test it out... what I-am doing wrong?

```
[/^ts-(d+)$/, ([, d]) => `text-size-[4px]`],
['ts-8', () => ({ 'font-size': '25rem' })],
```

aspiiire:
Ok thanks, I have provided an example of what I want to achieve on the playground link that you have shared

chris-zhu:
> Ok thanks, I have provided an example of what I want to achieve on the playground link that you have shared

Where is the link?

aspiiire:
Sorry how can I share my code from this playground?

chris-zhu:
> Sorry how can I share my code from this playground?

Share the URL link.

antfu:
We temporarily close this due to the lack of enough information.
Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) to reopen the issue.
Thanks.

> [Why reproduction is required](https://antfu.me/posts/why-reproductions-are-required)

PR #1660, opened 2022-09-29
feat(playground): add share button

chris-zhu:
Because of this comment. https://github.com/unocss/unocss/issues/1646#issuecomment-1261094530


playground/src/components/Editor.vue
```diff
@@ -129,6 +139,12 @@ onMounted(() => {
           </div>
 
           <div class="pl-1 ml-auto space-x-2 text-sm md:text-base flex items-center flex-nowrap">
+            <button
+              :class="copied ? 'i-carbon-checkmark-outline text-green' : 'i-carbon-share'"
```
antfu, line 143: We are using ri icons in the playground

PR #1672, opened 2022-10-01
fix(vite): improve `svelte-scoped` mode

### LINKED ISSUE #1671: extractorSvelte doesn't work in 'svelte-scoped' mode (w/ repro)
Resuming #663 - `svelte-scoped` mode and `svelteExtractor` do not work together as explained in this [comment](https://github.com/unocss/unocss/issues/663#issuecomment-1260373110).

I have created a minimal reproduction by means of forking this repo and creating PR #1672 that contains an additional sveltekit example demonstrating using `svelte-scoped` mode that is titled `sveltekit-scoped`. If you undo my changes to `svelte-scoped.ts` you'll see the bug in that the BG Color bar will not change to red, nor will the 2nd logo fly in.

The bug fix was accomplished by passing the filename (id) into the `uno.generate` method in `SvelteScopedPlugin`.

I've also updated the default sveltekit example which gets it working again after SvelteKit made a large number of breaking changes on the road to Release Candidate. Now that they are in RC, I think the example will be safe until v2. This example uses a global stylesheet and is mostly identical in function to the prior example.

jacob-8:
Fixes #1671


examples/sveltekit-scoped/README.md
```diff
@@ -0,0 +1,12 @@
+# SvelteKit (release candidate) + UnoCss Vite Plugin (svelte-scoped mode)
+
+- Works with @sveltejs/kit@1.0.0-next.506 (release candidate) and @sveltejs/adapter-auto@1.0.0-next.80
+- Uses PresetUno and PresetIcons
+- Uses `svelte-scoped` mode
+- Uses `extractorSvelte` to be able to use `class:red-bg-200={true}` in components
+
+## Notes
+
+- To use preflights add `<style uno:preflights global></style>` to your root `+layout.svelte`
```
antfu, line 10: Thanks for the addition! Can we also have the docs about the attribute to the Vite plugin? Thanks

jacob-8: Yes, I will add that now. I'm assuming you are talking about https://github.com/unocss/unocss/tree/main/packages/vite#sveltekit.

examples/sveltekit/vite.config.ts
```diff
@@ -0,0 +1,44 @@
+// @ts-nocheck
```
jacob-8, line 1: I don't like adding this, but for some odd reason I get this error on my `config` object: `Excessive stack depth comparing types '{ plugins: (Plugin_2[] | Plugin_2[])[]; }' and 'UserConfig'.ts(2321)` - I don't know how to debug and assume maybe it's because I'm using unocss via linking instead of installing. I don't know, but I don't think it hurts the purpose of the examples.

examples/sveltekit-scoped/README.md
jacob-8: Added. How does https://github.com/unocss/unocss/pull/1672/commits/68033cb3c4b1e2533e3cc3ddc273a31254b018c5 look?

examples/sveltekit/vite.config.ts
antfu: It's because multiple Vite installed and the type is not aligned.

jacob-8: Ok, I understand and see how you fixed it by using defineConfig(). 

Sorry for all the capitalization, spacing, and semicolon oddities. I'm used to using ESLint + Prettier and am still figuring out how your repo functions. Thanks for the cleanup and merge!

PR #1692, opened 2022-10-07
feat(vite)!: svelte-scoped automatic compilation

### LINKED ISSUE #1676: Ideas for better Svelte Scoped support
UnoCSS works GREAT for use in Svelte projects that want to use a global `uno.css` import, but the `svelte-scoped` mode doesn't work well. There is a need for a robust scoped mode not only to help keep global app css to a minimum but also css scoping is essential for component libraries built with SvelteKit so that only the CSS necessary for the components in use adds to the page weight. For my own repos, I'm going to make a new `svelte-scoped` mode that's more robust. Please let me know if you would appreciate it as either an upgrade of the current `SvelteScopedPlugin` or an alternative `SvelteScopedCompiledPlugin`.

I would like it to use class name compilation all the time without having to write in the `:uno:` trigger needed in [@unocss/transformer-compile-class](https://github.com/unocss/unocss/tree/main/packages/transformer-compile-class/). In normal component composition, Svelte's CSS scoping via adding a specific component hash onto every style by default is great because it allows you to easily style an input in the component with `input { padding: 3px; }` and not even having to define a class. But Svelte's CSS scoping quickly breaks down when using the more complicated classes that utility CSS frameworks provide like `rtl:__`, `dark:__` and `.space-x-1` as these classes depend on parent and children components.

Let me walk through the proposed behavior. This will serve as an outline for future tests that would be needed.

## Parent dependent classes like `rtl:` and `dark:`

Uno (+ Tailwind, Windi) will turn `rtl:right-0` into `[dir="rtl"] .rtl\:right-0{right:0rem;}` which the Svelte compiler will turn into `[dir="rtl"].svelte-vzopk9 .rtl\:right-0.svelte-vzopk9{right:0rem}` and strip out because that component doesn't have any elements with `[dir="rtl"]`. To solve this problem, I need Uno to generate `:global([dir="rtl"] .uno-3hashza){right:0rem}` which the Svelte comiler will turn into the desired `[dir="rtl"] .uno-3hashza{right:0rem}`. Uno's hash serves the same purpose as Svelte's component hash except it lets the layout's `dir` attribute influence children element styles. 

Proposed Before:
```svelte
<div class="ltr:left-0 rtl:right-0">
</div>
```

Proposed After:
```svelte
<div class="uno-3hashza">
</div>

<style>
  :global([dir="ltr"] .uno-3hashza) {
    left: 0rem;
  }
  :global([dir="rtl"] .uno-3hashza) {
    right: 0rem;
  }
</style>
```

## Children affecting classes like `.space-x-1` 

This is the same problem except with children components. If an element in your component wants to add space between five `Button` children components it won't work because Uno compiles `.space-x-1` into `.space-x-1>:not([hidden])~:not([hidden])` which Svelte compiles into `.space-x-1.svelte-dly1zh>.svelte-dly1zh:not([hidden])~.svelte-dly1zh:not([hidden])` and strips out as it does not find at least two sibling children elements with that specific component's hash class. 

Proposed Before:
```svelte
<div class="space-x-1">
  <div>Logo</div>
  <Button>Login</Button>
</div>
```

Proposed After:
```svelte
<div class="uno-3hashza">
  <div>Logo</div>
  <Button>Login</Button>
</div>

<style>
  :global(.uno-3hashza > :not([hidden]) ~ :not([hidden])) {
    --un-space-x-reverse: 0;
    margin-left: calc(0.25rem * calc(1 - var(--un-space-x-reverse)));
    margin-right: calc(0.25rem * var(--un-space-x-reverse));
  }
</style>
```

*Because of these first two situations (parent component and child component dependent classes), I consider a `svelte-scoped` mode without compilation to be useless because eventually you will run into one of these two situations unless you have a simple app, in which case you would probably be better served by a global mode.*

## Nested Component styles

Sometimes it's nice to have the flexibility of adding the `class` prop to a component which will place those styles in a child component using `class="{$$props.class} foo bar"`. Some external libraries also offer a class prop on their components for additional customization in this manner. This won't work using a component's default Svelte style hash.

Proposed Before:
```svelte
<Button class="px-2 py-1">Login</Button>
```

Proposed After:
```svelte
<Button class="uno-4hashza">Login</Button>

<style>
  :global(.uno-4hashza) {
    padding-left:0.5rem;
    padding-right:0.5rem;
    padding-top:0.25rem;
    padding-bottom:0.25rem;
  }
</style>
```

## Conditional `class:` feature

Class names added using Svelte's class directive feature, `class:text-sm={bar}`, will also be compiled.

Proposed Before:
```svelte
<div class:text-sm={bar}>World</div>
```

Proposed After:
```svelte
<div class:uno-2hashza={bar}>World</div>

<style>
  :global(.uno-2hashza) {
    font-size: 0.875rem;
    line-height: 1.25rem;
  }
</style>
```

The class directive shorthand usage of `class:text-sm` where `text-sm` is both a class and a variable is also supported. In my opinion this is a bad habit, but the plugin will change `class:text-sm` into `class:uno-2hashza={text-sm}`.

## Already existing features

The component would continue to keep the benefits of the current [@unocss/transformer-compile-class](https://github.com/unocss/unocss/tree/main/packages/transformer-compile-class/):
- Unmatched classes like `foo` will be ignored and left alone inside a class attribute
- If two different elements have the same tokens in their class string (regardless of order) they will be given the same compiled class name.

## Summary

The examples listed above are just a few of the situations where a scoped mode breaks down when styles are dependent on parent or children elements. Rather than write [custom variants](https://github.com/unocss/unocss#custom-variants) for every situation that is broken, it's simpler to use a compilation mode. As well, it often makes for a smaller component size as it is more common to have fewer elements with many classes than many elements with fewer classes. In those situations where you do have many elements with fewer classes, the classes tend to be repetive, like `.mb-3` and those will only need one style rule to address them all, the same as uncompiled mode. 

To summarize this:

```svelte
<div class="bg-red-100 text-lg">Hello</div>

<div class:text-sm={bar}>World</div>

<div class="fixed flex top:0 ltr:left-0 rtl:right-0 space-x-1 foo">
  <div class="px-2 py-1">Logo</div>
  <Button class="py-1 px-2">Login</Button>
</div>

<style>
  div {
    --at-apply: text-blue-500 underline;
  }
  .foo {
    color: red;
  }
</style>
```

will be transformed into this:

```svelte
<div class="uno-1hashza">Hello</div>

<div class:uno-2hashza={bar}>World</div>

<div class="uno-3hashza foo">
  <div class="uno-4hashza">Logo</div>
  <Button class="uno-4hashza">Login</Button>
</div>

<style>
  :global(.uno-1hashza) {
    --un-bg-opacity: 1;
    background-color: rgba(254, 226, 226, var(--un-bg-opacity));
    font-size: 1.125rem;
    line-height: 1.75rem;
  }

  :global(.uno-2hashza) {
    font-size: 0.875rem;
    line-height: 1.25rem;
  }

  :global(.uno-3hashza) {
    position: fixed;
    display: flex;
  }
  :global([dir="ltr"] .uno-3hashza) {
    left: 0rem;
  }
  :global([dir="rtl"] .uno-3hashza) {
    right: 0rem;
  }
  :global(.uno-3hashza > :not([hidden]) ~ :not([hidden])) {
    --un-space-x-reverse: 0;
    margin-left: calc(0.25rem * calc(1 - var(--un-space-x-reverse)));
    margin-right: calc(0.25rem * var(--un-space-x-reverse));
  }

  :global(.uno-4hashza) {
    padding-left:0.5rem;
    padding-right:0.5rem;
    padding-top:0.25rem;
    padding-bottom:0.25rem;
  }
  
  div {
    --un-text-opacity: 1;
    color: rgba(59, 130, 246, var(--un-text-opacity));
    text-decoration-line: underline;
  }

  .foo {
    color: red;
  }
</style>
```

After which the Svelte compiler will remove the `:global()` wrappers, then add it's own scoping hash just to the `div` and `.foo` rules.

## Other Notes

- Because compilation is universal, a stricter class extractor will be used to avoid picking up and converting what UnoCSS thinks are matching class names. For example, at present the manually written style `font-variant-numeric: tabular-nums;` triggers a false positive and causes UnoCSS's default preset to add `.tabular-nums` to the component's styles. This would be a disaster if the plugin were to rewrite this style as `font-variant-numeric: .uno-4hashza;`. As such, I don't presently have a clear plan on how this plugin could interop with Attributify mode. 

- The UnoCSS Inspector doesn't work in the current `svelte-scoped` mode because it depends on adding matched classes to the list of tokens and files to the list of modules. Since `svelte-scoped` mode doesn't do anything globally, the proper method of feeding classes and their files to the UnoCSS context object that the Inspector pulls from needs implemented.

- Creating a component library via a SvelteKit project and the `svelte-package` tool (the recommended Svelte method) doesn't use Vite at all so UnoCSS as a Vite plugin will have no effect when packaging up a library for sharing. If the plugin proposed above is written correctly, it would be simple to add a tiny wrapper to make it possible to alternatively use it as a svelte preprocessor that could be placed into the `svelte.config.js` file. This would allow UnoCSS to be used for component libraries as well as allow the VSCode Svelte Extension's "Show Compiled Code" preview to reflect the changes made by UnoCSS. *This would make it trickier to interop with the UnoCSS Inspector, but perhaps that can be addressed later.*

jacob-8:
Closes #1676

I've written test cases for every situation described in #1676 and a few less common ones as well. As can be seen by running the tests and the updated `svelte-scoped` example project, the `SvelteScopedPlugin` is now fully ready for use in exactly the same manner as is available using global styles, except that each component carries the weight of its own CSS.

Merging this will prepare the way for #1718.


packages/vite/src/modes/svelte-scoped/transform.ts
```diff
@@ -0,0 +1,129 @@
...
+        return
+      const className = queueCompiledClass([token])
+      const start = match.index! + 'class:'.length
+      s.overwrite(start, start + match[1].length, `${className}={${token}}`)
+    }
+
+    if (s.hasChanged())
+      code = s.toString()
+
+    // TODO: return a source map
```
jacob-8, line 109: Does the transformSFC function need to return a source map? The old one didn't but I think it maybe should be doing that. I can look into it later in a future PR, but would be open to advice on it this time around.

examples/sveltekit-scoped/package.json
```diff
@@ -14,6 +14,7 @@
     "@sveltejs/kit": "next",
     "@unocss/core": "link:../../packages/core",
     "@unocss/preset-icons": "link:../../packages/preset-icons",
+    "@unocss/transformer-directives": "^0.45.26",
```
antfu, line 17: Better use a link here

jacob-8: I wanted to do that but I don't understand the error I receive when I do:

```
failed to load config from C:\dev\unocss\unocss\examples\sveltekit-scoped\vite.config.ts     
error when starting dev server:
Error: TRANSFORM_ERROR: C:\: Export namespace should be first transformed by `@babel/plugin-proposal-export-namespace-from`.
C:/dev/unocss/unocss/node_modules/.pnpm/css-tree@2.2.1/node_modules/css-tree/lib/index.js:0:0    
at C:/dev/unocss/unocss/packages/transformer-directives/src/index.ts:4:16
at jiti (C:\dev\unocss\unocss\node_modules\.pnpm\jiti@1.15.0\node_modules\jiti\dist\jiti.js:1:247497)    
at file:///C:/dev/unocss/unocss/packages/transformer-directives/dist/index.mjs:4:71      
at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
at async Promise.all (index 0)    
at async ESMLoader.import (node:internal/modules/esm/loader:533:24)    
at async loadConfigFromBundledFile (file:///C:/dev/unocss/unocss/examples/sveltekit-scoped/node_modules/.pnpm/vite@3.1.6/node_modules/vite/dist/node/chunks/dep-db16f19c.js:63505:21)     
at async loadConfigFromFile (file:///C:/dev/unocss/unocss/examples/sveltekit-scoped/node_modules/.pnpm/vite@3.1.6/node_modules/vite/dist/node/chunks/dep-db16f19c.js:63400:28)            
at async resolveConfig (file:///C:/dev/unocss/unocss/examples/sveltekit-scoped/node_modules/.pnpm/vite@3.1.6/node_modules/vite/dist/node/chunks/dep-db16f19c.js:63017:28)    
at async createServer (file:///C:/dev/unocss/unocss/examples/sveltekit-scoped/node_modules/.pnpm/vite@3.1.6/node_modules/vite/dist/node/chunks/dep-db16f19c.js:62083:20)
```

packages/vite/src/modes/svelte-scoped/transform.ts
jacob-8: Ok, in working on #1719 I decided to figure it out myself so I just updated the transformSvelteSFC function to return sourcemaps when applicable.

examples/sveltekit-scoped/package.json
jacob-8: Just discovered that I don't need to add that to the package.json at all because it's already in core. So I removed it.

jacob-8:
I put this back into draft mode as I've been using this in production via my own temporarily published packaged (#1725), and I made a couple changes that I want to document before merge.

jacob-8:
@antfu Can you please take another look at this to see if it's good to merge or give advice on what to change? 

I'd love to help with some of the other SvelteKit related issues like #1774 as well as dig into further explorations of how to smoothly integrate the Inspector and typography plugin in a context where a global stylesheet is not the default place to put all styles. But I think this PR is already big enough.

antfu (approved):
This is awesome, thank you!

test/svelte-scoped.test.ts
```diff
@@ -23,19 +25,40 @@ describe('svelte-scoped', () => {
...
 
-  async function transform(code: string, { combine } = { combine: true }) {
-    return (await transformSvelteSFC(code, 'Foo.svelte', uno, { combine }))?.code
+  async function transform(code: string, { combine = true, format = true } = {}) {
+    const transformed = (await transformSvelteSFC(code, 'Foo.svelte', uno, { combine }))?.code
+    if (transformed && format) {
+      return prettier(transformed, {
+        parser: 'svelte',
+        plugins: [prettierSvelte],
+      })
```
jacob-8, line 35: @antfu, wow, thank you for this! So much easier to read the test results. Initially I tried `prettierCSS` as seen elsewhere but obviously that didn't work since there's more than CSS in the test results.

PR #1695, opened 2022-10-08
feat(core): generator events

chu121su12:
Add events after each of tokens and preflights are generated. This should help usecases like #1617.

I'm unsure what data should be passed to the event. As for now, tokens & preflights event gets what each of them would normally produce, except in a cloned set to prevent modification.

### DIFF
--- a/test/generate-async.test.ts
+++ b/test/generate-async.test.ts
@@ -46,3 +46,81 @@ describe('generate-async', () => {
     expect(order).eql([1, 2])
   })
 })
+
+describe('firing-events', () => {
+  test('config-event', async () => {
+    const order: number[] = []
+    const uno = createGenerator()
+
+    uno.events.on('config', () => order.push(order.length))
+    expect(order).eql([])
+
+    uno.setConfig({})
+    expect(order).eql([0])
+  })
+
+  test('tokens-generated-event', async () => {
+    const order: number[] = []
+    const uno = createGenerator({
+      rules: [
+        [/^rule$/, () => new Promise(resolve => setTimeout(() => {
+          order.push(1)
+          resolve('/* rule */')
+        }, 10))],
+      ],
+      preflights: [
+        {
+          getCSS: () => new Promise(resolve => setTimeout(() => {
+            order.push(2)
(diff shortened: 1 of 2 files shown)


Dunqing:
Should `this` and `tokens/preflight` be passed to the event?

Dunqing:
Also, I think the event names should have keywords like `after` or `end`? because these events emit after all the work is finished.

chu121su12:
@Dunqing I was copying the `config` event, so pardon me for the naming. `tokens` event send out the tokens that are generated in the current generate cycle, so does the `preflights`.

test/generate-async.test.ts
```diff
@@ -46,3 +46,81 @@ describe('generate-async', () => {
     expect(order).eql([1, 2])
   })
 })
+
+describe('firing-events', () => {
```
antfu, line 50: We could have a separate events.test.ts for this.

test/generate-async.test.ts
```diff
@@ -46,3 +46,81 @@ describe('generate-async', () => {
...
+  test('config-event', async () => {
+    const order: number[] = []
+    const uno = createGenerator()
+
+    uno.events.on('config', () => order.push(order.length))
```
antfu, line 55: I am not very sure what `order.length` is used for, maybe we could just use a fixed number?

antfu (commented):
Do you consider this an alternative to https://github.com/unocss/unocss/pull/1617 or a complement to it?

I don't have a strong opinion on this, but my concern is that this will increase the package size for a relatively rare use case. And honest I would prefer for rules to keep as pure as possible and not rely on the event system. (e.g. the `config` hook is for up-level integrations but is not designed to be access directly in rules)

chris-zhu:
If the event system is too heavy, what about using lifecycles in the config?

chu121su12:
> ...this will increase the package size for a relatively rare use case. And honest I would prefer for rules to keep as pure as possible and not rely on the event system...
I didn't realize the `x.event` is only used for specific integration. And if that's the case, I'd choose not to pursue this addition; Except there's more reason otherwise

PR #1698, opened 2022-10-08
feat(preset-mini)!: update @media variant

chu121su12:
With `getComponent` helper, `media-x` variant can more safely accept bracket value.

This PR remove the auto variable for the media-x value, instead opting for bracket to achieve similar functionality. Bracket value is now supported and theme lookup is unchanged.

This also aligns `media-x` syntax with `support-x` syntax


chris-zhu (approved):

chu121su12:
Closed in favor of #1704

PR #1701, opened 2022-10-08
feat(nuxt): add `inspector` option support

### LINKED ISSUE #1700: feat(nuxt): add support for `inspector` option
Support for `@unocss/inspector` in nuxt. Would be happy to make a PR 👍🏽

Lexpeartha:
Resolves #1700


Lexpeartha:
Running tests throws following error:
```
ERROR: [plugin: externalize-deps] Failed to resolve entry for package "@unocss/vite". The package may have incorrect main/module/exports specified in its package.json.
```

Is it expecting `@unocss/vite` to be a dependency?

packages/nuxt/src/inspector.ts
```diff
@@ -0,0 +1,24 @@
...
+import * as chalk from 'chalk'
+import pluginInspector from '../../inspector/node'
+
+export function addInspector() {
+  const nuxt = useNuxt()
+  const inspectorRoute = '/_unocss'
+
+  if (nuxt.options.dev && nuxt.options.builder === '@nuxt/vite-builder') {
+    nuxt.options.vite.plugins = nuxt.options.vite.plugins || []
+    nuxt.options.vite.plugins.push(pluginInspector({} as never))
```
antfu, line 12: Isn't the Vite plugin already enabling the inspector? I am not sure if this is needed here. We might just need to pass the option to Vite plugin

Lexpeartha: what is the option you are referring to?

Lexpeartha:
Also does inspector support webpack? from what I've seen it doesn't seem like it does?

antfu:
No, the inspector is a Vite-only feature.

Lexpeartha:
> No, the inspector is a Vite-only feature.

In that case I think it's done, the only thing left is to fix tests, which I'm not really sure why are they broken.

packages/nuxt/src/index.ts
```diff
@@ -25,6 +27,7 @@ export default defineNuxtModule<UnocssNuxtOptions>({
     attributify: false,
     webFonts: false,
     icons: false,
+    inspector: false,
```
antfu, line 30: It's true by default in Vite plugin, any reason to have it default to false? https://github.com/unocss/unocss/blob/c1aa6d03b9c4cd7124582348df0e42d48639da66/packages/vite/src/types.ts#L7

packages/nuxt/src/inspector.ts
```diff
@@ -0,0 +1,20 @@
...
+  const nuxt = useNuxt()
+  const inspectorRoute = '/__unocss'
+
+  if (nuxt.options.dev && nuxt.options.builder === '@nuxt/vite-builder') {
+    nuxt.hook('listen', (_, listener) => {
+      const logger = useLogger()
+
+      const viewerUrl = `${withoutTrailingSlash(listener.url)}${inspectorRoute}`
+
+      logger.info(`UnoCSS inspector available at: ${
```
antfu, line 15: I am not so sure about if the log is necessary.

Lexpeartha: I thought it would be cool to do that since Tailwind module does something very similar?
That also makes it very easy to check upon the inspector from the console where process is running

sudongyuer (approved):

Issue #1681, opened 2022-10-04
Only apply hover styles when support


Miles-C:
This solves the ["sticky hover" problem](https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/) where tapping something that includes a hover style on mobile will persist that hover style until tapping elsewhere.

copy this code for your unocss.config.ts
```js
// unocss.config.ts
variants: [
    (matcher) => {
      // The prefix cannot start with "-" or "hover", you can customize the prefix.
      if (!matcher.startsWith('&hover:')) return matcher
      return {
        // slice `hover:` prefix and passed to the next variants and rules
        matcher: matcher.slice(7),  // The 7 here represents the number of characters in the prefix.
        parent: [`@media (hover: hover) and (pointer: fine)`],
        selector: (s) => `${s}:hover`,
      }
    },
  ],
```

when using:
```html
<!-- Use prefix: &hover -->
<div class="&hover:text-red &hover:dark:text-blue">Hello World</div>
```

CSS Generated：
```css
@media (hover: hover) and (pointer: fine)
.\&hover\:text-red:hover {
    --un-text-opacity: 1;
    color: rgba(248,113,113,var(--un-text-opacity));
}

@media (hover: hover) and (pointer: fine)
.dark .\&hover\:dark\:text-blue:hover {
    --un-text-opacity: 1;
    color: rgba(96,165,250,var(--un-text-opacity));
}
```

Tip: Do not use with normal `hover:`, because hover will overwrite `&hover:` 
```html
<!-- hover:text-blue --working, &hover:text-red --no woking -->
<div class="&hover:text-red hover:text-blue">Hello World</div>
```

PR #1710, opened 2022-10-09
feat(preset-wind): sticky hover

chu121su12:
Thanks to #1681 this PR introduces sticky hover as a standalone variant. The sigil was `@` instead of `&` as in the linked issue.
However, due to its standalone nature, there's no integration with css functions available in pseudo so chaining this variant may require variable variant.

### DIFF
--- a/packages/preset-wind/src/variants/experimental.ts
+++ b/packages/preset-wind/src/variants/experimental.ts
@@ -1,0 +1,9 @@
+import type { Variant } from '@unocss/core'
+import { variantMatcher } from '@unocss/preset-mini/utils'
+
+export const variantStickyHover: Variant[] = [
+  variantMatcher('@hover', input => ({
+    parent: `${input.parent ? `${input.parent} $$ ` : ''}@media (hover: hover) and (pointer: fine)`,
+    selector: `${input.selector || ''}:hover`,
+  })),
+]
--- a/packages/preset-wind/src/variants/experimental.ts
+++ b/packages/preset-wind/src/variants/experimental.ts
@@ -1,9 +1,13 @@
 import type { Variant } from '@unocss/core'
+import { warnOnce } from '@unocss/core'
 import { variantMatcher } from '@unocss/preset-mini/utils'
 
 export const variantStickyHover: Variant[] = [
-  variantMatcher('@hover', input => ({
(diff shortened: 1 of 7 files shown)


packages/preset-wind/src/variants/experimental.ts
```diff
@@ -0,0 +1,9 @@
+import type { Variant } from '@unocss/core'
+import { variantMatcher } from '@unocss/preset-mini/utils'
+
+export const variantStickyHover: Variant[] = [
```
antfu, line 4: Can we add a `## Experimental` field in `preset-wind/README.md` describe it a bit? Thanks

chu121su12: Do you think it is necessary to add warn on usage?

antfu: Maybe. We could have a utility like `warnOnce` or something bound to the generator.

chris-zhu (approved):

packages/preset-wind/src/variants/experimental.ts
```diff
@@ -0,0 +1,13 @@
+import type { Variant } from '@unocss/core'
+import { warnOnce } from '@unocss/core'
+import { variantMatcher } from '@unocss/preset-mini/utils'
+
+export const variantStickyHover: Variant[] = [
+  variantMatcher('@hover', (input) => {
+    warnOnce('The @hover variant is experimental and may be changed in breaking ways at any time.')
```
antfu, line 7:
```suggestion
    warnOnce('The @hover variant is experimental and may not follow semver.')
```

antfu (approved):

antfu (approved):

zyyv:
@chu121su12 Hi, can you help `@hover` compatibility problem  with pseudo, like `group-@hover` `peer-@hover`, or another logic to resovle it. I'm sorry I can't fix this 😢.

chu121su12:
@zyyv Those pseudos need modifier, so you'd want to add something like:

```peer-invalid-@hover```

I'm uncertain if originally the pseudos can work without modifier.

https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state
https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-sibling-state

Issue #1702, opened 2022-10-08
shims.d.ts slow VSCode down.


withyellow:
```ts
// shims.d.ts

import type { AttributifyAttributes } from '@unocss/preset-attributify';

declare module 'react' {
  interface HTMLAttributes<T> extends AttributifyAttributes {}
}

```

Code above will slow some part of typescript language service of VSCode, eg. error report. IntelliSense works fine.

Steps to repoduce please checkout: https://github.com/Xwil/unocss-slow-down-vs-code.

It happens in `Vite+React+TS` and `Next.js+TS` projects.

antfu:
What would you expect for the fix? I am not sure if there is anything we could do.

Xwil:
> What would you expect for the fix? I am not sure if there is anything we could do.

Well, I am not sure about this too, but I think a hint may be helpful, it may take a long time to find out those codes may cause the editor to slow down.

Flower-F:
Yes, it will slow down the TypeScript inference. But we must to do that to avoid the limitation of React, or we can't use the Attributify well.

ffxixslh:
Maybe you can try this to avoid the slow:

```ts
// src/shims.d.ts
import * as React from 'react'
declare module 'react' {
  interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T> {
    flex?: boolean
    relative?: boolean
    text?: string
    grid?: boolean
    before?: string
    after?: string
    shadow?: boolean
    absolute?: boolean
  }
}
```

But it needs you to add the attributes manually.😂

Tzxhy:
for this problem, i decide not to use Attributify.

PR #1712, opened 2022-10-09
doc: add warning to AttributifyAttributes ts usage.

Xwil:
https://github.com/unocss/unocss/issues/1702

### DIFF
--- a/packages/preset-attributify/README.md
+++ b/packages/preset-attributify/README.md
@@ -119,6 +119,8 @@ presetAttributify({
 
 ## TypeScript Support (JSX/TSX)
 
+⚠️ Use with caution, those codes may cause your editor's TS language service to slow down.
+
 Create `shims.d.ts` with the following content:
 
 > By default, the type includes common attributes from `@unocss/preset-uno`. If you need custom attributes, refer to the [type source](https://github.com/antfu/unocss/blob/main/packages/preset-attributify/src/jsx.ts) to implement your own type.


antfu:
I would like to know the more specific number of the expected slow down, like 2.4x slower or just 10% etc.?

Xwil:
It's like 0.1s to 3~5s, both of VSCode and WebStorm.

Flower-F:
> I would like to know the more specific number of the expected slow down, like 2.4x slower or just 10% etc.?

I test it in my Windows laptop, and the TS inference time will be 2~3s slower.

Flower-F:
> I would like to know the more specific number of the expected slow down, like 2.4x slower or just 10% etc.?

Can we give an alternate to the user, such as

```tsx
import type { AttributifyAttributes } from '@unocss/preset-attributify';

declare module "react" {
  interface HTMLAttributes<T> extends Partial<Record<string, any>> { }
}
```

I fount it doesn't  slow down the TypeScript inference, but of course it might cause some other errors when using an non-existent property.

Xwil:
> > I would like to know the more specific number of the expected slow down, like 2.4x slower or just 10% etc.?
> 
> Can we give an alternate to the user, such as
> 
> ```tsx
> import type { AttributifyAttributes } from '@unocss/preset-attributify';
> 
> declare module "react" {
>   interface HTMLAttributes<T> extends Partial<Record<string, any>> { }
> }
> ```
> 
> I fount it doesn't slow down the TypeScript inference, but of course it might cause some other errors when using an non-existent property.

What about [Typescript Language Service Plugin](https://github.com/Microsoft/TypeScript/wiki/Writing-a-Language-Service-Plugin) ?

PR #1714, opened 2022-10-09
fix(cli): load config twice

### LINKED ISSUE #1689: `@unocss/cli` warn about duplication of preset
image

To reproduce, install `@unocss/cli`, create `uno.config.js` and add any preset then run the CLI. Warning message is always appeared in screen.

After some digging, I think `loader.load()` doesn't work correctly. It's duplicate entries of `presets` array instead of "merge" or something and that cause the problem.

https://github.com/unocss/unocss/blob/76b883f3d412b61be85f35b08ebc5f616abef818/packages/config/src/index.ts#L57

avxit:
This should fix #1689 .

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -31,8 +31,13 @@ export async function build(_options: CliOptions) {
   const cwd = _options.cwd || process.cwd()
   const options = await resolveOptions(_options)
 
-  const ctx = createContext<UserConfig>(options.config, defaultConfig)
-  const configSources = (await ctx.updateRoot(cwd)).sources
+  async function loadConfig() {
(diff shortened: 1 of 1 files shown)


sudongyuer:
test failed

chris-zhu (approved):

avxit:
> test failed

Yeah..., wrong `cwd` and the test failed. I don't know whether should I modified `createContext` to accept `cwd` or not?

chris-zhu:
> 

Yes, i agree with you. But it looks a bit ...

avxit:
> Yes, i agree with you. But it looks a bit ...

I've updated the PR with `ctx.updateRoot` to update correct `cwd`.

antfu (approved):

PR #1716, opened 2022-10-09
feat(preset-mini): support content-raw-[xxx]

sudongyuer:


packages/preset-mini/src/_rules/static.ts
```diff
@@ -70,6 +70,7 @@ export const contentVisibility: Rule[] = [
 
 export const contents: Rule[] = [
   [/^content-\[(.+)\]$/, ([, v]) => ({ content: `"${v}"` })],
+  [/^content-raw-\[(.+)\]$/, ([, v]) => ({ content: `"${v}"` })],
```
xhebox, line 73: Should be "content: `${v}`"

antfu, line 73:
```suggestion
  [/^content-raw-\[(.+)\]$/, ([, v]) => ({ content: v })],
```

Isn't this supported to be like this?

chu121su12:
Can I suggest alternate syntax of `content-[raw:x]` or `content-[string:x]` instead?

chris-zhu:
> Can I suggest alternate syntax of `content-[raw:x]` or `content-[string:x]` instead?
> 
> Edit: using `h.bracket` and something [like this](https://github.com/unocss/unocss/commit/43f0b48ec011df6b36f192c2412fff337ae65bc7)

Cool, I prefer this syntax.

PR #1725, opened 2022-10-11
feat(svelte): Add svelte-preprocess-unocss preprocessor for using UnoCSS in `svelte-package` [paused]

### LINKED ISSUE #1719: feat: Add a svelte-preprocess-unocss preprocessor to support usage in component libraries built using svelte-package
As mentioned in #1676, I would like to create a simple `@unocss/svelte-preprocess-unocss` preprocessor for use in `svelte.config.js` files for situations where someone is creating a component library using `svelte-package` which doesn't run through Vite at all. 

This will be a very simple wrapper utilizing `createGenerator`, the updated `transformSvelteSFC` function from #1692, and a way to run any added transformers.

Progress is being made in #1725

jacob-8:
Closes #1719

**Edit**: Since it appears the hope is to add support in Svelte for `svelte-package` to take advantage of Vite plugins this PR is parked for the moment simply to illustrate how a svelte-preprocessor could be created. I will await the results of https://github.com/sveltejs/vite-plugin-svelte/issues/475

~~I need help with:~~
~~- [ ] Please give advice on where to place documentation for this package. It's not a Vite plugin so it doesn't belong with the other `SvelteKit` docs. Perhaps a separate Svelte page, that's referenced from the main page and the Vite page?~~
~~- [ ] Add new package to NPM.~~


userquin:
@jacob-8 for inline deps check the externals entry in preset icons: https://github.com/unocss/unocss/blob/main/packages/preset-icons/build.config.ts

jacob-8:
> @jacob-8 for inline deps check the externals entry in preset icons: https://github.com/unocss/unocss/blob/main/packages/preset-icons/build.config.ts

Thanks for pointing me in the right direction. Realized the preprocessor is only using `createGenerator` and not `createContext` so I simplified a bit.

userquin:
@jacob-8 can you take a look these?:
- https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/config.md#usevitepreprocess 
- https://github.com/sveltejs/vite-plugin-svelte/tree/vite-preprocess-css-api (experimental)

I was talking with `@dominikg` and it seems we should avoid using another scoping solutions in svelte

dominikg:
drafted https://github.com/sveltejs/vite-plugin-svelte/issues/475  which may enable better support for vite based preprocessors in svelte-package or other svelte library tooling. Please add feedback there if possible.

jacob-8:
> I was talking with `@dominikg` and it seems we should avoid using another scoping solutions in svelte

Thanks! That would be much appreciated. 

> drafted [sveltejs/vite-plugin-svelte#475](https://github.com/sveltejs/vite-plugin-svelte/issues/475) which may enable better support for vite based preprocessors in svelte-package or other svelte library tooling. Please add feedback there if possible.

I've already published this preprocessor to a temporary package and am using it for my projects so I'm not in a rush for a fix here. I'll follow along and look forward to a good solution.

jacob-8:
For those who still are looking for such, you can find the [`svelte-preprocess-unocss`](https://www.npmjs.com/package/svelte-preprocess-unocss) package on npm and read about it at https://github.com/jacob-8/svelte-scoped-uno/blob/main/packages/svelte-preprocess-unocss/README.md

jacob-8:
> We are unsure if this limitation is specific to the `svelte-scoped` mode or if it is a bug.

It is a limitation. This preprocessor is heading towards being released as part of the UnoCSS family of packages (watch https://github.com/unocss/unocss/pull/2552). Even though that package is not available yet, there are a few parts of the docs that are applicable to your question. Please read [how it works](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#how-it-works) and then the proper [usage](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#usage).

For your use case, I'm going to assume building a component library since you're using the preprocessor and not the current UnoCSS vite plugin. I think you'd be better off using class directives (`class:bg-green-200={color === 'green'}`), but I don't know your situation, so if you must have classes in your script block I would recommend you do something like this:

```svelte
<script lang="ts">
  export let color: "blue" | "green";
  $: computedClass = 'background-' + color;
</script>

<div class="{computedClass}" />

<style>
  :global(.background-blue) {
    --at-apply: bg-blue-200;
  }
  :global(.background-green) {
    --at-apply: bg-green-200;
  }
</style>
```

jacob-8:
> ```svelte
> <script>
>   $: variable = condition ? 'flex' : 'block'; 
> </script>
> <div class={variable}></div>
> ```

Then write it like this:

```svelte
<div 
  class:flex={condition} 
  class:block={!condition} />
```

You can also do an inline expression but the above example is the recommended way.

```svelte
<div class="{condition ? 'flex' : 'block'}" /> 
```

Reference: https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#usage

jacob-8:
> Unfortunately, it seems that using classes inside the script block won't work for our case then.

Yeah probably not, though it's hard to give advice without having a better understanding of your use case. If you're in an app, you can use the safelist, but if you're developing one-off components that you want to be able to be used by themselves as part of a component library you're best off just using `--at-apply`.

PR #1743, opened 2022-10-14
docs: improve sentence

hoishing:
typo in the param name

### DIFF
--- a/packages/preset-icons/README.md
+++ b/packages/preset-icons/README.md
@@ -80,7 +80,7 @@ By default, this preset will choose the rendering modes automatically for each i
 - `?bg` for `background-img` - renders the icon as a background image
 - `?mask` for `mask` - renders the icon as a mask image
 
-For example, `vscode-icons:file-type-light-db`, an icon with colors that will be rendered as a mask image and bypass it's colors. Use `vscode-icons:file-type-light-db?bg` to render it as a background image.
+For example, `vscode-icons:file-type-light-pnpm`, an icon with colors (the `svg` doesn't contain `currentColor`) that will be rendered as a background image. Use `vscode-icons:file-type-light-pnpm?mask` to render it as a mask image and bypass it's colors.
 
 ## Configuring collections and icons resolvers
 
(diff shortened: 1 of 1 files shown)


packages/preset-icons/README.md
```diff
@@ -80,7 +80,7 @@ By default, this preset will choose the rendering modes automatically for each i
 - `?bg` for `background-img` - renders the icon as a background image
 - `?mask` for `mask` - renders the icon as a mask image
 
-For example, `vscode-icons:file-type-light-db`, an icon with colors that will be rendered as a background image. Use `vscode-icons:file-type-light-db?bg` to render it as a mask image and bypass it's colors.
+For example, `vscode-icons:file-type-light-db`, an icon with colors that will be rendered as a background image. Use `vscode-icons:file-type-light-db?mask` to render it as a mask image and bypass it's colors.
```
chris-zhu, line 83:
```suggestion
For example, `vscode-icons:file-type-light-db`, an icon with colors that will be rendered as a mask image and bypass it's colors. Use `vscode-icons:file-type-light-db?bg` to render it as a background image.
```

We are use mask image by default.

chris-zhu (changes requested):

antfu (approved):

userquin:
uhmmm, the original sentence is correct: when `currentColor` is detected, the icon will be always rendered as masked image, otherwise will be rendered as a bg icon, check the logic here: https://github.com/unocss/unocss/blob/main/packages/preset-icons/src/core.ts#L93

The problem is that ` vscode-icons:file-type-light-db` has no colors, but it do not contain `currentColor`, so it will be always rendered as a bg icon and not as a masked icon.

userquin:
maybe we can choose another icon and also update the sentence to include "doesn't contain `currentColor`":
 ```md
For example, `vscode-icons:file-type-light-db`, an icon with colors (the `svg` doesn't contain  `currentColor`) that will be rendered as a background image. Use `vscode-icons:file-type-light-db?bg` to render it as a mask image and bypass it's colors.
```

**Comment by userquin** [comment]
anyway, both sentences are wrong, the sentence should be:
```md
For example, `vscode-icons:file-type-light-db`, an icon with colors (the `svg` doesn't contain `currentColor`) that will be rendered as a background image. Use `vscode-icons:file-type-light-db?mask` to render it as a mask image and bypass it's colors.
```

Issue #1750, opened 2022-10-15
VSCode extension : support electron-vite


michael-dm:
electron-vite use a special `electron.vite.config.ts` file, it is not currently detected by unoCSS extension.

https://evite.netlify.app/guide/project-structure.html

azaleta:
you can use a separate `unocss.config.ts/uno.config.ts`

antfu:
https://github.com/unocss/unocss/pull/1751#issuecomment-1279907398

PR #1751, opened 2022-10-16
chore(vscode): support electron.vite.config

sudongyuer:
maybe solved this (not sure) #1750


antfu:
Thanks but I think this is too specific. In general, we would recommend using a separate `unocss.config.js` to have better IDE integration.

Issue #1748, opened 2022-10-14
Support variant group like feature in children and other selector related rules


jd-solanki:
Hi 👋🏻 ,

Recently I tried https://uno.antfu.me/?s=children-[.a-drawer]-(bg-red_text-white) to add two styles/rules using single selector/rule.

This is slightly similar to the [variant group](https://github.com/unocss/unocss/tree/main/packages/transformer-variant-group) feature.

Grouping support for rules like the below (where the actual rule is added in the class itself) will be better for DX
- `children-[.a-drawer]-(bg-red_text-white)` 
- `children-[.a-drawer]-bg-red children-[.a-drawer]-text-white`

sudongyuer:
Hi, my friend. You can use this below
<img width="1216" alt="image" src="https://user-images.githubusercontent.com/76603360/196018972-4bb55bea-91e4-4d94-943d-7350e6eba791.png">

sudongyuer:
#1753

chris-zhu:
I don't recommend that you mix things like this, `_` is more of a placeholder in rules with bracket.
e.g.
`children-[.a-drawer]-(bg-red_text-white_p-[2px_4px])` That doesn't look good.

PR #1753, opened 2022-10-16
chore(transform-variant-group): add test case

sudongyuer:


sudongyuer:
#1748

chris-zhu:
Seems like no difference, the above case can already be included.

https://github.com/unocss/unocss/blob/e5e02757e13ba3f503008b1880bc0c49aef65b7d/test/variant-group.test.ts#L32-L34

sudongyuer:
this is a usecase test

PR #1754, opened 2022-10-16
feat(transformer-variant-group): support underscore to expand

sudongyuer:
#1748

### DIFF
--- a/packages/core/src/utils/variantGroup.ts
+++ b/packages/core/src/utils/variantGroup.ts
@@ -16,7 +16,7 @@ export function expandVariantGroup(str: string | MagicString, seperators = ['-',
         if (!seperators.includes(sep))
           return from
         return body
-          .split(/\s/g)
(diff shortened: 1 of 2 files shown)


chu121su12:
There's bracket utility with underscores that's substituted as space. Could you add tests for those scenarios?

https://github.com/unocss/unocss/blob/e5e02757e13ba3f503008b1880bc0c49aef65b7d/test/assets/preset-mini-targets.ts#L296

chris-zhu:
> There's bracket utility with underscores that's substituted as space. Could you add tests for those scenarios?
> 
> https://github.com/unocss/unocss/blob/e5e02757e13ba3f503008b1880bc0c49aef65b7d/test/assets/preset-mini-targets.ts#L296

It's not pass the test, i think this syntax conflicts with bracket, I don't like this kind of analysis.🤔️

PR #1756, opened 2022-10-16
feat(preset-web-fonts): try using local fonts

chris-zhu:
As the mdn [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/src) said, if we use webFonts inline import, try using `local` first.

### DIFF
--- a/packages/preset-web-fonts/src/index.ts
+++ b/packages/preset-web-fonts/src/index.ts
@@ -23,10 +23,10 @@ export function normalizedFontMeta(meta: WebFontMeta | string, defaultProvider: 
 
 function convertToLocalFontFace(css: string) {
   return css.replace(/@font-face\s*{[^}]*}/g, (match) => {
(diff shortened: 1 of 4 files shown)


sudongyuer (changes requested):
supports these font-family names
  `font-family: myFirstFont`
  `font-family: "myFirstFont"`
  `font-family: 'myFirstFont'`

sudongyuer (approved):

antfu:
I am not so sure, it's this the responsibility of Google Fonts? If they didn't do this probably there are some reasons.

chris-zhu:
> I am not so sure, it's this the responsibility of Google Fonts? If they didn't do this probably there are some reasons.

I'll keep watching and continue to investigate why.

chris-zhu:
I will close this PR, details in https://github.com/google/fonts/issues/5418#issuecomment-1281724382.

Issue #1760, opened 2022-10-16
Question about the generation of default css variable names


itmanyong:
For the default automatically generated css variables mounted to the style name, there will be incompatibility in specific platforms can not recognize the error reported, for example: the applet platform css does not support * this situation. It would be great if you could configure whether to import this part or provide configuration items for this part of the developer level custom configuration! There are several proposals on this issue.
![image](https://user-images.githubusercontent.com/53500371/196042817-81db22f1-16f5-4445-9917-d2575b8142d0.png)

1. css variable mounted style name to provide an interface to define, rather than directly default, pseudo-examples are as follows.
```js
//unocss -config
{
  ...
  variableClass:".root .header"  //default =  *
}
//Generate Style
.header::before,
.header::after,
.root::before,
.root::after{...}
// Or a function that uses the returned content as the style, would that be a little more free and controllable for developers?
```
2.Provide switch control whether to generate? It does not feel good maybe

itmanyong:
This can be done with a platform plug-in

PR #1762, opened 2022-10-17
feat(presets): customize preflight's root

chu121su12:
To add fature asked in #1760 but already closed by the time this PR's code is done.


packages/preset-mini/src/preflights.ts
```diff
@@ -8,7 +8,8 @@ export const preflights: Preflight[] = [
     getCSS(ctx: PreflightContext<Theme>) {
       if (ctx.theme.preflightBase) {
         const css = entriesToCss(Object.entries(ctx.theme.preflightBase))
-        return `*,::before,::after{${css}}::backdrop{${css}}`
+        const root = ctx.theme.preflightRoot ?? ''
```
antfu, line 11: I would prefer to make the entire selector customizable: 

```suggestion
        const root = ctx.theme.preflightRoot ?? ':root,:root::before,:root::after'
```

chu121su12: Updated to arrayable since there's separated `::backdrop`

packages/preset-mini/src/preflights.ts
```diff
@@ -8,7 +8,11 @@ export const preflights: Preflight[] = [
     getCSS(ctx: PreflightContext<Theme>) {
       if (ctx.theme.preflightBase) {
         const css = entriesToCss(Object.entries(ctx.theme.preflightBase))
-        return `*,::before,::after{${css}}::backdrop{${css}}`
+        let roots = toArray(ctx.theme.preflightRoot ?? [])
+        if (!roots.length)
+          roots = ['*,::before,::after', '::backdrop']
```
antfu, line 13:
```suggestion
        let roots = toArray(ctx.theme.preflightRoot ?? ['*,::before,::after', '::backdrop'])
```

Can we do this? I think it's also reasonable to pass an empty array to disable this preflight

Issue #1767, opened 2022-10-17
VS Code extension adds "color circle" next to `font-black`, which describes a weight, not a color


AaronBeaudoin:
It's all in the title. The UnoCSS VS code extension adds a "color circle" UI decoration next to `font-black`, but that class is describing a font weight, not a color, so the circle shouldn't be there.

![image](https://user-images.githubusercontent.com/26592486/196288765-35e1249e-8158-4468-931b-9abd873ee9a6.png)

PR #1769, opened 2022-10-18
fix(vscode): skip util contains none color define

azaleta:
Linked issue: #1767

### DIFF
--- a/packages/vscode/src/utils.ts
+++ b/packages/vscode/src/utils.ts
@@ -64,7 +64,7 @@ export async function getColorsMap(uno: UnoGenerator, result: GenerateResult) {
   const colorsMap = new Map<string, string>()
 
   for (const i of result.matched) {
-    if (await isColorUtility(uno, i))
+    if (!(await isColorUtility(uno, i)))
       continue
 
     const matchedValueless = i.match(matchedValuelessAttributifyRE)?.[0]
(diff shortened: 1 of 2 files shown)


packages/vscode/src/utils.ts
```diff
@@ -97,3 +100,9 @@ export function isSubdir(parent: string, child: string) {
   const relative = path.relative(parent, child)
   return relative && !relative.startsWith('..') && !path.isAbsolute(relative)
 }
+
+async function isColorUtility(uno: UnoGenerator, utilName: string) {
+  const css = (await getPrettiedCSS(uno, utilName)).css
```
antfu, line 105: Do we need to prettify this? We could probably use `uno.genreate` directly.

azaleta: I think result.css is a long string contains all css generated.
https://github.com/unocss/unocss/blob/0ba0d5a10e9ccd6c1da98dbea99cf9ddd94458b6/packages/vscode/src/annotation.ts#L98

Considering different pattern of css class name, It's hard to drag css of one utility. 

css define could be:
```
1. [font~="black"]{font-weight:900;}  (<div font="black" text-red >)

2. .font-black,[font-black=""]{font-weight:900;}   (<div font-black text-red >)
```

We could use `font-black` and Regex  to get utility css from `result.css` in pattern 2
But need more operation for pattern 1. 
So I am using prettify function just for get the utility css only.
Not for the prettify :)

antfu: I see. We could directly call `uno.generate` again with that utility, it's been cached by UnoCSS and it should be fast. We don't need prettify on this.

PR #1783, opened 2022-10-20
chore: update netlify configuration

zyyv:
pnpm supported, refer to (https://github.com/netlify/build-image/pull/845).

Bug of monorepo(https://github.com/netlify/build-image/issues/855)


Dunqing (commented):
Now that `netlify.toml` has been moved to the `root workspace`, you have to set the `publish` to `interactive/dist`, and I don't think you need to set `NETLIFY_USE_PNPM` because the `root workspace` has `pnpm-lock.yaml`, that finds `pnpm`.

chris-zhu:
> Now that `netlify.toml` has been moved to the `root workspace`, you have to set the `publish` to `interactive/dist`.

Yes, i agree with you. But very strange,
- if it is `dist`, publish is the directory is `interactive/docs`.
- if it is `interactive/dist`, publish is the directory is `interactive/interactive/docs`.

> I don't think you need to set NETLIFY_USE_PNPM because the root workspace has pnpm-lock.yaml, that finds pnpm.

Even though now the netlify config is in the root directory, it can't find `pnpm`, it still uses `npm`

Dunqing:
> Yes, i agree with you. But very strange,
> 
> * if it is `dist`, publish is the directory is `interactive/docs`.
> * if it is `interactive/dist`, publish is the directory is `interactive/interactive/docs`.

image

I guess the `base directory` is already configured in the netlify ui.
 

> Even though now the netlify config is in the root directory, it can't find `pnpm`, it still uses `npm`, i have to use `NETLIFY_USE_PNPM` env variant to use `pnpm`😢

Same as above, cannot find pnpm in  `repo/interactive`, keep `NETLIFY_USE_PNPM`😕

Issue #1796, opened 2022-10-25
[unocss] unmatched utility "2xl:2xl:2xl:xl:lg:__container"


sondh0127:
`container` no longer work if prefix was defined customize `h-container`
![image](https://user-images.githubusercontent.com/62163604/197811221-15859396-434c-404a-bab2-e565044d1c40.png)

```
presets: [
  presetUno({
    prefix: 'h-',
    variablePrefix: 'h-'
  }),
]
```

antfu:
Would be great if you can send a PR with the failing tests.

PR #1801, opened 2022-10-26
test: add failing test for `container` with prefix

sondh0127:
add failing test for `container` with prefix - related to #1796


antfu:
Fixed by #1807, thank you!

PR #1803, opened 2022-10-26
feat(preset-wind): experiment rule of `zoom`

### LINKED ISSUE #1799: (feature request): add utility class for zoom css property
Hello, y'all.

So I noticed that the css property `zoom` does not have a utility class in unocss. I find this useful and humbly ask if it could be added.
And as an example , It would be useful with the `print` variant.

I came across a situation were i needed to change the zoom level of a page during printing using css.

See implementation using css

```css
@media print {
    .result {
        zoom: 58%;
    }
}
```

See proposed unocss equivalent
```html
<body class="print:zoom-58%"></body>
```

😁😁😁

**EDITED**

So I made some more research and found out that the use of the `zoom` property is kind of discouraged for some reasons. So i can understand why it might have not been added by default to unocss.

Though I still find it useful in some use case 🤷‍♂️

See reference link https://developer.mozilla.org/en-US/docs/Web/CSS/zoom

chris-zhu:
close #1799


antfu:
Emm, any reason to use a non-standard feature over the standard `scale`?

therealokoro:
> Emm, any reason to use a non-standard feature over the standard `scale`?

@antfu yes sir. In my use case, `scale` property didn't give the expected result but the `zoom` property did. And this was same issue for some persons over at stackoverflow.

antfu:
Since this is a non-standard and discouraged feature, there will never be a time point for this "experimental" state to graduate - I prefer not to have it built-in.

@therealokoro If you really need it and are willing to take the risk, I think that's the strength of UnoCSS you can define the rule directly within your config with three lines of code.

PR #1807, opened 2022-10-27
fix(core): support shortcut prefix

azaleta:
Linked issue: #1796 

The orignal test file is using `presetMini` which does not include the `container` shortcut
-> change to `presetUno`

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -394,7 +394,7 @@ export class UnoGenerator {
     return cssBody
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/generator/index.ts
```diff
@@ -555,7 +560,7 @@ export class UnoGenerator {
         .map(async (i) => {
           const result = isString(i)
             // rule
-            ? await this.parseUtil(i, context, true) as ParsedUtil[]
+            ? await this.parseUtil(i, context, true, meta.prefix ? meta.prefix : undefined) as ParsedUtil[]
```
chris-zhu, line 563:
```suggestion
            ? await this.parseUtil(i, context, true, meta.prefix) as ParsedUtil[]
```

chris-zhu (approved):

PR #1811, opened 2022-10-28
fix(vscode): attributify value starts with number

zhmushan:
The string to be tested here is the **source code** and does not need to be "escaped".

### DIFF
--- a/packages/shared-common/src/index.ts
+++ b/packages/shared-common/src/index.ts
@@ -1,5 +1,5 @@
 import type { UnoGenerator } from '@unocss/core'
-import { e, isAttributifySelector, regexClassGroup } from '@unocss/core'
+import { isAttributifySelector, regexClassGroup } from '@unocss/core'
 import MagicString from 'magic-string'
 
 // https://github.com/dsblv/string-replace-async/blob/main/index.js
@@ -78,7 +78,7 @@ export function getMatchedPositions(code: string, matched: string[], hasVariantG
 
   // attributify values
   attributify.forEach(([, name, value]) => {
-    const regex = new RegExp(`(${e(name)}=)(['"])[^\\2]*?${e(value)}[^\\2]*?\\2`, 'g')
+    const regex = new RegExp(`(${name}=)(['"])[^\\2]*?${value}[^\\2]*?\\2`, 'g')
     Array.from(code.matchAll(regex))
       .forEach((match) => {
         const escaped = match[1]
--- a/packages/shared-common/src/index.ts
+++ b/packages/shared-common/src/index.ts
@@ -1,5 +1,5 @@
 import type { UnoGenerator } from '@unocss/core'
-import { isAttributifySelector, regexClassGroup } from '@unocss/core'
+import { escapeRegExp, isAttributifySelector, regexClassGroup } from '@unocss/core'
 import MagicString from 'magic-string'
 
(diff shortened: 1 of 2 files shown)


test/pos.test.ts
```diff
@@ -15,7 +15,7 @@ describe('matched-positions', async () => {
       ],
     })
 
-    expect(await match(uno, '<div border="b gray4"></div>'))
+    expect(await match(uno, '<div border="b gray4 2"></div>'))
```
zhmushan, line 18: escape("2") => "\\32 "

test/pos.test.ts
```diff
@@ -28,6 +28,11 @@ describe('matched-positions', async () => {
             20,
             "[border=\\"gray4\\"]",
           ],
+          [
+            21,
+            22,
+            "[border=\\"2\\"]",
```
sudongyuer, line 34: Should be escape  like this `"[border=\"\\32 \"]"`

zhmushan: Here it must be "2" and not "\32", but I forget why, wait for me to sleep first, I will give you an answer tomorrow.

sudongyuer: ok,  i have already push the  changes , you can undo it , if you consider it's wrong

sudongyuer: There are some context about this issue
https://drafts.csswg.org/cssom/#escape-a-character-as-code-point

packages/shared-common/src/index.ts
```diff
@@ -78,7 +78,7 @@ export function getMatchedPositions(code: string, matched: string[], hasVariantG
 
   // attributify values
   attributify.forEach(([, name, value]) => {
-    const regex = new RegExp(`(${e(name)}=)(['"])[^\\2]*?${e(value)}[^\\2]*?\\2`, 'g')
+    const regex = new RegExp(`(${name}=)(['"])[^\\2]*?${value}[^\\2]*?\\2`, 'g')
```
chu121su12, line 81: Shouldn't `e` at least be replaced with `escapeRegEx`?

test/pos.test.ts
zhmushan: image
@sudongyuer vscode already does escape (I'm guessing), so if we do it again, nothing will show up

packages/shared-common/src/index.ts
zhmushan: I looked at this function, but I'm sorry, I didn't figure out what it means. Can you explain it in detail?

sudongyuer: > I looked at this function, but I'm sorry, I didn't figure out what it means. Can you explain it in detail?

https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape

chu121su12: I'm sorry the function name was `escapeRegExp`. (see [search](https://github.com/unocss/unocss/search?q=escapeRegExp))
https://github.com/unocss/unocss/blob/d337ce401788a898208e3ebdb94b2e6c3c0ab01b/packages/core/src/utils/escape.ts#L1

This is usually done if you are using string from variable while constructing regex, similar to escaping string variable with `e()` or the like before using it in html.

zhmushan: @chu121su12 Thanks, you are right, I have replaced `e()` with `escapeRegExp()` and added tests.

Issue #1793, opened 2022-10-24
Apply css rules based on html data attribute


mtt-artis:
I'm using Radix-UI and i'd like to apply css rules based on html data attribute.

Is there a syntax or a variant to do so ?

something like this:
```tsx
const A = ({state}: {state: "active" | "error"}) => (
  <button
    data-state={state}
    className='[data-state="active"]:bg-blue [data-state="error"]:bg-red'
  >
    my button
  </button>
)
```

therealokoro:
Hmmmmmm 🤔, this could be very useful you know. I guess this should fall under custom selectors

mtt-artis:
so i made a custom variant in my unocss config

```js
import react from '@vitejs/plugin-react';
import Unocss from 'unocss/vite';

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [
    react(), 
    Unocss({
      variants: [
        (matcher) => {
          const [, attribute = '', value = ''] = matcher.match(/^\[([^\/]+)\/?(.*)\]:/) ?? []
          if (!attribute) {
            return { matcher }
          }
          const attributeValue = value ? `="${value}"` : value;
          return {
            matcher: matcher.replace(/^\[.+\]:(.*)$/, '$1'),
            selector: (s) => `[${attribute}${attributeValue}]${s}`,
          }
        }
      ]
    }),
  ]
})
```

and use it like this

```jsx
const A = ({state}: {state: "active" | "error"}) => (
  <button
    data-state={state}
    className='[data-state/active]:bg-blue [data-state/error]:bg-red'
  >
    my button
  </button>
)
```

i had some issues if i kept the `=` so i replaced it with a `/`,  i wish someone could make a `=` version 
and I removed the quotes as they might interact with the class string

PS: is there a discord community ?

chris-zhu:
```jsx
const A = ({state}: {state: "active" | "error"}) => (
  const className = state === 'active' ? 'bg-blue' : 'bg-red'
  <button className='{className}'>
    my button
  </button>
)
```
Excuse me for not knowing Radix-UI, but wouldn't it be simpler?

> ps: here is UnoCSS discord community. [https://discord.gg/xagTXneNHd](https://discord.gg/xagTXneNHd)

chu121su12:
[...]

PR #1816, opened 2022-10-29
feat(preset-mini,preset-wind): add aria-x and data-x support

chu121su12:
Addresses #1793 with slight syntax change to align with tailwind


mzaien:
Thank you @chu121su12, I was working on it locally but you made it 🚀

PR #1821, opened 2022-10-29
feat(preset-mini): add container query syntax

chu121su12:
This PR adds container query variant using `@<size>/<label>` or `@[<bracket>]/<label>` syntax. The `size` syntax looks up `size` in `theme.containers` while `bracket` syntax use `h.bracket` to resolve the required `@` value.


antfu:
For reference: https://github.com/tailwindlabs/tailwindcss-container-queries

chu121su12:
I was reading that but then I was thinking of fully dynamic bracket rule.

PR #1824, opened 2022-10-30
chore(vite): update inject code

### LINKED ISSUE #1823: [unocss-hmr] TypeError: __vite_ssr_import_meta__.hot.send is not a function
import global uno.css in the histoire.set.up
```ts
// src/histoire.setup.ts
import 'uno.css'
```
histoire.config.ts
image

vite.config.ts
image

uno.config.ts
image

and run dev can't not find the import.meta.hot.send function
image

image

how can i fix that

sudongyuer:
fix #1823

### DIFF
--- a/packages/vite/src/modes/global/dev.ts
+++ b/packages/vite/src/modes/global/dev.ts
@@ -164,7 +164,7 @@ export function GlobalModeDevPlugin({ uno, tokens, affectedModules, onInvalidate
         // inject css modules to send callback on css load
         if (layer && code.includes('import.meta.hot')) {
           return `${code}
-if (import.meta.hot) {
+if (import.meta.hot && import.meta.hot.send) {
   try { await import.meta.hot.send('${WS_EVENT_PREFIX}', ['${layer}', __vite__css.slice(2,${2 + HASH_LENGTH})]); }
   catch (e) { console.warn('[unocss-hmr]', e) }
   if (!import.meta.url.includes('?'))


packages/vite/src/modes/global/dev.ts
```diff
@@ -164,7 +164,7 @@ export function GlobalModeDevPlugin({ uno, tokens, affectedModules, onInvalidate
         // inject css modules to send callback on css load
         if (layer && code.includes('import.meta.hot')) {
           return `${code}
-if (import.meta.hot) {
+if (import.meta.hot && import.meta.hot.send) {
```
chris-zhu, line 167: If `viteHotContext` is created successfully, then `import.meta.hot.send` must exist, which is an invalid judgment. I don't think the problem is here🤔️

sudongyuer: I've looked into it and debug , sometimes `import.meta.hot.send` were `undefined`

chris-zhu: Unbelievable 😅

sudongyuer: So was i 🤣

antfu: This API is introduced in Vite 2.9, which UnoCSS requires it. If this is `undefined` I think it's probably because they have older vite installed.

sudongyuer: > 
[#1823](https://github.com/winchesHe/Idux-interactive-histoire/blob/main/package.json#L30) provide the reproduction vite version was `^3.1.6`

antfu:
I would suggest finding the root cause instead of relying on this workaround. Thanks.

PR #1830, opened 2022-11-02
fix(core): fix for svelte extractor

### LINKED ISSUE #1828: Svelte `class:` regression in 0.46.2
```svelte
<a
  class:text-gray-800={$page.url.pathname.startsWith('/test')}
  class:bg-gray-200={$page.url.pathname.startsWith('/test')}>
```

UnoCSS is able to extract the class attributes above in v0.46.1 but fails in v0.46.2 with no app config or code changes.

chu121su12:
~Should~ close #1828

~Caveat: this would make svelte extractor not supporting `data-x` and `aria-x` variants directly.~


antfu:
Does that mean in Svelte you can't use rules with `=`? Would that be a bit confusing and inconsistent?

chu121su12:
The two variants if not using theme mostly depends on the `=` in attribute selector (ex.: `data-[a~=b]:rule`). And since the update brings back splitting on `=`, rules with `=` may never get matched directly; except when manually generated via `uno.generate()`.

Although if the syntax is per the issue `class:text-orange-400={foo}`, I think we can do left + right trim, of ex. `/class:(.*)\s*=\s*/`, as another workaround; due to the other split will happen at the character `{`. The extractor already split and do the left trim:
https://github.com/unocss/unocss/blob/73d00d61e4ea68a24adeb3b4893a93f7921662ea/packages/core/src/extractors/svelte.ts#L11

chu121su12:
Added the right trim as of [a36f54b](https://github.com/unocss/unocss/pull/1830/commits/a36f54bbbe53ff404470384853aced4976f3db43). @pheuter If you don't mind, please check the syntax usage in the test.

chu121su12:
I'm sorry I don't know svelte syntax much. If the syntax is `abc={def` without any spaces then the latest update should work the same. If there's more combination other than the `={`, it may warrant another workaround.

pheuter:
The only other separator that comes to mind is `="{}"` in addition to `={}`

chu121su12:
`="{}"` should be processed the same way as `={}`

PR #1836, opened 2022-11-02
chore: typo

Simon-He95:


antfu:
Thanks but it's intentional.

PR #1845, opened 2022-11-04
test: correct test cases

Dunqing:
I think we should keep the test cases correct.

### DIFF
--- a/test/assets/preset-mini-targets.ts
+++ b/test/assets/preset-mini-targets.ts
@@ -66,6 +66,8 @@ export const presetMiniTargets: string[] = [
   'border-x-size-2',
   'border-t-size-2',
   'border-width-3',
+  'border-width-[2em]',
+  'border-width-[calc(1em-1px)]',
   'border-size-unset',
   'border-x-width-3',
   'border-t-width-3',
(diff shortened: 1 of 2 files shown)


sudongyuer:
What does that do ？

zyyv (approved):

test/assets/preset-mini-targets.ts
```diff
@@ -114,8 +116,6 @@ export const presetMiniTargets: string[] = [
 
   // border - color
   'border-[#124]',
-  'border-[2em]',
-  'border-[calc(1em-1px)]',
```
chu121su12, line 118: These two are negative cases in which brackets get resolved to color by default

zyyv: I think it's unnecessary.

Dunqing: I know, but when I first saw it I thought it was correct until I saw test.snap, I didn't know it could only change colors, which could cause some misunderstanding.

chu121su12: Maybe we can refactor to separate all negative cases? The `text-[x]` rule also has similar tests

Dunqing: Yes, that's fine. But we should remove it before refactoring.

PR #1846, opened 2022-11-04
fix(unocss): missing export default

Dunqing:
Missing export default, but its type contains export default, refer https://github.com/unocss/unocss/pull/1484


sudongyuer:
> Missing export default, but its type contains export default, refer #1484

Is this necessary ?

PR #1851, opened 2022-11-07
feat: improving some types

Dunqing:
### Common improvements
1. Make `createGenerator` support generic. (will use for test

### Theme improvements

> The code is supported, but the type is not

1. The `colors`  support `nested`.
2. The `fontSize` support `string | [string, string]`.

### DIFF
--- a/packages/preset-mini/src/_theme/types.ts
+++ b/packages/preset-mini/src/_theme/types.ts
@@ -8,7 +8,7 @@ export interface ThemeAnimation {
   counts?: Record<string, string | number>
 }
 
-interface Colors {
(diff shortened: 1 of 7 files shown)


packages/preset-mini/src/_theme/types.ts
```diff
@@ -8,6 +8,10 @@ export interface ThemeAnimation {
   counts?: Record<string, string | number>
 }
 
+interface Colors {
```
antfu, line 11:
```suggestion
export interface NestedObject {
```

antfu (approved):

PR #1857, opened 2022-11-08
chore(vite): upgrade vite & improve hmr

sudongyuer:


antfu:
This API is implemented only a few weeks ago. I don't feel the benefit of writing a few lines is worth breaking previous vite versions. I prefer to hold this until Vite 4 or later. Thanks.

PR #1872, opened 2022-11-12
fix(vite): avoid applying css transform on bundle generation

### LINKED ISSUE #1850: `__VITE_PUBLIC_ASSET__hash__` doesn't get replaced in the generated output
> I've created a minimal reproducible example [here](https://github.com/equt/reference-public-in-css).

For the class `bg-[url(/not-an-image.jpg)]`, it works correctly in the development mode, while outputs

```css
.bg-\[url\(\/not-an-image\.jpg\)\] {
  --un-url: url(__VITE_PUBLIC_ASSET__032ad66c__);
  background-image: var(--un-url);
}
```

in the build mode.

Seems like the generated CSS didn't go through Vite's post-process but instead retrieved directly via the chunk hash? Just note it here for further debugging.

equt:
close #1850

### DIFF
--- a/packages/vite/src/modes/global/build.ts
+++ b/packages/vite/src/modes/global/build.ts
@@ -175,9 +175,9 @@ export function GlobalModeBuildPlugin({ uno, ready, extract, tokens, filter, get
               .replace(HASH_PLACEHOLDER_RE, '')
             chunk.source = await replaceAsync(css, LAYER_PLACEHOLDER_RE, async (_, __, layer) => {
               replaced = true
-              return await applyCssTransform(layer === LAYER_MARK_ALL
+              return layer === LAYER_MARK_ALL
                 ? result.getLayers(undefined, Array.from(vfsLayers))
-                : result.getLayer(layer) || '', `${chunk.fileName}.css`, options.dir)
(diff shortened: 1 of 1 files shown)


equt:
Currently, the global mode for the Vite integration will not replace the generated `__VITE_PUBLIC_ASSET__hash__` asset URL as I've mentioned in #1850.

And as it turns out, inside the post transformer

https://github.com/unocss/unocss/blob/d1b72c3094d83050ef7e9dbb79025272fb5d3b00/packages/vite/src/modes/global/build.ts#L178-L180

Only the pre-plugin `vite:css` from Vite gets called

https://github.com/unocss/unocss/blob/d1b72c3094d83050ef7e9dbb79025272fb5d3b00/packages/vite/src/modes/global/build.ts#L27-L43

Since the post-one is missing, we could never turn the `__VITE_PUBLIC_ASSET__hash__` back to what it should really be.

This only happens for the global mode, and only for the build rather than the dev.

This PR simply removes the `applyCssTransform`, as I don't see any necessary it should be performed upon a generated chunk.

sudongyuer (approved):

PR #1878, opened 2022-11-15
feat(astro): introduce the `injectExtra` option

### LINKED ISSUE #1877: Add support for custom injections in `@unocss/astro`
~~The current behavior of the `@unocss/astro` is somehow confusing.~~

https://github.com/unocss/unocss/blob/d5a6a6ae9bbb55a043d6ca147a575dbc421bd62a/packages/astro/src/index.ts#L39-L41

~~the reset file was introduced by default, while no dependencies or peer dependencies have been marked~~

https://github.com/unocss/unocss/blob/d5a6a6ae9bbb55a043d6ca147a575dbc421bd62a/packages/astro/package.json#L43-L50

It did, but adding `@unocss/reset` seems to be overhead, thus I'm suggesting that we should deprecate both two injection options, and introduce a unified `inject` option. That would allow users to inject anything they want into the output.

equt:
This commit also marks the existing `injectReset` & `injectEntry` as deprecated. Users with those options set should have no breaking changes on their existing code, but only receive a warning via the linting tool.

close #1877

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -16,6 +16,12 @@ export interface AstroIntegrationConfig<Theme extends {} = {}> extends VitePlugi
    * @default true
    */
   injectEntry?: boolean | string
+
+  /**
+   * Inject extra imports for every astro page
+   * @default []
+   */
+  injectExtra?: string[]
 }
 
 export default function UnoCSSAstroIntegration<Theme extends {}>(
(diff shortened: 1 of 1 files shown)


antfu:
I don't think we should depreciate the existing ones. But maybe name it `injectExtra: []` or so.

equt:
> I don't think we should depreciate the existing ones. But maybe name it `injectExtra: []` or so.

make sense

Issue #1870, opened 2022-11-11
Feature Request: Resolve styles in pug with VSCode extension


quanzaiyu:
Now VSCode extension cannot resolve styles in pug.

![image](https://user-images.githubusercontent.com/13745132/201277918-849efbce-e727-46d9-accf-1af96008d340.png)

Hover:

![image](https://user-images.githubusercontent.com/13745132/201277959-db50f5d0-e116-4b15-b6a6-620880b516d0.png)

Expect to add this feature in the next release.

PR #1881, opened 2022-11-16
feat(vscode): support annotation for pug

azaleta:
Linked issue: #1870


packages/shared-common/src/index.ts
```diff
@@ -114,6 +170,9 @@ export async function getMatchedPositionsFromCode(uno: UnoGenerator, code: strin
   for (const i of transformers?.filter(i => i.enforce === 'post') || [])
     await i.transform(s, id, ctx)
   const hasVariantGroup = !!uno.config.transformers?.find(i => i.name === 'variant-group')
-  const result = await uno.generate(s.toString(), { preflights: false })
-  return getMatchedPositions(code, [...result.matched], hasVariantGroup)
+
+  const { pug, code: pugCode } = await isPug(uno, s.toString(), id)
```
antfu, line 174: I think it's better to judge it by the `id` instead of guessing from the syntax. To avoid extra perf cost for non-pug files.

antfu (changes requested):

packages/shared-common/src/index.ts
azaleta: @antfu 
The `id` check will be processed by `extractorPug`
https://github.com/unocss/unocss/blob/fac166fe16cde894cf641a3d1cec271d0d0cc0fe/packages/extractor-pug/src/index.ts#L21-L37

```
uno.generate('div.p1.ma')
```
will return null.

The extractor is need for transform pug language(which uno.generate could not recognize)
to get the matched css.

antfu: Ideally `uno.generate` should be fed with already extracted tokens.

azaleta: I think
```
uno.generate('div.p1.ma')
```
will generate correctly at first.
But when I check the extractor implements, I found it returns undifiend :)
https://github.com/unocss/unocss/blob/f2fda68c503d623bd18b8baa60e52cd5bcb59dce/packages/extractor-pug/src/index.ts#L18-L37

This makes `uno.generate -> applyExtractors` could not get extracted tokens
So I call pugExtractor.extract() intend
this will make content change to below before it is be passed to `uno.generate`

```
<template lang='pug'>
div.p3
</template>


<div class="p3"></div>    ----> add extracted part
```

I could do some thing like substring to cut the `<template>` part (may be transform them to extracted tokens?)
What do you think?

PR #1886, opened 2022-11-17
chore: update `nuxt` version

zyyv:


packages/nuxt/src/index.ts
```diff
@@ -63,7 +63,7 @@ export default defineNuxtModule<UnocssNuxtOptions>({
 
     // Nuxt 2
     if (isNuxt2()) {
-      nuxt.hook('config', (config) => {
+      nuxt.hook('app:resolve', (config) => {
```
zyyv, line 66: I'm not sure is this hook right?

Issue #1900, opened 2022-11-22
build, applyCssTransform not work


MellowCo:
[minimal reproducible example](https://github.com/MellowCo/unocss-weapp)

in wechat miniprogram use rpx as size unit，but in uniapp it will according to the compilation environment, rpx will be transferred to rem in h5，so applyCssTransform will be used.

but unocss@0.46.5 in build mode , applyCssTransform was deleted, https://github.com/unocss/unocss/pull/1872

MellowCo:
[minimal reproducible example](https://github.com/MellowCo/unocss-weapp)
# dev
> run in h5
```shell
nr dev:h5
```

> rpx will be transferred to rem, is ok

![image](https://user-images.githubusercontent.com/47939250/203331553-f13e86e0-4951-498c-8194-7072ddbb211d.png)

MellowCo:
[minimal reproducible example](https://github.com/MellowCo/unocss-weapp)

# build
> run build h5
```shell
nr build:h5
```
build in `dist/build/h5`, open it in browser

> rpx not transferred to rem , applyCssTransform was deleted

![image](https://user-images.githubusercontent.com/47939250/203333062-51bf9532-f54d-4140-8a5c-adb00cf4f77b.png)

PR #1902, opened 2022-11-22
revert(vite): applyCssTransform on bundle generation (#1900)

MellowCo:


MellowCo:
It looks like vite has processed the public folder

image

https://github.com/vitejs/vite/blob/892916d040a035edde1add93c192e0b0c5c9dd86/packages/vite/src/node/plugins/css.ts#L204-L210

https://github.com/vitejs/vite/blob/892916d040a035edde1add93c192e0b0c5c9dd86/packages/vite/src/node/plugins/asset.ts#L262-L281

PR #1927, opened 2022-11-29
Update nuxt readme with alternative options config

jbreuil:
Just adding the alternative way to specify option in nuxt config.


antfu:
Thank you. But I don't think we should recommend that usage.

Issue #1939, opened 2022-11-30
transformer-compile-class does not work with *.phtml and *.php files


pbuzdin:
transformer-compile-class does not work with classes inside *.phtml and *.php files

I created an example for this issue https://codesandbox.io/s/recursing-bhaskara-q930ho?file=/package.json:105-209

There are 3 files in src folder: phtml, php and html

You can run an npm for them:
npm run build-phtml
npm run build-php
npm run build-html

PR #1940, opened 2022-11-30
feat: add php and phtml to the default includes (#1939)

pbuzdin:
transformer-compile-class does not works with classes inside *.phtml and *.php files

this fix might help


antfu:
I don't mind having them, but not sure if it would work as they might no go into Vite's pipeline

PR #1971, opened 2022-12-06
feat(preset-mini)!: remove default border-styles on border-x utility

### LINKED ISSUE #1967: Losing border style
Please have a look on below examples:

In [the first example](https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIDuBaATABigAtsBGQgIwHsAnEAUxqmrsexGSPpHQD5gA9OAi8gA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmcAvnOlBCHAOQCuWEAxgM6u0BQnqAHpLBQYAhvQA28NJhz5CwIgAoEnOHCjjUrAFxwA2itV7azeqxjUAtOrGpaAGkRxmEMdB20oqZLTIBdA752BjAAFqggqDrKhnAARp7CANaQ2DDaiAYxtLwc7gDMAEwADGC89pmGtKwgtO4AbAAsJWVBMaq0IN7uAOx1AByl5W10YkS1dACMRQUNg61t2WLjtBMFfc1DCwW8S%2B4TAKx5dXMV7bw7ywdHJzGk86ScpACUQA&options=N4IgzgLgTglgxhEAuaBXApgXyA), everything works fine, the box has dashed border style

In [the second example](https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIDuBaATABigAtsBGQgIwHsAnEAUxqmrsexGSPpHQD5gA9OAi8AUKEixEKDDgJRhYBtgCeZBZCX1sAcxr16AO2wBWfIUXKOSLjzFQoE6PGRp0AF3oAPd9hhHPJgAHPD4oUkchSHtHYSlXDE8fPwDGKBDcMNxI4RineJkPb19-Q0D00N4oAGYc6PEokSA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmcAvnOlBCHAOQCuWEAxgM6u0BQnqAHpLBQYAhvQA28NJhz5CwIgAoEnOHCjjUrAFxwA2itV7azeqxjUAtOrGpaAGkRxmEMdB20oqZLTIBdA752BjAAFqggqDrKhnAARp7CANaQ2DDaiAYxtLwc7gDMAEwADGC89pmGtKwgtO4AbAAsJWVBMaq0IN7uAOx1AByl5W10YkS1dACMRQUNg61t2WLjtBMFfc1DCwW8S%2B4TAKx5dXMV7bw7ywdHJzGk86ScpACUQA&options=N4IgzgLgTglgxhEAuaBXApgXyA), when I add another element with class "divide-dashed", the first element loses its border style

The class "border-dashed" is generated before class "border", so the border style is overridden.

chu121su12:
Aligns with tailwind; close #1967.

Reference: https://tailwindcss.com/docs/border-width


sibbng:
> Hi, I'd prefer to keep it as it is and make it more flexible isn't it?
> 
> :( Too bad my projects are all using `b` to have a border and now I have to add `b-solid` for it. 😭

No, you don't need to. `solid` is the default.

sibbng (approved):

chu121su12:
I'd argue without the set style is more flexible.

As for the default, solid style is set in both

https://github.com/unocss/unocss/blob/cdb4d402f5dea2192e7e3862b7b98351e3ae02f4/packages/reset/tailwind.css#L6-L11

and

https://github.com/unocss/unocss/blob/cdb4d402f5dea2192e7e3862b7b98351e3ae02f4/packages/reset/antfu.css#L1-L4

zyyv (approved):

onur-caplena:
Even if adding CSS resets is a solution, there are use cases where one doesn't want to use a reset. In my case for example I am using unocss as a complementary to Naive-UI. Including resets are breaking pretty much everything. There are only 2 options left: 

1. either add `b-solid` everywhere it is used
2. add 
```
*,
::before,
::after {
  border-width: 0;
  border-style: solid;
}
```
to global css file or preflights with `getCSS`

Now even if the second option is ok, it is probably not very convenient to do when you are using another UI framework. As far as I can see, it is not overlapping with Naive UI, but might be something to consider.

sibbng:
@onur-caplena I posted an alternative solution on https://github.com/unocss/unocss/issues/2285#issuecomment-1452311355

PR #1973, opened 2022-12-06
refactor: improve sorting

sibbng:
This PR does a few things:

### 1. refactors `mergeSelectors` to merge only consecutive selectors
Previously it was merging classes across whole generated results. This was changing the order of some utilities and cause some issues like #1967 and #1978.
   
I tested the merge behaviour with cssnano, esbuild, and lightningcss. As far as I see, they are all only merging the consecutive selectors. I don't even know why we have this feature though.

### 2. Consistent sorting

I made some improvements to our sorting logic. These improvements are bringing it closer to the output of Tailwind. But they caused some mergeable selectors to be rendered far from each other. To fix that, I needed to make more changes.

Roughly sorting changes are:
- Variants always come last
- Pseudos come just before variants 
- Arbitrary CSS properties come before pseudos
- CSS variables (object-$something), come before arbitrary CSS properties


zyyv:
Expired PR that has not been updated for a long time.

PR #1979, opened 2022-12-09
feat(preset): add px to rem preset

chenxch:
## reason
I personally like the unocss design very much, so I applied it to a recently developed mobile project. In addition to using `unocss`, the project also used `vant4` components. Since the design draft is `750px`, `postcss-pxtorem` was used for conversion during compilation. It is normal in the development mode. In the production mode, it is necessary to judge `vant` to return different rootValues. I don’t know if it is because of asynchronous processing that the unocss conversion has a cardinality error.

```ts
// vite.config.ts
import postCssPxToRem from 'postcss-pxtorem'

{
css: {
    postcss: {
      plugins: [
        postCssPxToRem({
          rootValue({ file }) {
            return file.includes('vant') ? 37.5 : 75
          },
          propList: ['*'],
        }),
      ],
    },
  },
}
```

## Why submit this plugin?
I found a rem-to-px conversion in the issue, why is there no px-to-rem? Maybe you can use `postcss-pxtorem`, but when developing in parallel with vant, this problem occurs, you can weigh whether to add this plugin.

### DIFF
--- a/packages/preset-px-to-rem/LICENSE
+++ b/packages/preset-px-to-rem/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 Xc
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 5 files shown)


chenxch:
Maybe I can add this plugin through the community, if possible, please close this pr.

zyyv:
> Maybe I can add this plugin through the community, if possible, please close this pr.

Thanks for your pr. But a long time ago, there was the same pr commited, but this is not what we want, because there are many converted units, the official preset is just a case display, you can use it as your personal preset

PR #1981, opened 2022-12-10
Use async function call instead of top level await

FrankFang:
Some browsers (like wechat's built-in browser) does not support top level await for now


antfu:
The top-level await is on purpose to block the rendering to avoid FOUC. Maybe we should make a gentle fallback, while keeping top-level await on browsers that support it.

PR #1987, opened 2022-12-12
chore(vite): improve hmr

sudongyuer:


sudongyuer:
`Unocss` now using vite 4

antfu:
Thank you! However I am still not convinced by this change, since we will still support Vite 2 and 3.

PR #2026, opened 2022-12-29
feat: remove un-cloak to html node as well as body

vbrajon:
For one use case, I have an "un-cloak" attribute added automatically to the html tag dynamically before the body exist.
Plus I find this intuitive to have the attribute on the html tag, the fact that both work seems better to me.
Thanks for your work !

You can try here the following 2 links (to copy paste in URL):

data:text/html,<html un-cloak>
  <style>[un-cloak]{display:none}</style>
  <script type="module">
    import initUnocssRuntime from "https://patch.deno.dev/@unocss/runtime?dev";
    initUnocssRuntime();
  </script>
  <div>will show</div>
</html>

data:text/html,<html un-cloak>
  <style>[un-cloak]{display:none}</style>
  <script type="module">
    import initUnocssRuntime from "https://esm.sh/@unocss/runtime?dev";
    initUnocssRuntime();
  </script>
  <div>will not show</div>
</html>

The patch is available here: https://dash.deno.com/playground/patch

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -172,6 +172,7 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
     const html = body && body.outerHTML
     if (html) {
       await extract(`${html} ${decodeHtml(html)}`)
+      removeCloak(defaultDocument.documentElement)
       removeCloak(body)
     }
   }


vbrajon:
To check that this works as expected you can do:

document.documentElement.setAttribute('un-cloak', '')

In the preview playground, the runtime automatically delete it: https://deploy-preview-2026--unocss.netlify.app/play/ 
In the production, the runtime let it be: https://uno.antfu.me/play/

PR #2028, opened 2022-12-29
feat: The performance will be improved when there are many layers

SnowingFox:


SnowingFox:
@zyyv hi, can u help me review this?

zyyv:
> @zyyv hi, can u help me review this?

Sure, let me see it.

zyyv:
I personally feel that the performance optimization point here is very small. If there are many layers with complex logic, it is necessary to optimize. It is just one more `filter` loop but I think it can be ignored. 
This is my personal opinion 😅.

SnowingFox:
> I personally feel that the performance optimization point here is very small. If there are many layers with complex logic, it is necessary to optimize. It is just one more `filter` loop but I think it can be ignored. This is my personal opinion 😅.

My idea is that `unocss` as a library, it should consider many situations and optimize according to different situations

zyyv:
I feel a bit too worried here. In actual development, there are not many layers meeted.
Most layers come from each different preset, If there are too many layers, it will also bring worry more.

Let's open it to make more people review it.

antfu:
Same concern. I don't feel the change would have much perf impact, unless we have a benchmark to compare.

zyyv:
> Same concern. I don't feel the change would have much perf impact, unless we have a benchmark to compare.

I updated the benchmarks(`nr bench`) before and after the change and there is only a 2ms gap. :( Sorry I forgot to save the benchmark.

zyyv:
Let's keep it in `draft` status until there is consensus. Please feel free to open or close it.

Issue #1817, opened 2022-10-29
Color preview doesn't show up if there is `>` next to it.


MrFoxPro:
UnoCSS extension: 0.46.1
https://user-images.githubusercontent.com/17229619/198821883-2ed6a1bb-c0ed-47dc-a7ba-4951766f55a1.mp4

chu121su12:
This is by design because we need to match the `>` sign in arbitrary variant instead of splitting it. (ex. `[&>:hover]:something`)

PR #2036, opened 2022-12-31
feat(vscode): better color preview

zam157:
Related issues: #1348 #1767 #1817

The old method of matching colors is **not perfect**. As Fu [commented](https://github.com/unocss/unocss/issues/1348#issuecomment-1207460076): 

> I guess we could infer the color from the generated CSS instead of relying on the theme object.

That's what I've done in this PR

image

image


eweren:
Stumbled on this PR and I guess this is related: When relying on CSS variables for colors this is not resolved. Like here: 
image

Is resolved as black, but should be green (the variable has the same value as the var(...)-fallback value right here

@antfu, @zam157

zam157:
> Stumbled on this PR and I guess this is related: When relying on CSS variables for colors this is not resolved. Like here: 
> image
> 
> Is resolved as black, but should be green (the variable has the same value as the var(...)-fallback value right here
> 
> @antfu, @zam157 

Please open a new issue, and I'll try to solve it if I have time

PR #2047, opened 2023-01-05
fix(preset-attributify): adjust check `prefixOnly` order

### LINKED ISSUE #2037: upgrade from 0.46.5 to 0.47.0 breaks a react project
Upgrade from 0.46.5 to 0.47.0 breaks my build (a react SPA project). Below is the diff before and after, running unocss cli:

Screenshot 2022-12-31 at 20 28 25

and the only places with the word container:

Screenshot 2022-12-31 at 20 29 31

and my unocss.config.ts file:

Screenshot 2022-12-31 at 20 34 41

zyyv:
close #2037 

~Validate `prefixOnly` option in advance. Update the test, found that there was a bug before😅.~


sibbng (commented):
This is not fixing issue #2037.

1. Modified lines are added in [v0.48.0](https://github.com/unocss/unocss/releases/tag/v0.48.0), so it's unrelated to the issue, And the tests are correct. This is expected behaviour for `prefixedOnly` option. It will only apply to attributes, not classes defined in `class` attribute.

2. Both 0.46.5 and 0.47.0 versions behave the same with the config in the [first comment of the issue](https://github.com/unocss/unocss/issues/2037#issue-1515110684). This is expected behaviour for the default splitter and [it will be applied if strict mode is not enabled](https://github.com/unocss/unocss/blob/53a3164c96f59451d9b773cd976642bbd70dfec4/packages/preset-attributify/src/index.ts#L31-L32) or presetAtrributify is completely disabled, like in the first comment.

3. The real issue [in this comment](https://github.com/unocss/unocss/issues/2037#issuecomment-1368212381). And it's not even related to these lines. It's related to these lines in following order: 
https://github.com/unocss/unocss/blob/53a3164c96f59451d9b773cd976642bbd70dfec4/packages/preset-attributify/src/extractor.ts#L43-L49
https://github.com/unocss/unocss/blob/a3c8bed20e8182f99ed217283e3da6a3836991e5/packages/core/src/utils/helpers.ts#L12-L14
We should return false if the selector ends with colon `:` character.

zyyv:
> This is not fixing issue #2037.
> 
> 1. Modified lines are added in [v0.48.0](https://github.com/unocss/unocss/releases/tag/v0.48.0), so it's unrelated to the issue, And the tests are correct. This is expected behaviour for `prefixedOnly` option. It will only apply to attributes, not classes defined in `class` attribute.
> 2. Both 0.46.5 and 0.47.0 versions behave the same with the config in the [first comment of the issue](https://github.com/unocss/unocss/issues/2037#issue-1515110684). This is expected behaviour for the default splitter and [it will be applied if strict mode is not enabled](https://github.com/unocss/unocss/blob/53a3164c96f59451d9b773cd976642bbd70dfec4/packages/preset-attributify/src/index.ts#L31-L32) or presetAtrributify is completely disabled, like in the first comment.
> 3. The real issue [in this comment](https://github.com/unocss/unocss/issues/2037#issuecomment-1368212381). And it's not even related to these lines. It's related to these lines in following order:
>    https://github.com/unocss/unocss/blob/53a3164c96f59451d9b773cd976642bbd70dfec4/packages/preset-attributify/src/extractor.ts#L43-L49
>    
>    https://github.com/unocss/unocss/blob/a3c8bed20e8182f99ed217283e3da6a3836991e5/packages/core/src/utils/helpers.ts#L12-L14
>    
>    We should return false if the selector ends with colon `:` character.


I also realized that the test was correct, forgive my carelessness. I have located the error? Allow me to fix it later due to time difference

Hongbusi (approved):

PR #2048, opened 2023-01-06
chore: fix failed tests

Hongbusi:
After the css file generated by build is compressed, `[font~="mono"]` has no quotation marks.

image

### DIFF
--- a/test/fixtures.test.ts
+++ b/test/fixtures.test.ts
@@ -88,7 +88,7 @@ describe.concurrent('fixtures', () => {
     const css = await getGlobContent(root, 'dist/**/*.css')
 
     expect(css).contains('.w-200px')
-    expect(css).contains('[font~="mono"]')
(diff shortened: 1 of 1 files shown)


Hongbusi:
Ci is normal, there is a problem with the local environment. very strange?😂

zyyv:
CI was error, this right css should has quotes. May your local environment has some problems, try again.

Hongbusi:
I have tried several times locally and there are problems. I have also tried switching the node version. I don't know where the problem is with the local environment.

zyyv:
> I have tried several times locally and there are problems. I have also tried switching the node version. I don't know where the problem is with the local environment.

But your another branch works well. Emm... how about remove your project in your local? and re-clone it.

Hongbusi:
I found this problem in another branch.

But when I tried the method you suggested, the problem did not reproduce. That's strange.😂

Issue #2043, opened 2023-01-03
New sveltekit preprocess cause production break


kamontat:
Sveltekit stable version released with [new default preprocess][new-default-preprocess] cause production build not include preflights and safelist. 

I create stackblitz for demo the issue [here][stackblitz]. 
You can toggle between `vitePreprocess` and `preprocess` variable on svelte.config.js to see the problem.

[new-default-preprocess]: https://kit.svelte.dev/docs/integrations#preprocessors
[stackblitz]: https://stackblitz.com/edit/unocss-unocss-ncmap7?file=svelte.config.js

mustofa-id:
I believe this is due you're utilizing the `svelte-scoped` mode. As mentioned in [svelte-scoped mode example](https://github.com/unocss/unocss/tree/main/examples/sveltekit-scoped#notes), this mode is not currently work with vite plugins.

> Vite plugins can't yet be used to preprocess files emitted by svelte-package as it does not use Vite. Follow https://github.com/sveltejs/vite-plugin-svelte/issues/475 to see when this will be made possible. In the meantime a [temporary svelte preprocessor wrapper](https://www.npmjs.com/package/temp-s-p-u) was published to enable using svelte-scoped mode in component libraries and other context that don't use Vite.

fehnomenal:
Was bitten by the same issue. Using the "old" preprocess works just fine. Now I'm wondering what it would take to allow importing uno.css for the preflights and safelist instead of `<style uno:preflights uno:safelist global></style>`?

Currently `uno.css` can not be found because it is only resolved when using the global mode.

fehnomenal:
Lol this was simple. Just adding

```js
    resolveId,
    async load(id){
      if(!id.match(RESOLVED_ID_RE)){
        return;
      }

      const { css } = await uno.generate("", { preflights:true, safelist:true })
      return css
    },
```

into the `SvelteScopedPlugin` seems to be enough.

Will prepare a PR soon.

PR #2061, opened 2023-01-10
feat(svelte-scoped): allow importing `uno.css` for preflights/safelist in svelte-scoped mode

fehnomenal:
```ts
import 'uno.css';
```
instead of
```svelte
<style uno:preflights uno:safelist global></style>
```

This enables using `vitePreprocess` instead of `svelte-preprocess`.

See #2043 for details

### DIFF
--- a/packages/vite/src/modes/svelte-scoped/index.ts
+++ b/packages/vite/src/modes/svelte-scoped/index.ts
@@ -1,7 +1,7 @@
 import type { Plugin, ResolvedConfig } from 'vite'
 import { createFilter } from '@rollup/pluginutils'
 import type { UnocssPluginContext } from '@unocss/core'
-import { defaultExclude } from '../../integration'
+import { RESOLVED_ID_RE, defaultExclude, resolveId } from '../../integration'
 import { transformSvelteSFC } from './transform'
 
 export * from './transform'
@@ -20,6 +20,14 @@ export function SvelteScopedPlugin({ ready, uno }: UnocssPluginContext): Plugin 
         config.include || [/\.svelte$/, /\.svelte\.md$/, /\.svx$/],
         config.exclude || defaultExclude,
       )
+    },
+    resolveId,
+    async load(id) {
+      if (!id.match(RESOLVED_ID_RE))
+        return
+
+      const { css } = await uno.generate('', { preflights: true, safelist: true })
+      return css
     },
     transform(code, id) {
       if (!filter(id))


antfu:
I am afraid this changes the behavior of `uno.css` under different contexts, which could make misalignments with users' expectations. 

I think probably we could have more explicit ones:

```ts
import 'uno.css?preflight=true&safelist=true&utilties=false';
```

Not very sure tho.

What do you think @jacob-8?

fehnomenal:
Yeah, I would be ok with that. Waiting for @jacob-8's opinion. 

I would then throw if `utilities` is not explicitly `false`. The other ones can be used directly for generating the file.

jacob-8:
Thanks for bringing this up, @fehnomenal! I've been using `svelte-scoped` mode in multiple projects since the update we did to it and have also come to the conclusion that we need to be able to import a global uno css file still, just like the regular UnoCSS mode as there are still some CSS items that need to be global:
- [ ] resets
- [ ] preflights
- [ ] safelist
- [ ] certain plugins like `prose` carry add a large amount of css that is wasteful to be included in every Svelte component that `.prose` class is found in

There may be other things too, but those are the ones I've run into so far.

I have an unsolved problem, however. There is no way to control the order of which SvelteKit places CSS into the `head` tag, and it may be different between `dev` and `build`. So then the `uno.css` file needs to be imported into the `app.html` file right above `%sveltekit.head%` and not in the root `+layout.svelte` so that these global styles can then be overridden by regular classes used in components. A good example of this is how `.prose` has a default `max-width` that you sometimes want to overwrite without having to use `important`. This is not a problem in an app where all styles sit in one global file and can be ordered appropriately, but it is a problem when we spread the CSS across our components in `svelte-scoped` mode. This problem would be easy to solve if Vite's [`transformIndexHtml` hook](https://vitejs.dev/guide/api-plugin.html#transformindexhtml) worked in SvelteKit. It does in Vite + Svelte, but not SvelteKit, and I'm not sure why. I asked about it once but received no response. Perhaps I can dig more, unless you have any ideas on that note @antfu.

If we can brainstorm a preferred way to place the aforementioned UnoCSS styles (ones needing applied globally) into the `head` tag above SvelteKit added styles, then we'll be set and I would love to create a PR for such. Currently my best idea is to just have the `svelte-scoped` plugin write needed global styles to a file sitting in the static folder which the user manually references in their `app.html`, but I think we can do better than that.

fehnomenal:
Ok, so I thought about the problem a bit (or rather a lot :sweat_smile:).

I was able to solve the problem for building with the following steps:
1. Detect if sveltekit: in `configResolved` search for any plugins whose name includes `'sveltekit'`
2. If sveltekit: `this.emitFile({ type: 'asset', name: 'uno.css', source: '<generated css from uno>' })` inside `buildStart` and save the refId
    **Sadly, this does not work in dev**
3. In `renderChunk`
    - find the chunk that contains `hooks.server.(ts|js)`
    - parse the code to find the `handle` export
    - augment it with a new `handle` function that utilizes `transformPageChunk` to finally replace `<head>` with something like `<head><link rel="stylesheet" href="${this.getFileName(unoCssRef)}" />`

Again, this does not work in dev because `emitFile` does not work in dev...

Sadly, I found no way to inject a file into the pipeline that is not imported anywhere - or even to inject this import.

jacob-8:
> does not work in dev

Thanks for chewing on this! Yeah, we need to keep looking for a solution that works in dev also. I posted https://github.com/sveltejs/kit/issues/8619 and will watch to see if that gains any traction, otherwise will proceed with the workaround of writing an actual `uno.css` file into the static folder.

jacob-8:
@fehnomenal can you look over the documentation updates in #2098 and see if the proposal fits the bill?

jacob-8:
Just came to another alternative that is a little simpler for users:

1. They include `<link rel="stylesheet" href="%sveltekit.assets%/uno-preflights-safelist.css">` in their `app.html` head tag after resets and before `%sveltekit.head%`.

2. As done by Vite-PWA for the manifest.json file, the `svelte-scoped` Vite plugin could create a `uno-preflights-safelist.css` file in build using the `generateBundle` hook, and in dev it will be served at `/uno-preflights-safelist.css` using server middleware. 
 
I haven't yet used the generateBundle hook but I'll see about trying a proof of concept. It's the Chinese lunar new year anyhow, so we should go slow on this to give Anthony a break to celebrate.

jacob-8:
> In `renderChunk`
> * find the chunk that contains `hooks.server.(ts|js)`
> * parse the code to find the `handle` export
> * augment it with a new `handle` function that utilizes `transformPageChunk` to finally replace `<head>` with something like `<head><link rel="stylesheet" href="${this.getFileName(unoCssRef)}" />`

@fehnomenal what would this hook implementation look like?

fehnomenal:
Thanks for picking this up @jacob-8 :smile: 

I was sleepless at night, so I thought about different ways to tackle the problem, one also involving a middleware for the vite server. But in the end I really like your idea of the indirection via replacing `__UnoCSS_Svelte_Scoped_global_styles__` at runtime. Well, I don't really really like it but there seems to be no other way currently.
I have an idea (and working PoC) to replace this with a file in prod. I will open a PR to your PR. There is just one thing missing which is basepath handling. It seems that vite does not know about the `kit.paths.base` setting...

> @fehnomenal what would this hook implementation look like? Presently, it looks like using `transformPageChunk` is more promising as I can't find a way to make my latest method cache-busting.

As I wrote this I only thought about build time. Now I think this should be in `transform` and directly depend on the current vite command. I will include this as a sepaate commit in my PR.

fehnomenal:
I guess this won't be necessary anymore with the more sane approach of #2098

Issue #2051, opened 2023-01-07
[Feature Request] "Compatibility Mode" for preset-typography


lixiang810:
It seems that `preset-typography` doesn't work well on some *traditional* browsers because it used the `:where()` selector, which does not work on Chrome before version 88 and Firefox before version 78 according to the [browser compatibility table on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:where#browser_compatibility). It's partially same for `:not()` selector too.

Since I'm working on a website whose users use very old browsers, I need a workaround. It seems like that add a option to disable `not` utility (default to `not-prose`) is the only solution, because with it disabled, the `:where()` and `:not()` selectors are no longer needed.

The main change is below:

https://github.com/unocss/unocss/blob/5812486cd64295cc0b663c0ef2164249d14cc3e1/packages/preset-typography/src/preflights/index.ts#L25-L29

```typescript
// change to
if (match) {
  const matchStr = match[0]
  s = s.replace(matchStr, '')
  if (compatibilityMode) return `${escapedSelector} ${s}${matchStr}`
  return `${escapedSelector} :where(${s})${notProseSelector}${matchStr}`
}
```

If you think this is a good idea, I can make a pull request.

PR #2064, opened 2023-01-10
feat(preset-typography): compatibility mode(#2051)

lixiang810:
See #2051


zyyv:
Hi, it seems to not support for `not-prose` with use compatibility mode.
[Playground](https://deploy-preview-2064--unocss.netlify.app/play/?html=DwEwlgbgBADgTgewM4FMB8AoKVgzQgOwFoQUBbAQwJCgHaowCkAXK5uqAMwFcAbXgJ5QAxtxYIyYAF4UARrxTAA9Hiw4YI3hSRIAvACICCZkXjIU%2B-MVKVqHRizYce-IaPGSZ8xSszLwEGhAA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGABV%2BkUlACGYABaCqIhQFVcESgF84WKBBBwA5F2MS6dW0iQYAHpFjps6rgA28Jg4%2BEQkwKQAFMioUIH0AFxwANpUqCm2Elx0MNYAtPEBGLbkiHBSAdDJtrRotnCmALpUTbJ0WtAw2TB0ybGodtm5BR1dPbY1MJ4w%2BQGkcNMes9DquKQYcJ0AbhhQiUuz0%2BoBrqimsvqMfanpcqIwRhBRAJSywvKMLGycPHz8r3e9wUylUGm0AIGqAAkFJwOpeBxgAFgDB%2BABZCCYZJsLgYO6mN53K4wSTSGJ3VB0CQnDDJACMADoAExA1ASNC4GpaVhgPoAen59BAjI6-LOg0JshahKQQA&options=N4XyA)

lixiang810:
> Hi, it seems to not support for not-prose with use compatibility mode.

This is expected and is explained in the jsdoc and README.

zyyv:
@lixiang810 I modified the code a bit and removed `:is`. I think we should drop all pseudo-classes when using compatibility mode, what do you think?

lixiang810:
> @lixiang810 I modified the code a bit and removed `:is`. I think we should drop all pseudo-classes when using compatibility mode, what do you think?

Nice, I forgot it. The `:is()` selector is also unavaliable before Chrome 88.

zyyv:
> > @lixiang810 I modified the code a bit and removed `:is`. I think we should drop all pseudo-classes when using compatibility mode, what do you think?
> 
> Nice, I forgot it. The `:is()` selector is also unavaliable before Chrome 88.

Let me check it again, it seems that the wrong code was i changed.😅

antfu:
I am a bit hesitant about the name. `compatibility` is a relevant term, anything could be incompatible if their browsers are old enough. While browser list is probably too heavy for us, I think we could have an object instead:

```ts
compatibility: {
  noColonWhere: true,
  noColonIs: true
}
```

lixiang810:
> ```typescript
> compatibility: {
>   noColonWhere: true,
>   noColonIs: true
> }
> ```

The compatibility between `:where()` and `:is()` is the same. Browsers that do not support `:where()` also do not support `:is()`, and vice versa. It seems that it is unnecessary for these two pseudo-classes.

antfu:
> The compatibility of :where() and :is() is the same. Browsers that do not support :where() also do not support :is(), and vice versa. It seems that it is unnecessary for these two pseudo-classes.

This statement is only true in the context of the current existing browsers you are aware of. This may not always be the case, as you mentioned the exception. Users could also use a custom browser engine or anything, that does not always support them at the same time. Also in the future, there could be `:foo` `:bar` or anything. Having them in a single options is not practical.

lixiang810:
Let's make a compromise: use these in the configuration and keep the original, cleaner options in the code.

lixiang810:
OK. Now we have a fine-grained control.

zyyv (approved):

PR #2066, opened 2023-01-10
feat(vite): introduce `hmrTopLevelAwait` option

### LINKED ISSUE #2025: 不兼容
在chrome85版本中，不能引入import "uno.css",会报__uno.css:9 Uncaught SyntaxError: Unexpected reserved word,

### LINKED ISSUE #2031: WeChat APP has a built-in browser. Using unocss in the "development environment" will make the project inaccessible
dev：
image
build：
image

LittleSound:
## Summary

Sometimes I need to develop in restricted environments to debug the compatibility of certain code on a particular platform, but recently unocss has been using top-level await's in development mode, causing my code to not work properly in these restricted environments, which makes it very inconvenient to have to build and release the project for testing.

fix #2031, fix #2025, close #1981

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -396,6 +396,12 @@ export interface ConfigBase<Theme extends {} = {}> {
    * @default false
    */
   details?: boolean
+
+  /**
+   * Whether to use the top-level await
+   * @default true
+   */
+  topLevelAwait?: boolean
 }
 
 export type AutoCompleteTemplate = string
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -396,12 +396,6 @@ export interface ConfigBase<Theme extends {} = {}> {
    * @default false
    */
   details?: boolean
-
-  /**
-   * Whether to use the top-level await
-   * @default true
-   */
-  topLevelAwait?: boolean
 }
 
 export type AutoCompleteTemplate = string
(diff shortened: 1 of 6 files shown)


zyyv:
Similar to #1981

antfu:
I think we should better make an option, so ppl can opt-out if needed, while the majority could still benefit from it.

LittleSound:
> I think we should better make an option, so ppl can opt-out if needed, while the majority could still benefit from it.

I added a switch to the top-level await feature in the configuration. I'm not familiar with the unocss code yet, so feel free to point out if there is something wrong with the changes

playground/vite.config.ts
```diff
@@ -13,7 +13,9 @@ export default defineConfig({
   },
   plugins: [
     Vue(),
-    Unocss(),
+    UnoCSS({
+      topLevelAwait: false,
```
zyyv, line 17: You can test it on playground.

packages/core/src/types.ts
```diff
@@ -396,6 +396,12 @@ export interface ConfigBase<Theme extends {} = {}> {
    * @default false
    */
   details?: boolean
+
+  /**
+   * Whether to use the top-level await
+   * @default true
+   */
+  topLevelAwait?: boolean
```
antfu: It should be in Vite plugin's config. And maybe `hmrTopLevelAwait`?

zyyv: This is not latest code, i have moved.

LittleSound:
Thank you very much for helping me to improve the code

Issue #2067, opened 2023-01-10
Regex Endless Loop stalling build when using attributify mode and < char


MiniDigger:
This is a really weird issue and took me waaaaaaaaaayyyy to long to figure out.
I converted a project from windi to unocss today and everything worked great, locally, but my build on github actions stalled. upon verification locally, I could reproduce that.
now followed a few hours of digging and I have pinned the issue down: using < in classes in certain combinations causes a regex catastrophic backtracking.
I managed to debug my way thru stuff and eventually managed to create this minimal repro: https://stackblitz.com/edit/github-y3a65o?file=app.vue
notice how stuff seems to work in pnpm run dev, but pnpm run build just freezes. Removing the < in line 9 fixes the freeze. removing the space + d behind color-red also fixes the freeze.
The issue is this regex: https://github.com/unocss/unocss/blob/main/packages/preset-attributify/src/extractor.ts#L11
If you take that regex and the code that unocss scans for that component into regex101, it will complain about catastrophic backtracking: https://regex101.com/r/Ib1eKA/1
(I obtained the code by console logging the code var here https://github.com/unocss/unocss/blob/main/packages/preset-attributify/src/extractor.ts#L24, before the regex match)
Remove the < and it works, remove the space after the color-red and it works too (as in no match, but no endless loop either).

thats pretty much all find I found out tonight, hopefully that is somewhat helpful to hunt this one down, I tries to understand the regex, but its 2am now and I didn't come far, lol

ps, I know < should be lt in unocss, after talking about the solution to this issue with friends, but I still think this is a valid issue, < could be located elsewhere in the code and theoretically trigger this I think.

zyyv:
Hi, check this docs. https://github.com/unocss/unocss/blob/main/packages/preset-wind/README.md#differences-from-windi-css

MiniDigger:
yes, I was made aware, I said in the last sentence, but its still a valid issue

chu121su12:
Smaller subset of string that's causing this:

https://regex101.com/r/48y4HT/1

PR #2069, opened 2023-01-11
feat(core): make generator can timeout

chu121su12:
As touched in #2067, due to the nature of the usage of regex in the presets, I'd like to propose to add timeout option to the ```uno.generator()```

I'm still not sure what happend to the function calling the regex when a catastrophic backtrack happened but for now the generator will just exit early.

There are 3 possible sections to apply the detection independently: extractor, token parsing and preflights. But, as a proof of concept, for now it will be just as a global wrapper to the main generator function.

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -121,6 +121,29 @@ export class UnoGenerator<Theme extends {} = {}> {
     options: GenerateOptions = {},
   ): Promise<GenerateResult> {
     const {
+      timeout = 0,
+    } = options
+
+    if (timeout === 0)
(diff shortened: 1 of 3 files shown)


antfu:
Thank you. But I don't know if this belong to the core. As the same situation could happen with any promise you have. And if the performance issue was caused by RegEx, I don't think this could help much as the execution is Sync (you can't abort sync operations).

I would suggest to not have it, and could suggest users to do it on the userland if you really consider this is a helpful trick.

chu121su12:
The regex part bothered me too since there are little resource on what happened on the backtrace.

PR #2071, opened 2023-01-11
fix(core): replace scope placeholder globally

### LINKED ISSUE #2068: Using multiple variants can be problematic
When I use both dark and scope variant, the output is wrong.

image

The possibly problematic code is as follows:
```
// The expression does not have global matching enabled
export const regexScopePlaceholder = / \$\$ /
export const hasScopePlaceholder = (css: string) => css.match(regexScopePlaceholder)

function applyScope(css: string, scope?: string) {
  if (hasScopePlaceholder(css))
    return css.replace(regexScopePlaceholder, scope ? ` ${scope} ` : ' ')
  else
    return scope ? `${scope} ${css}` : css
}
```

chu121su12:
Fix #2068 

I'd tag this as breaking since the replacement is now global while exported and being used in 2 other places:

https://github.com/unocss/unocss/blob/badcf8666a43bcb640d1a2d1f8f8318b2ec27047/packages/vscode/src/selectionStyle.ts#L54

https://github.com/unocss/unocss/blob/badcf8666a43bcb640d1a2d1f8f8318b2ec27047/packages/transformer-directives/src/apply.ts#L66

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -637,11 +637,10 @@ export function createGenerator<Theme extends {} = {}>(config?: UserConfig<Theme
   return new UnoGenerator<Theme>(config, defaults)
 }
 
-export const regexScopePlaceholder = / \$\$ /
-export const hasScopePlaceholder = (css: string) => css.match(regexScopePlaceholder)
-
+const regexScopeCheck = / \$\$ /
+export const regexScopePlaceholder = / +\$\$ +/g
 function applyScope(css: string, scope?: string) {
-  if (hasScopePlaceholder(css))
+  if (css.match(regexScopeCheck))
     return css.replace(regexScopePlaceholder, scope ? ` ${scope} ` : ' ')
   else
     return scope ? `${scope} ${css}` : css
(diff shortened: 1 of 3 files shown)


antfu:
LGTM, but I didn't get what is breaking.

packages/core/src/generator/index.ts
```diff
@@ -637,11 +637,10 @@ export function createGenerator<Theme extends {} = {}>(config?: UserConfig<Theme
...
 }
 
-export const regexScopePlaceholder = / \$\$ /
-export const hasScopePlaceholder = (css: string) => css.match(regexScopePlaceholder)
-
+const regexScopeCheck = / \$\$ /
+export const regexScopePlaceholder = / +\$\$ +/g
 function applyScope(css: string, scope?: string) {
-  if (hasScopePlaceholder(css))
+  if (css.match(regexScopeCheck))
```
antfu, line 643: export const regexScopePlaceholder = /\s\$\$\s+/g
export const hasScopePlaceholder = (css: string) => css.match(/\s\$\$\s/)

function applyScope(css: string, scope?: string) {
  if (hasScopePlaceholder(css))

antfu (approved):

PR #2081, opened 2023-01-15
docs: add create-react-app example

wangfengyuan:
I once tried to use unocss in an existing create-react-app project, refer to  https://github.com/unocss/unocss/discussions/1781 , for the create-react-app project, it is not easy to import unocss to use, it took me a while to solve it. the main thing is you have to figure out the problem and modify and override the default configuration, In order to prevent others from encountering the same problem, I would like to add a usage example


wangfengyuan:
I fixed the eslint script error and added some comment

antfu:
I am not even sure if CRA should still be recommended today - their team seems not actively maintain it, and migrating CRA project to Vite sounds straightforward.

Anyway, I am ok to have it, as long as we can keep the example minimal as possible, which means:

- No assets (images, robot.txt, etc), we could use icons preset for the logo
- No documentation of CRA/React themselves - as it's not related to the example
- No CSS if possible

Thanks

KeyToLove:
> I once tried to use unocss in an existing create-react-app project, refer to #1781 , for the create-react-app project, it is not easy to import unocss to use, it took me a while to solve it. the main thing is you have to figure out the problem and modify and override the default configuration, In order to prevent others from encountering the same problem, I would like to add a usage example

I'm stuck right here just now ,thanks for your demo :)

KeyToLove:
Did you met HMR problem ?  i try use this configuration, but the HMR is broke.
When i change `class` to update styles, i have to refresh my browser. The same problem also occur in webpack project without `CRA`,maybe it's a `@unocss/webpack` issue ?

wangfengyuan:
> 



> 


> Did you met HMR problem ? i try use this configuration, but the HMR is broke. When i change `class` to update styles, i have to refresh my browser. The same problem also occur in webpack project without `CRA`,maybe it's a `@unocss/webpack` issue ?

seem to be the webpack did not detect the virtual file _virtual_/__uno.css has changed, i guess

antfu:
Closing it as I consider CRA to be soft sunsetting (new React docs no longer mention it).

PR #2082, opened 2023-01-15
feat(runtime): add alternate variant with its own <style> element

chu121su12:
Add a separate layer with its own variant & style element to enable css override without adding important modifier.

The exsiting `<style>` element is intentionally appended early (before `<head>`) so as to make unocss act as base styling. This in turn making overriding style difficult and may resort to selector manipulation or using the important keyword/modifier. In userspace, developer can move the style element around, but considering the preflights & reset styles are here, moving the style element around may not be optimal.

This PR make unocss runtime produce 2 `<style>` elements and split the generator output into two parts. One with a configurable layer variant will be on the new `<style>` element, placed after `</body>`. The rest assume the previous behavior having the css before the `<head>`.

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -1,6 +1,6 @@
 import type { GenerateResult, UnoGenerator, UserConfig, UserConfigDefaults } from '@unocss/core'
 import { createGenerator, isString, toArray } from '@unocss/core'
-import { autoPrefixer, decodeHtml } from './utils'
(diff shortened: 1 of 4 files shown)


antfu:
I am not very sure if this option can be confusing. Or do you think it's better that we support arbitrary numbers of layers and the combination of style elements, so it has the full ordering power as in the build pipeline, instead of treating some layers specially.

antfu:
Close in favor of https://github.com/unocss/unocss/pull/2131

PR #2093, opened 2023-01-18
feat(preset-mini)!: allow empty string for variablePrefix

### LINKED ISSUE #2085: Allow variablePrefix to be disabled
Please allow me to pass `false` to the `variablePrefix` to fully remove the `un-` prefix from all generated custom css properties. An empty string doesn't work right now due to the `.length` check in the transformer

chu121su12:
Close #2085

Tagged as breaking since previously empty string will become the default `--un` instead.


zyyv:
How to set the type of `variablePrefix` to `boolean | string`?
- If variablePrefix's value is false, it means to disable `variablePrefix` default string(use empty string). 
- If variablePrefix's value is true, it means use `un-`. 
- If it's a string, it means user custom variablePrefix.

PR #2098, opened 2023-01-20
feat(svelte-scoped): inject preflights/safelist to app.html

### LINKED ISSUE #1920: [examples/sveltekit-scoped/] Trying to update the example with `preset-typography` without much luck.
Hello @jacob-8 I have made a fork of the example with some minimal changes that I describe in the README header.

Here is the repo: https://github.com/bySabi/example-unocss-sveltekit-scoped

I think the problem is that the `preset-typography` doesn't get loaded globally.
 
The _Visual Studio Code UnoCSS_ extension also does not highlight the `prose` class

It would be nice to have a `preset-typography` example with SvelteKit. I am not able to find any.

### LINKED ISSUE #2043: New sveltekit preprocess cause production break

jacob-8:
closes #2043 and is a result of the conversation in #2061 - the main point is that we need a way to have some global CSS styles put into the head tag before Svelte adds styles from components (injected in place of %sveltekit.head%) for CSS specificity reasons.

- [x] update Svelte/SvelteKit example apps to latest versions (SvelteKit 1.0!)
- [x] deprecate previous `<style uno:preflights uno:safelist global>` method in SvelteKit by logging a warning to the console 
- [x] update `unocss:svelte-scoped` to place preflights and safelist into a global styles tag (dev) or a global styles hashed and cacheable file (build) that is included in `head` tag of `app.html` or `index.html`
- [x] check if a user has properly implemented the transformPageChunk hook in `hooks.server.js` if they have tried to implement it
- [x] add example of how to use typography plugin, closes #1920.
- [x] add support for preset-web-fonts and give an example usage 
- [x] docs: list the UnoCSS imports in a simpler manner in the example repos and put the config into uno.config.ts
- [x] ignore safelist classes when found in component
- [x] docs: update instructions regarding presets usage and which ones work with `svelte-scoped` mode
- [x] docs: update docs for `svelte-scoped` usage outside of SvelteKit with a new example repo: `vite-svelte-scoped`
- Unrelated: fixed import typo in `packages/unocss/src/preset-tagify.ts`

After merging, create a new issue: How to **optionally** include a reset option since the normal route of importing a reset from uno in a main.ts file is not available? I don't know how to optionally do this since no custom config can be passed to `svelte-scoped` beyond the normal UnoCSS config.


packages/vite/src/modes/svelte-scoped/index.ts
```diff
@@ -3,37 +3,72 @@ import { createFilter } from '@rollup/pluginutils'
...
     },
+
+    configureServer: logErrorIfTransformPageChunkHookNotRight,
+
+    // build hook
+    renderChunk(code, chunk) {
+      if (isSvelteKit && chunk.moduleIds.some(id => isServerHooksFile(id))) {
+        const base = viteConfig.base ?? '/'
+        const unoCssHashedLinkTag = `<link href="${base}${this.getFileName(unoCssFileReferenceId)}" rel="stylesheet" />`
+        return code.replace(GLOBAL_STYLES_PLACEHOLDER, unoCssHashedLinkTag.replaceAll(/'/g, '\''))
```
jacob-8, line 70: @fehnomenal why is replaceAll needed here when it's just a link tag that we control?

fehnomenal: Oh, you're right it isn't needed. I just copied it from the other location...

fehnomenal:
Re single quotes:

Now the quotes are dynamically checked and escaped. It works with all kind of quotes.

---

Re hooks.server location:

I split the check to both include `hooks` and `server`. So someone putting the file into `src/hooks/server.js` will get no error.

jacob-8:
> Now the quotes are dynamically checked and escaped. It works with all kind of quotes.

Great, thanks for making all three quote types work!

I think we should leave off the `openingQuote !== closingQuote` check as that's the job of the code editor's static analysis tools. It's impossible for me to make that sort of typo without big red notices.

jacob-8:
@fehnomenal - I have been mentally piling up a number of updates to this plugin, safelists and preflights styles location being the foremost of them. Thanks for being the one to carve out a path and get this going!

fehnomenal:
I just tried to push a commit and saw you also noticed the backticks problem :smile:

packages/vite/src/modes/svelte-scoped/global.ts
```diff
@@ -40,7 +40,7 @@ export function logErrorIfTransformPageChunkHookNotRight(server: ViteDevServer)
     res.write = function (chunk, ...rest) {
       const str = (chunk instanceof Buffer) ? chunk.toString() : ((Array.isArray(chunk) || 'at' in chunk) ? Buffer.from(chunk).toString() : (`${chunk}`))
 
-      if (str.includes('%unocss.global%') || str.includes(GLOBAL_STYLES_PLACEHOLDER)) {
```
jacob-8: @fehnomenal thanks!

fehnomenal:
I just noticed I still need the import functionality from my original proposal in #2061. I am building widgets in plain svelte that are to be embedded into other sites without being possible to modify the html.

Maybe `resolveId` and `load` can be guarded behind `!isSvelteKit`. What do you think, @jacob-8 ?

jacob-8:
> I am building widgets in plain svelte that are to be embedded into other sites without being possible to modify the html.

Can you add more details about your goal and how you are in a situation where you are using Vite but can't touch the html? Maybe a repro? 

Are you building a component library that uses `svelte-package` to package up? I'm also doing the same in a few repos of mine and use UnoCSS via a [svelte preprocessor](https://www.npmjs.com/package/temp-s-p-u) that was supposed to be temporary. I was recommended by userquin to not add it to this repo, but to wait until Svelte introduces a way to run Vite plugins in the `svelte-package` pathway. I don't see that happening soon and so I plan to update my plugin and make it easily findable once this PR merges. 

I recommend you compile the classes on build time and don't make the consuming projects depend on UnoCSS. I do this in a few packages and can share examples if helpful.

fehnomenal:
> Can you add more details about your goal and how you are in a situation where you are using Vite but can't touch the html? Maybe a repro?

Yeah sure. We want to provide some functionality for the web pages of clients. They must be able to insert some html code (some `script` tags and maybe a little bit of configuration somewhere) into their cms or plain html files and when visiting the page the spa will be loaded and access our backend.

Honestly, this is not new but already exists as an older react codebase but I was hoping to build the next generation of widgets with svelte and unocss. Anyway, we need also a documentation page for each widget and fixed urls for the script files so the clients do not need to update their markup.

So, with the background out of the way to answer your questions: I use vite for the dev server and I can indeed touch the html but only the html of the documentation page and not the host pages.

To get stable urls I have for the search widget the following rollup inputs:
```js
      rollupOptions: {
        input: {
          'widgets/search': path.resolve(__dir, 'src', 'widgets', 'search', 'index.ts'),
          'search': path.resolve(__dir, 'search.html'),
        },
        // more
```

After building I get these files:
```
build
├── assets
│  ├── index.9c52b632.css              from uno.css whether there is ?inline appended or not
│  ├── search.5c167a30.css             styles used in /search.html these are tailwind resets which I don't want to force onto all clients
│  └── search.bb96848f.js              preload polyfill
├── search.html                        documentation page
└── widgets
   └── search.js                       file to be included by the clients, code for the svelte components
```

Now `search.html` loads both css files from `assets/` and the main widget script via the polyfill chunk (I hope this is only the case because it is the only svelte widget right now and changes after adding more widgets). This is fine for the documentation page but I need somehow this `index.hash.css` file either inlined or have a static name so I can give an url to the clients.

Maybe you have a better idea how to solve this?

jacob-8:
Thanks for the context, @fehnomenal. I've never built components use Vite + rollupOptions so I'm still a little fuzzy without a repro. I've only done full apps with SvelteKit, and component libraries with `svelte-package`.

If each of your widgets are only 1 component, could you do `<style uno:preflights uno:safelist>` without the global flag as that is already inlined? Or include the uno:... snippets in every component? If that doesn't work because of some dynamic classes which wouldn't include svelte's css scoped component hash, then can you just stick with `svelte-preprocess` and not move to the `vitePreprocess` if that is what broke things?

fehnomenal:
> Thanks for the context, @fehnomenal. I've never built components use Vite + rollupOptions so I'm still a little fuzzy without a repro. I've only done full apps with SvelteKit, and component libraries with `svelte-package`.

Alright, I will prepare a little repro for you later :joy: 

> If each of your widgets are only 1 component, could you do `<style uno:preflights uno:safelist>` without the global flag as that is already inlined? Or include the uno:... snippets in every component? If that doesn't work because of some dynamic classes which wouldn't include svelte's css scoped component hash, then can you just stick with `svelte-preprocess` and not move to the `vitePreprocess` if that is what broke things?

You're probably right but I was afraid of the deprecation. I also need to check whether the `:root {...}` stuff is correctly applied without `global`. I'd rather not use `global` to prevent changing the host pages and I had hoped that the "global" css file only included the css vars. Also I want to use `svelte-scoped` because of the nice compilation without the `:uno:` prefix :joy:

fehnomenal:
`<style uno:preflights uno:safelist>` works really great. At least as long as I have no real preflights and only the css variables, I guess.

> Alright, I will prepare a little repro for you later :joy:

Here you go, @jacob-8: https://github.com/fehnomenal/vite-svelte-widgets. I tried to keep it simple but this approach I want to use in production. The code should be rather self-explanatory.

jacob-8:
> I noticed the deprecation is only for sveltekit which is fair enough

Yes, without the ability to determine the order in which SvelteKit place styles into the head tag, `<style uno:preflights uno:safelist global>` leads to some really confusing problems where CSS specificity ordering is the culprit. For example, why won't `max-w-full` override the default width set by `.prose`? The default usage of utility styles tools is premised on the usage of a global stylesheet where the utility tool is smart and just takes care of all the ordering for you. We've broken that model with this beefed up scoped mode that has inter-component abilities and need to keep guardrails in place for those not aware of the dangers.

As to the wider Svelte ecosystem, anything can be built and your use case is a good example (thanks for the repro). At the risk of too many options, I think a note about using `uno:preflights uno:safelist` needs brought back into the documentation.

fehnomenal:
Btw, how do you feel about including the shortcuts into the compiled class? It seems strange for me to get the nice hash `uno-yj5215` followed by `card` :joy:

jacob-8:
@fehnomenal I just added a documentation note explaining your use case is available and I made the SvelteKit warning more clear (as well as fixed a bug that would have kept `uno:preflights` from working outside SvelteKit, lol, after creating a test for that use case).

jacob-8:
> Btw, how do you feel about including the shortcuts into the compiled class? It seems strange for me to get the nice hash `uno-yj5215` followed by `card` 😂

I strongly prefer when classes state what they do as it's much easier for inspection. We're only hashing because we have to for the inter-component usefulness, and this is why in dev mode we keep the class name in the hash (albeit a bit ugly to make it work). In the case of a shortcut, we don't need to hash for functionality and I feel we would be losing something by doing away with the name.

If you're referring just to production for a little minification benefit, I could see an argument being made, but first the transform function needs some refactoring and clean up to make it easier to modify. At the moment, if I understand your idea, this would add complexity so I'd like to pass on it.

fehnomenal:
Hi @jacob-8, I thought about it just so the classes are consistent. But I guess having stable shortcut classes is good for me to allow the clients somewhat to customize the styling.

jacob-8:
We had a few more changes here that we made after marking it ready, but I think we're all set for your review, @antfu.

jacob-8:
@antfu, please let me know if this PR is too messy to easily review. It took us a bit to figure out everything, and now that we know what we want, it'd be easy to break the changes apart into smaller pieces if that'd help the review process.

fehnomenal:
I rebased on the latest main

jacob-8:
@fehnomenal In our extended wait for these changes to reviewed and merged, I've realized that for all things `svelte-scoped` related it is probably better to create a separate project that depends on UnoCSS. Here's my primary reasons:
- Since we don't have just 1 stylesheet that everything is dumped into, many things don't work right:
  - UnoCSS inspector
  - UnoCSS recommended way of importing reset
  - many of the transformers
  - and more
- The maintainers [rejected](https://github.com/unocss/unocss/pull/1725#issuecomment-1282163798), and probably rightly so, the idea of a svelte preprocessor. This is needed for libraries that use `svelte-package`. I also wanted to add an example project of this in use, but the Svelte examples in this repo were already getting to be too many.
- The feedback loop is too slow for making updates (as evidenced in this PR)
- Tests are far removed from the actual code. I find I do more testing when things are colocated.

Previously there was talk of UnoCSS being the base for the next version of Windicss - so I thought, hey, I should just create [svelte-scoped-uno](https://github.com/jacob-8/svelte-scoped-uno) and so I did. All of the improvements of this PR, are implemented there as well as [svelte-preprocessor-unocss](https://github.com/jacob-8/svelte-scoped-uno/blob/main/packages/svelte-preprocess-unocss/README.md), and a few other things. Please feel free to create PRs in that repo for anything further you'd like to add. :)

A big thank you, @fehnomenal, for all your help on these improvements! You can install by reading https://github.com/jacob-8/svelte-scoped-uno/blob/main/packages/svelte-scoped-uno/README.md#setup

jacob-8:
@antfu Since we have a more full-featured `svelte-scoped` option available now at https://github.com/jacob-8/svelte-scoped-uno - what updates would you like me to make to the UnoCSS project? Do you want me to create a new PR to simplify down the docs and point `svelte-scoped` people over to my project that I've branched off? I feel like you'll get more of these sort of requests, https://github.com/unocss/unocss/issues/1920, if we don't point people to the full working option.

fehnomenal:
I love it and will try the new approach right now :heart:

jacob-8:
> I love it and will try the new approach right now :heart: 

We can also pass options to the plugin like how you saw I do with the reset.

Issue #1764, opened 2022-10-17
unocss vscode extension root configuration item does not support configuring multiple subdirectories 


zhump:
my project directory

```
├── node_modules
├── package.json
├── packages
        ├── web-demo1
        ├── web-demo2
├── pnpm-lock.yaml
├── pnpm-workspace.yaml
```
<img width="728" alt="image" src="https://user-images.githubusercontent.com/7962067/196161735-eaa8109a-aa89-41bb-9e80-5071c9a1ff23.png">

unocss vscode extension root configuration item does not support configuring multiple subdirectories 
I would like to support configuring multiple folders
```
{
  "unocss.root": "packages/*"
}
```

how can i solve this problem？

azaleta:
web-demo1 or web-demo2
do not support multi entry now.

zhump:
> web-demo1 or web-demo2 do not support multi entry now.

Are there plans to support this feature? tailwindcss is currently supported.

dxhuii:
我也有这样的问题，啥时候能支持？或者这样也行
```json
{
  "unocss.root": ['apps/a', 'apps/b', 'packages/a', 'packages/b']
}
```
数组的形式也行。

SKT1yang:
monorepo项目就是没提示，很郁闷啊 啥提示没有

SKT1yang:
项目运行没问题，{
  "unocss.root": 'config/vite/src/unocss.config.ts'
}
外部vscode插件不生效，

tobiasdalhof:
Do you have different unocss configs across your monorepo?  

If no, why not create a single unocss config file in your project root and pass the config path in your vite config?

```ts
// vite.config.ts in ./example

import { resolve } from 'path'
import { defineConfig } from 'vite'
import UnoCSS from 'unocss/vite'

export default defineConfig({
  plugins: [
    UnoCSS(resolve(__dirname, '../unocss.config.ts')),
  ],
})

```

panzhangguo:
> 项目运行没问题，{ "unocss.root": 'config/vite/src/unocss.config.ts' } 外部vscode插件不生效，

处理了吗

PR #2101, opened 2023-01-20
feat(vscode): configure multiple subdirs

MellKam:
https://github.com/unocss/unocss/issues/1764

now you can pass array to unocss.root
```json
"unocss.root": ["apps/a", "apps/b"]
```
or just a string as before 
```json
"unocss.root": "apps/c"
```

### DIFF
--- a/packages/vscode/package.json
+++ b/packages/vscode/package.json
@@ -41,7 +41,13 @@
           "description": "Disable the UnoCSS extension"
         },
         "unocss.root": {
-          "type": "string",
+          "type": [
+            "array",
+            "string"
+          ],
+          "items": {
+            "type": "string"
+          },
           "description": "Project root that contains the UnoCSS configuration file"
         },
         "unocss.underline": {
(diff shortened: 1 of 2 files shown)


MellKam:
now you can pass array to unocss.root
```json
"unocss.root": ["apps/a", "apps/b"]
```
or just a string as before 
```json
"unocss.root": "apps/c"
```

tobiasdalhof:
Why not use the first unocss config found when searching the file tree in reverse for convenience?

MellKam:
> Why not use the first unocss config found when searching the file tree in reverse for convenience?

But what if I want to use more than one config in my project? For example I have uikit and frontend app in one monorepo.

tobiasdalhof:
> UPD: Maybe I misunderstood you. Do you want to search from the currently open file up the directory tree to the first configurator found?

Yes, that's what I ment

antfu (approved):

PR #2107, opened 2023-01-24
feat: allow to return generated css inline

fehnomenal:
See https://github.com/unocss/unocss/pull/2098#issuecomment-1401844999

Vite seems to need `?inline` at the resolved id.
- [ ] docs?
- [ ] test
- [ ] get this actually to work lol


lisonge:
@fehnomenal 

hello, why the module `virtual:uno.css?inline` still has sideEffects ?

when `vite build`, why the `default export` of `virtual:uno.css?inline` is `''` ?

fehnomenal:
I'm sorry, I really cannot answer your questions. I guess the exported value depends on the uno config and the mode.

lisonge:
> exported value depends on the uno config

I didn't find it, where is it ?

lisonge:
about it, you can see <https://github.com/unocss/unocss/issues/2533>

PR #2119, opened 2023-01-27
Fix class syntax in Scanning chapter

damienroche:


antfu:
Thank you. But here, it was just explaining the idea, it doesn't need to work, or bond to any framework syntax. I'd avoid to introduce Vue's concept here, as UnoCSS works agnostically.

Issue #2127, opened 2023-01-29
The background color of some button styles in component libraries are lost when referencing the tailwind.css file.


kirklin:
## Issue

When using the unocss/reset package, the background color of some button styles in component libraries are lost when referencing the tailwind.css file. 
For example, this can be quickly reproduced in both the naive ui and element ui.
![reproduced](https://user-images.githubusercontent.com/17453452/215326031-e6987827-1ff0-4866-996a-13b23e1636d2.gif)
The problem is caused by this CSS, which overrides the button styles of third-party UI libraries:
```css
/*
1. Correct the inability to style clickable types in iOS and Safari.
2. Remove default button styles.
*/

button,
[type='button'],
[type='reset'],
[type='submit'] {
  -webkit-appearance: button; /* 1 */
  background-color: transparent; /* 2 */
  background-image: none; /* 2 */
}

```
## Solution

To make it easier for future integration of the unocss package in projects, I have submitted a pull request . This pull request will create a css file that inherits from tailwind. And comment out this line of code.

```css
button,
[type='button'],
[type='reset'],
[type='submit'] {
    -webkit-appearance: button; /* 1 */
    /*will affect the button style of most component libraries, so disable it*/
    /*background-color: transparent; !* 2 *!*/
    background-image: none; /* 2 */
}
```
## Expected Outcome

The expected outcome is that the background color of the button styles will no longer be lost when using the unocss/reset package.

kirklin:
Link -> https://github.com/tailwindlabs/tailwindcss/discussions/5969

YunYouJun:
`[type='button']` level is hight, i think maybe we can just exclude `[type='button']` `background-color: transparent` for it.

PR #2128, opened 2023-01-29
feat(reset): add tailwind-compat.css

kirklin:
fix background color of some button styles in component libraries are lost when referencing the tailwind.css file.
Link: https://github.com/unocss/unocss/issues/2127


antfu:
That's something I have wanted for a while! Here are a few questions about it that we better address before having it:

- We should give a clear definition of how it works. Like "Based on Tailwind's preflight but without background color", or a list of the changes in the README. This will help users understand it, and also for future maintenance to know how to update (or keep sync with Tailwind).
- Maybe `@unocss/reset/minimal.css`? Since both `uno-` and `reset` are already present and does not provide more information.

kirklin:
Do I need to rename `uno-reset.css` to `minimal.css` ?

PR #2130, opened 2023-01-29
feat(playground): upgrade to codemirror 6

sibbng:
Everything should be fine. Tested in inspector too. Seems ok.

Unrelated; how do I get rid of these type errors?

image


antfu:
Nice work, thanks!

PR #2131, opened 2023-01-30
feat(runtime)!: split uno-layer into separate <style>

chu121su12:
Alternative to #2082

> https://github.com/unocss/unocss/issues/2082#issuecomment-1403474842

Each uno-layer now get inside its own `<style>` element. One concern is the style placement in dom is done only once and may be different depending on the location of the previous layer's `<style>`


antfu (approved):
LGTM, would you consider this breaking change or not?

chu121su12:
Technically yes as this PR changes how the <style>s in dom are generated. If developers move the original <style> around, the behavior now may/may not work differently. Otherwise, if the developers do not move the original <style> element they would not precieve any change.

Please mark this as breaking if deemed so.

PR #2142, opened 2023-02-03
feat(preset-wind): add support for max-width in container

praburangki:
Enable custom max-width for container classes.

References: https://getbootstrap.com/docs/5.3/layout/containers/#sass


packages/preset-wind/src/rules/container.ts
```diff
@@ -10,34 +11,39 @@ export const container: Rule<Theme>[] = [
...
+            if (matchBp && !isString(themePadding))
               padding = themePadding?.[matchBp] ?? padding
           }
         }
       }
 
       const css: CSSObject = {
-        'max-width': width,
+        'width': '100%',
+        'max-width': maxWidth,
```
antfu, line 46:
```suggestion
        'max-width': maxWidth || '100%',
```

Can we do this and avoid adding a new attribute?

praburangki: The reason that was added so the content is always expand the container. If we remove the `width` then it would not get that effect.

antfu (approved):

PR #2155, opened 2023-02-06
feat(core,preset-attributify)!: allow matching bracket-only last value

### LINKED ISSUE #2147: Custom variants with custom values do not work in attributify mode
### Issue description
In `attributify mode`, combining custom `variants` and `values` does not work as expected.
Consider: `<div bg="[&:nth-child(3)]:[#fff]"> ... </div>`
This should apply the `#fff` `background-color` to the `third` div element, but instead incorrectly attempts to apply `&:nth-child(3)` as the `background-color`.

### Version
`v0.49.4`

### Issue reproduction
See [playground](https://uno.antfu.me/play/?html=DwEwlgbgBARg5gXgEQG0BkAuAdgFwBYC0AxnmADYgAUAjAJQC6GATgKYhIB8AwgK4DOOAPYBbKBACGTMONxQA1LB7kcBMFjHiyPFlADugpgGs%2BwAPTgIHAFChIsREjyCILJhhQBiagCYAzABYAVgA2ek4AISUyFTUNKRkceSgifiFRCS0dfSMTc0hrW2giMnE%2BPmQAM0FBJHtkdAA6KsFGdGx8YlIKSl8Gdy8-INDOXgEROOlZBRSx9M1tKBBBFj4oLEFE7MMAGmTU8Yl42TBVtSIDViIcMgBPKH42KFKNTKhYuBYsV3EcR64AZX%2BZgsBQsyRKZUq1SgjWarUwuEIJHIVH8fXgBE8PgCITC3H2cyOiWmBJeCy2q34ajg4NKfBWwPyQA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAKq4IlAL5wsUCCDgByLpIDGdOsqRIMAD0ix02AIZcANvEw58REsFIAKZMNoMYdAFxwA2lTdRGAkIJwBKIVRqd0YWNk4ePn5wyIBdJGkwpCA&options=N4IgzgLgTglgxhEAuaBXApgGhFdYAOA9gHZgwBu6yAZgIYA2Y6AvkA).

### Current fix
Moving the custom variants or values into `unocss.config.ts` solves this problem. This is usually better, however inline customization would be desirable for rapid prototyping. Alternatively, use `classes`.

chu121su12:
Closes #2147

This PR addresses the issue in part by making `VariantFunction` as awaitable, as the attributify variant will need to await 2 generator functions to check if the token is parseable or not.

### DIFF
--- a/packages/autocomplete/src/create.ts
+++ b/packages/autocomplete/src/create.ts
@@ -63,7 +63,7 @@ export function createAutocomplete(uno: UnoGenerator): UnocssAutocomplete {
       return cache.get(input)!
 
     // match and ignore existing variants
-    const [, processed, , variants] = uno.matchVariants(input)
+    const [, processed, , variants] = await uno.matchVariants(input)
     let idx = processed ? input.search(escapeRegExp(processed)) : input.length
     // This input contains variants that modifies the processed part,
     // autocomplete will need to reverse it which is not possible
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -87,7 +87,7 @@ export class UnoGenerator<Theme extends {} = {}> {
       return
     }
 
-    const applied = this.matchVariants(raw, current)
+    const applied = await this.matchVariants(raw, current)
 
     if (!applied || this.isBlocked(applied[1])) {
       this.blocked.add(raw)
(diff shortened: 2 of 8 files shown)


antfu:
I am a bit worried about the performance impact of this change. It seems not a very efficient solution. Any better idea?

antfu:
Thanks! Let's have #2215 + #2216 instead

PR #2170, opened 2023-02-08
feat(preset-mini)!: improve the type of theme, requires TS 4.9

Dunqing:
image

When I want to use `preset-mini`'s colors, then it appears that the `colors` is possibly `undefined`, but the theme colors are definitely present.


Dunqing:
It look like the builder does not support `satisfies` yet

package.json
```diff
@@ -78,6 +78,7 @@
     "bumpp": "^8.2.1",
     "codemirror": "^6.0.1",
     "codemirror-theme-vars": "^0.1.1",
+    "esbuild": "^0.17.7",
```
antfu, line 81: Why do we need this?

Dunqing:
I have now changed to update `unbuild` instead of adding `esbuild`, PR: #2188

### Because
The `satefiles` operator is supported in `esbuild@0.15.13`, but the unbuild use esbuild@0.15.11.

antfu:
It seems to work right now, or?

Dunqing:
Yes, ready to merge

PR #2174, opened 2023-02-09
feat(core)!: make select pseudo elements not to be moved around

chu121su12:
Preventive addition as in https://github.com/tailwindlabs/tailwindcss/issues/9977

The list of pseudo is pseudo elements (PE) that can be styled with pseudo classes (PC). 

By moving the PE at the end, the PE could not be styled via PC. This PR add support for specifying which PE that should not be moved around.

Tagged as breaking due to the default values added, which may shuffle the listed pseudo elements around

Edit: alternative to #2190

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -407,6 +407,11 @@ export interface ConfigBase<Theme extends {} = {}> {
    * @default false
    */
   details?: boolean
+
+  /**
+   * Excluded pseudo elements when moving to the end of selector
+   */
+  unsortedPseudoElements?: string[]
 }
 
 export type AutoCompleteTemplate = string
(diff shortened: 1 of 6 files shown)


packages/core/src/types.ts
```diff
@@ -407,6 +407,11 @@ export interface ConfigBase<Theme extends {} = {}> {
    * @default false
    */
   details?: boolean
+
+  /**
+   * Excluded pseudo elements when moving to the end of selector
+   */
+  unsortedPseudoElements?: string[]
```
sibbng, line 414: In my opinion, we don't need to provide this configuration to end users. I can't think of any scenario where this can be helpful 🤔

chu121su12: The function location in `generator.applyVariants` is tricky to add some kind of constants without config, except maybe importing it directly into `movePseudoElementsEnd` ?

```ts
import { DEFAULT_UNSORTED_PSEUDO_ELEMENTS } from './constants'

export function movePseudoElementsEnd(selector: string) {
  const excludedPseudo = DEFAULT_UNSORTED_PSEUDO_ELEMENTS

```

sibbng: We can inline it right before the function definition for better readability.

chu121su12: see #2190

sibbng:
Closing in favor of https://github.com/unocss/unocss/pull/2190

PR #2187, opened 2023-02-12
refactor(preset-attributify): improve element extractor

### LINKED ISSUE #1584: Attributify with Astro missing some styles in some cases
This is a known problem and has been mentioned in https://github.com/unocss/unocss/issues/105#issuecomment-1236294993. I'm opening a new issue here for further tracking it.

## The case

The attributify extractor fails to generate style for `text-red-500`

```html
<a href={href} text-red-500>Vite</a>
```

**Reproduction:** https://stackblitz.com/edit/withastro-astro-vkmvgi?file=src%2Fpages%2Findex.astro

## Why

As I've [mentioned](https://github.com/AllanChain/astro-uno/#what-it-does):

> Astro pre-processes the .astro files and transforms them into something like:
> 
> ```js
> return $$render``;
> ```
> UnoCSS expected a space after `<img` but Astro gives `$$addAttribute` part instead, making attributify mode not work well. Here we pass modified code with extra space to UnoCSS for extracting the tokens.

### Option 1: pre-process the source before extracting

This is what I've done in [astro-uno](https://github.com/AllanChain/astro-uno), but it's a bit hacky.

```js
        // Fixing attributify preset
        const attributifyPreset = options.presets.find(preset =>
          !Array.isArray(preset) && preset.name === '@unocss/preset-attributify'
        )
        if (!Array.isArray(attributifyPreset) && attributifyPreset) {
          const attributifyExtractor = attributifyPreset.extractors
            .find(extractor => extractor.name === 'attributify')
          const extract = attributifyExtractor.extract
          attributifyExtractor.extract = ({ code, original }) => extract({
            // Adding a space here, otherwise UnoCSS doesn't recognize the tag
            code: code.replace(/(<\w+)(\$\{\$\$addAttribute)/g, '$1 $2'),
            original
          })
        }
```

### Option 2: make Astro allow UnoCSS to be applied earlier

This is not quite possible since the use case is rare. (And the Astro guys chose not to include official UnoCSS integration, didn't they?)

### Option 3: add an option for custom extractors

We can add some APIs for creating custom extractors from existing ones and pass them to attributify. This is similar to option 1 but less hacky. Also, if we decide to support a new framework in the future but the existing extractors are not working well, we can create a new extractor to solve this.

### Other options

...

## Additional comments

This is the last fix `astro-uno` provides but is not included in `unocss/astro`. `astro-uno` can safely retire after `unocss/astro` provides this fix.

**EDIT 1:** Related issue: #1734

### LINKED ISSUE #2112: Preset attributify breaks combined code (```twig) and list syntax
We have encountered just another weird issue with UnoCSS parser.

See [first example](https://stackblitz.com/edit/vite-rubreg?file=docs/vite.config.js) on StackBlitz.

Steps to reproduce:

 - Extend default theme in `./docs/.vitepress/theme/index.ts` with import 'uno.css'
 - Add `presetAttributify()` to `./docs/vite.config.js`
 - create `./docs/example.md`
 - run `npm run docs:dev` or `npm run docs:build`

Issue:
 - in `dev` process, server doesn't send a response
 - in `build` process, the build stops without an exception or any feedback

It should work, as it works in `example-list-only` and `example-twig-only` examples.

sibbng:
Fixes #1584
Fixes #2112
Supersedes #2134

@chu121su12 I tried pasting [your code](https://github.com/unocss/unocss/issues/2067#issuecomment-1396298585) on playground and it was slowing down the editor. Every key store was taking 3-4 seconds on my machine. With this PR it seems instant as it should be. 

I also tested this PR with the provided repro in #2067, and it successfully builds.

### DIFF
--- a/packages/preset-attributify/src/extractor.ts
+++ b/packages/preset-attributify/src/extractor.ts
@@ -8,7 +8,7 @@ const strippedPrefixes = [
 ]
 
 const splitterRE = /[\s'"`;]+/g
-const elementRE = /<\w(?=.*>)[\w:\.$-]*\s((?:['"`\{].*?['"`\}]|.*?)*?)(?:>|\Z)/gs
+const elementRE = /<[^>\s]*\s((?:'.*?'|".*?"|`.*?`|\{.*?\}|[^>]*?)*)/gs
 const valuedAttributeRE = /([?]|(?!\d|-{2}|-\d)[a-zA-Z0-9\u00A0-\uFFFF-_:!%-.]+)=?(?:["]([^"]*)["]|[']([^']*)[']|[{]([^}]*)[}])?/gms
 
 export const defaultIgnoreAttributes = ['placeholder', 'fill', 'opacity']
@@ -22,7 +22,6 @@ export const extractorAttributify = (options?: AttributifyOptions): Extractor =>
     name: 'attributify',
     extract({ code }) {
       const result = Array.from(code.matchAll(elementRE))
-        .filter(match => match[0].endsWith('>'))
         .flatMap(match => Array.from((match[1] || '').matchAll(valuedAttributeRE)))
         .flatMap(([, name, ...contents]) => {
           const content = contents.filter(Boolean).join('')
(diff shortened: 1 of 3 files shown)


chu121su12 (approved):
LGTM. Using the example in https://regex101.com/r/48y4HT/1 the new matcher improves the loop from 403 to 13 steps.

PR #2190, opened 2023-02-13
feat(core)!: make select pseudo elements not to be moved around

chu121su12:
Alternative to #2174

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -647,11 +647,27 @@ function applyScope(css: string, scope?: string) {
     return scope ? `${scope} ${css}` : css
 }
 
+const excludedPseudo = [
+  '::-webkit-resizer',
+  '::-webkit-scrollbar',
+  '::-webkit-scrollbar-button',
+  '::-webkit-scrollbar-corner',
+  '::-webkit-scrollbar-thumb',
+  '::-webkit-scrollbar-track',
+  '::-webkit-scrollbar-track-piece',
+  '::file-selector-button',
+]
 export function movePseudoElementsEnd(selector: string) {
   const pseudoElements = selector.match(/::[\w-]+(\([\w-]+\))?/g)
   if (pseudoElements) {
-    for (const e of pseudoElements)
+    for (let i = pseudoElements.length - 1; i >= 0; --i) {
+      const e = pseudoElements[i]
+      if (excludedPseudo.includes(e)) {
+        pseudoElements.splice(i, 1)
+        continue
+      }
       selector = selector.replace(e, '')
+    }
     selector += pseudoElements.join('')
   }
   return selector
--- a/test/__snapshots__/order.test.ts.snap
+++ b/test/__snapshots__/order.test.ts.snap
@@ -10,7 +10,7 @@ exports[`order > fully controlled rules merged and sorted by body 1`] = `
 .css{--var:css;}"
 `;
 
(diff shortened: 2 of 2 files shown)


sibbng (changes requested):
This is good. But it turns out at the moment that no matter whether `movePseudoElementsEnd` is used or not pseudo-elements are always at the end of the selector. This is a regression that happened at some point. Unfortunately, [current test](https://github.com/unocss/unocss/blob/654c7f973878c41cc1c07207f85e439d9fd5869f/test/order.test.ts#L112-L115) is only testing `movePseudoElementsEnd` function. I did that to be sure pseudo-elements properly moved even in extreme cases.

First of all, we have to figure out what causing all pseudo-elements to be moved end of the selector.

Given that input:
```html
<input type="file" class="dark:hover:file:bg-red-600">
<input type="file" class="dark:file:hover:bg-red-600">
```
it should generate these selectors:
```css
.dark .dark\:file\:hover\:bg-red-600:hover::file-selector-button{...}
.dark .dark\:file\:hover\:bg-red-600::file-selector-button:hover{...}
```
but it generated this:
```css
.dark .dark\:file\:hover\:bg-red-600:hover::file-selector-button{...}
.dark .dark\:file\:hover\:bg-red-600:hover::file-selector-button{...}
```
Playgrounds
[Deploy Preview Playground](https://deploy-preview-2190--unocss.netlify.app/play/?html=DwSwdgDgrgLgBDAnhApgXgEQDMQBsUZwDGuAhgM7mYAmpATgNYBcAFgPYBuKdTO%2BTAIwDmAWjopqIgGwAGGRgB8AKFCRYCZOmx4CxMpRr1mfFK07dBo8ZNnzlQA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6sfFJ1bX1enkwVjCJTqRwfZYD0Pa4pBhwNQBuGFDBowN99k5mqGqyIgqN4ZFyokoqPgCUssLyjCxsnDx8-OeXRwqS0n6HqHQS6xihAEYAHQAJheqAkaFweSqrDAjQA9Aj6CAgdUEZsWmoLuUkNikEA&options=N4XyA) - ❌
[Tailwind Playground](https://play.tailwindcss.com/4xYMI8nNYg) - ✔

I tested the state of the repository before #987. It applies `:hover` and `::file-selector-button` in a different order as expected, but in the wrong order which later is fixed by #987. Generating pseudo-elements at the end of the selector is not normal even if we remove `movePseudoElementsEnd`. We have to address what caused this problem.

sibbng:
@chu121su12 I ran `git bisect` against #987 and it seems regression is caused by #1127. Some of these changes are reverted in #1162 but the source of the problem is coming from what's left. Probably around here:

https://github.com/unocss/unocss/pull/1127/files#diff-72ab94343d47e2266a978d9981390f8bb9a6546dbdf17d0727b6a6687a769ca4R123-R137

sibbng:
OK, I fixed the issue and ditched `movePseudoElementsEnd`. I always wanted to fix this on the generator level instead of later patching with regex. @chu121su12's work on #1127 led us here 🙏

sibbng (approved):

PR #2197, opened 2023-02-14
fix(vite): generate css files for specified rollupOptions in library mode

### LINKED ISSUE #2192: doesn't generate css files because it doesn't respect rollupOptions when vite builds in library mode
### UnoCSS version

0.49.4

### Describe the bug

Unocss doesn't generate css files because it doesn't respect rollupOptions when vite builds in library mode

for a library vite build configuration, eg. in vite.config.ts:

```javascript
export default defineConfig({
	build: {
		lib: {
			entry: resolve(__dirname, "index.ts"),
			name: "theme",
		},
		rollupOptions: {
			output: [
				{
					dir: "dist/es",
					format: "es",
					entryFileNames: "[name].mjs",
					assetFileNames: "assets/theme.css",
					preserveModules: true,
					preserveModulesRoot: resolve(__dirname, "../"),
					sourcemap: false,
				},
				{
					dir: "dist/cjs",
					format: "cjs",
					entryFileNames: "[name].cjs",
					assetFileNames: "assets/theme.css",
					preserveModules: true,
					preserveModulesRoot: resolve(__dirname, "../"),
					sourcemap: false,
				},
				{
					dir: "dist/umd",
					format: "umd",
					entryFileNames: "index.js",
					name: "Theme",
					sourcemap: false,
				},
				{
					dir: "dist/iife",
					format: "iife",
					entryFileNames: "index.js",
					name: "Theme",
					sourcemap: false,
				},
			],
		},
		outDir: resolve(__dirname, "dist"),
		emptyOutDir: true,
		cssCodeSplit: true,
	},
	plugins: [Unocss({ mode: "global", include: [/\.ts$/] })],
});
```

Problem lies in unocss->packages->vite->modes->global->build.ts, here is the problem code:
```javascript
const cssPost = cssPostPlugins.get(options.dir)
if (!cssPost) {
  this.warn('[unocss] failed to find vite:css-post plugin. It might be an internal bug of UnoCSS')
  return null
}
```
Problem is that options.dir is "dir/es", "dir/cjs", etc., as defined in rollupOptions.  Whereas cssPostPlugins is initialized as
```javascript
async configResolved(config) {
        const distDir = resolve(config.root, config.build.outDir)
        cssPostPlugins.set(distDir, config.plugins.find(i => i.name === 'vite:css-post'))
        cssPlugins.set(distDir, config.plugins.find(i => i.name === 'vite:css'))
        await ready
},
```
this `distDir` is absolute path to `outDir`: home/xxx/project/dist.  it will never match with options.dir

As you can see, cssPost will always be null, and code will always return null without proceeding.

Proposed solution is to add entries in cssPostPlugins and cssPlugins for each rolloutOptions `dir` property.  I tested this solution, and it works.

I can make a pr if you want this solution.  Although I don't understand why it's done this way, I mean, vite:css-post plugin can be found, why have to go through a map to look for it.

### Reproduction

will provide if required

### System Info

vite 4.1.1

### Validations

agufagit:
Closes #2192 

I don't know how to write unit test for this change, but I did test it manually.

### DIFF
--- a/examples/vue-cli4/package.json
+++ b/examples/vue-cli4/package.json
@@ -1,8 +1,9 @@
 {
   "private": true,
   "scripts": {
-    "dev": "vue-cli-service serve",
-    "build": "vue-cli-service build"
+    "dev": "cross-env NODE_OPTIONS=--openssl-legacy-provider vue-cli-service serve",
+    "build": "cross-env NODE_OPTIONS=--openssl-legacy-provider vue-cli-service build",
+    "build:macos": "vue-cli-service build"
   },
   "dependencies": {
     "vue": "^2.7.13"
@@ -11,6 +12,7 @@
     "@unocss/webpack": "link:../../packages/webpack",
     "@vue/cli-plugin-babel": "~4.5.15",
     "@vue/cli-service": "~4.5.15",
+    "cross-env": "^7.0.3",
     "unocss": "link:../../packages/unocss",
     "vue-template-compiler": "^2.7.14"
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -174,7 +174,7 @@ importers:
       simple-git-hooks: 2.8.1
       splitpanes: 3.1.5
       terser: 5.16.3
-      tsup: 6.6.3_uujdqti2krmttzhqvubwnsmcci
+      tsup: 6.6.3_typescript@4.9.5
       typescript: 4.9.5
       unbuild: 0.8.11
       unocss: link:packages/unocss
@@ -220,6 +220,7 @@ importers:
       '@unocss/webpack': link:../../packages/webpack
       '@vue/cli-plugin-babel': ~4.5.15
       '@vue/cli-service': ~4.5.15
+      cross-env: ^7.0.3
       unocss: link:../../packages/unocss
       vue: ^2.7.13
       vue-template-compiler: ^2.7.14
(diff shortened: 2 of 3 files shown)


agufagit:
@antfu can you review this?

antfu:
I am not against it, but I am also not very sure what's the case it's solving. Can you write a fixture test for it? Thanks

antfu:
You may refer to: https://github.com/unocss/unocss/blob/c4438351f1e0e6915bad121a5462b74c698a4b40/test/fixtures.test.ts#L49-L83

agufagit:
fixture test is done.

This case might be rare.  I was writing a UI components library https://ui.agufa.tech for Vue and Svelte based on Unocss last July.  There was a treeshaking problem with either Nuxt or Sveltekit (I don't remember which one exactly) with a single .js file, so I had to use rollupOptions per-module to generate js files.

That said, I still want this to work because I'm going to do a new project in Qwik, and I'm transforming this UI library to qwik syntax.  As qwik is relatively new, I want to keep using what was safely working because you never know when a new framework might blow up.

The failed checks are not related to my changes.  It's vue-cli4 test case that's failing, it also failed for other pull requests as I just checked, including the one just committed to main branch.  This error might be related to node versions, it's reproducible with node 18, but works with node 16 on my computer.  I'm not sure why it failed node 14 and 16 here, might be test environment change, might be an internal bug of vue-cli4

PR #2212, opened 2023-02-18
fix: update code of vite-react example to react 18 and fixing button in that example

Avaray:
Button was duplicated (I believe).
React 18 uses [different API](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis).

### DIFF
--- a/examples/vite-react/src/App.tsx
+++ b/examples/vite-react/src/App.tsx
@@ -16,20 +16,6 @@ function App() {
           >
             count is: {count}
           </button>
-
-          <button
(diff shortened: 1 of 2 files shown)


antfu:
The button is not duplicated - it was testing/demonstrating attributify mode

Avaray:
> The button is not duplicated - it was testing/demonstrating attributify mode

I see. 
I'm trying to learn git. Should I close this pull request and open new one with just one commit?

PR #2218, opened 2023-02-19
feat(postcsss): add postcss plugin

sibbng:
Initial PostCSS support.

Supports all directives:
- theme
- apply
- screen

Also adds `@unocss` at-rule

Syntax:
```css
@unocss <layer-name>[, <layer-name>];
```

Example:
Let's say we have 3 layers: 
- preflights
- default
- icons
```css
@unocss preflights; /*preflights will be rendered here*/
@unocss default; /*utilities will be rendered here*/
/* or one line: @unocss preflights, default; */ 

@unocss; /*fallback, always recommended, icons layer will be rendered here*/
```
---

Transformers are not supported, our bundler plugins already take care of this but they are not designed to work without 'uno.css' import. Later we may update them to only do the transformation.

Todos:

- [ ] Docs


zyyv:
Thanks for your hard working to make it better.🥳

Refer to the `directive`, is it better for us to name it `@uno` replace `@unocss` ? Abandon the suffix of `css`.

https://github.com/unocss/unocss/blob/f81cecbc253a6f2a046bc3f46438a0764f24f404/packages/transformer-directives/src/index.ts#L25

sibbng:
> Thanks for your hard working to make it better.🥳
> 
> Refer to the `directive`, is it better for us to name it `@uno` replace `@unocss` ? Abandon the suffix of `css`.
> 
> https://github.com/unocss/unocss/blob/f81cecbc253a6f2a046bc3f46438a0764f24f404/packages/transformer-directives/src/index.ts#L25

I'm strongly against abandoning the suffix. `unocss` is the main name of the project, and more Googleable for people who don't know about `unocss`. 

I'm more tend to do things one way. That's why I don't even suggest aliasing it or adding support for '--at-apply' and other vars. We could add support for those variables later since they are already supported in our bundler plugins. But IMO we should encourage them to use `@apply` over them.

sibbng:
> Thanks for your hard working to make it better.🥳
> 
> Refer to the `directive`, is it better for us to name it `@uno` replace `@unocss` ? Abandon the suffix of `css`.
> 
> https://github.com/unocss/unocss/blob/f81cecbc253a6f2a046bc3f46438a0764f24f404/packages/transformer-directives/src/index.ts#L25

`@unocss` at-rule takes a place as a placeholder for generated CSS, but the `--uno` variable is used as a `apply` directive. So we can't abandon the suffix since they are not the same thing. And having both `@uno` and `@unocss` at rules would be confusing.

packages/postcss/src/index.ts
```diff
@@ -0,0 +1,306 @@
...
+}
+
+interface PluginOptions {
+  content?: string[]
+  directiveMap?: {
+    apply: string
+    screen: string
+    theme: string
+  }
+}
```
antfu, line 200: Can we have a separate type file for type only info, and re-export them. I'd also advise having more explicit name `UnoPostcssPluginOptions` or so.

packages/postcss/src/index.ts
```diff
@@ -0,0 +1,306 @@
...
+          throw new Error('UnoCSS config file not found.')
+
+        if (!uno)
+          uno = createGenerator(cfg.config)
+
+        parseApply(root, uno, directiveMap?.apply || 'apply')
+        parseTheme(root, uno, directiveMap?.theme || 'theme')
+        parseScreen(root, uno, directiveMap?.screen || 'screen')
+
+        const entries = await fg(content ?? ['**/*.{html,js,ts,jsx,tsx,vue,svelte,astro}'], {
```
antfu, line 224: May this include/output list be reused?

antfu (commented):
Thanks for making this! Something I have wanted to do for a long time!

I think we could have it in the next minor release.

It would be great if we could also include some fixture tests to avoid future regression. If you are also planning on some future refactorings, we could mark it as experimental.

antfu:
The `--at-apply` is introduced for vanilla CSS, where `@apply` is not a valid syntax. Since we are in PostCSS now, I think it's even ok not to support variable usage at all.

packages/postcss/src/index.ts
sibbng: We can't. `fast-glob` expect a `micromatch` pattern, but `include/exclude` options are `picomatch` patterns. The key difference is `picomatch` allow regex, while `micromatch` doesn't.

fast-glob:
- https://www.npmjs.com/package/fast-glob#:~:text=This%20package%20uses%20a%20micromatch%20as%20a%20library%20for%20pattern%20matching.

@rollup/pluginutils: 
- https://www.npmjs.com/package/@rollup/pluginutils#:~:text=A%20valid-,picomatch,-pattern%2C%20or%20array

comparison table: 
- https://github.com/micromatch/picomatch#library-comparisons

sibbng: We can't. fast-glob expect a `micromatch` pattern, but include/exclude options are `picomatch` patterns. The key difference is `micromatch` allow "Brace expansion", while `picomatch` doesn't. And for some reason, fast-glob doesn't allow regexes in the first argument despite `micromatch` supporting it. 

fast-glob:
https://www.npmjs.com/package/fast-glob#:~:text=This%20package%20uses%20a%20micromatch%20as%20a%20library%20for%20pattern%20matching.

@rollup/pluginutils:
https://www.npmjs.com/package/@rollup/pluginutils#:~:text=A%20valid-,picomatch,-pattern%2C%20or%20array

comparison table:
https://github.com/micromatch/picomatch#library-comparisons

PR #2233, opened 2023-02-22
feat(core)!: remove VariantHandler.order

chu121su12:
This PR removes the `VariantHandler.order` since it is achievable with callback (as updated in tests). In light of the previous PR #2232, the more useful `VariantObject.order` can be used to properly reorder variant execution.


antfu:
What's the benefit of introducing this breaking change? I think we could keep it as it's actually more flexible than the order on the variant objects, and easier to use than `handle`. We could have order on the variant objects as an feature without breaking this, maybe?

chu121su12:
It certainly is not necessary to remove `VariantHandler.order`. Just to maybe reduce confusion because the other is using same option key, `order` while achieving totally different purpose.

> We could have order on the variant objects as an feature without breaking this

Please close this PR and use #2232 for this.

antfu:
Nice. Thanks

PR #2238, opened 2023-02-23
fix(preset-attributify): filter invalid extraction

### LINKED ISSUE #2037: upgrade from 0.46.5 to 0.47.0 breaks a react project

sibbng:
Fixes #2037


zyyv:
Simple and Violent😅, but it can solve the problem temporarily. I think the best way is to update the reg.

packages/preset-attributify/src/extractor.ts
```diff
@@ -37,7 +37,7 @@ export const extractorAttributify = (options?: AttributifyOptions): Extractor =>
           }
 
           if (!content) {
-            if (isValidSelector(name) && nonValuedAttribute !== false) {
+            if (isValidSelector(name) && nonValuedAttribute !== false && !name.endsWith(':')) {
```
antfu, line 40: Would this be too specific? Should we improve the rules to not handle those cases instead?

sibbng: Done 👍

sibbng:
> Simple and Violent😅, but it can solve the problem temporarily. I think the best way is to update the reg.

I tried, but couldn't come up with one. In my opinion @antfu's suggestion is better. I updated the PR with his suggestion. Now we can still target attributes that end with `:`, or people who need it.

XiNiHa:
This broke my use case ;(
Minimal repro:

PR #2240, opened 2023-02-23
fix(vscode): show suggestions in postcss files

sibbng:


packages/shared-common/src/index.ts
```diff
@@ -98,7 +98,7 @@ export function getMatchedPositions(code: string, matched: string[], hasVariantG
 
   // highlight for plain classes
   let start = 0
-  code.split(/([\s"'`;<>]|:\(|\)"|\)\s)/g).forEach((i) => {
+  code.split(/([\s"'`;<>*]|:\(|\)"|\)\s)/g).forEach((i) => {
```
sibbng, line 101: This is fixing the highlighting of classes if `@unocss-include` is provided in multiline comment syntax:
```css
/* @unocss-include */
@screen at-xs { 
  .class {
    display: flex;
    overflow: hidden;
    @apply flex-col-reverse;
  }
}
```

PR #2247, opened 2023-02-24
fix(preset-attributify): remove s flag from element regex

### LINKED ISSUE #2243: Certain token sequences break matching in attributify mode
### UnoCSS version

Started in `v0.49.5`, present in latest `v0.50.0`.

### Describe the bug

Using the attributify preset, certain very specific token sequences can break the matcher, causing certain utilities to be omitted from the generated CSS. The issue reproduction requires very specific conditions, e.g., multiple template literals, and type definitions in specific places. In the playground (or repo), moving the type definition to the top, changing it to `type Type = object;`, or removing the last line (template literal) all fix the issue. I tried enforcing the prefix by using the following preset settings: `presetAttributify({ prefix: "u-", prefixedOnly: true, strict: true })` without success. Downgrading to `v0.49.4` resolves the issue.

A possible source of this issue may be PR #2186. Although it was later somewhat addressed with PR #2207, there may still be an issue. I may be completely wrong on this (apologies if that is the case).

### Reproduction

See [playground](https://uno.antfu.me/play/?html=GYVwdgxgLglg9mABAQQBQEpEG8BQjEBOAplCAUgDwAmMAbosHHALxYAGAJFgKwC%2BbvRAHoAfAG4cvHFACeAByKIAKvMXNEAIgAeGiaEiwEiAEIZseQiTKUa9AEYBzZhuJUNw8ZJwQEAZyiIWojqnDz8EkA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owCCMMUwARgK4zDoCeZF1tAKpYIcAL5x0UCCDgByNiIDGVKnIBQ61AA9IsFBgCGbADbw0mHPkLAiACgTq45SjRhUAXHADaT5y4EYYQg7AEoSP2d%2BNwYmVg4ubjCI5wBdCLFQ9SA&options=N4IgLgTghgdgzgMwPYQLYgFwKgGzgUwBoQJ84AHJeASwDd9NIBXIkAd2oBMwALTAdgAcATmI981AOY8wAgEwA2AL5A) reproduction (yes, technically there is JavaScript in the HTML window, simulating the usage of .tsx files; should not matter), or my [repo](https://github.com/jozefmery/unocss-issue-repr).

### System Info

_No response_

### Validations

sibbng:
Fixes #2243

### DIFF
--- a/test/preset-attributify.test.ts
+++ b/test/preset-attributify.test.ts
@@ -128,6 +128,11 @@ describe('attributify', () => {
       ]}
     >`
 
+  const fixture6 = `<div foo={\`\${5}\`} foo={"\${5}"} foo={'\${5}'} >
+  \`'"
+  <div bg="red" >
+  `
+
   const uno = createGenerator({
     presets: [
       presetAttributify({ strict: true }),
@@ -208,6 +213,11 @@ describe('attributify', () => {
       ],
     })
     const { css } = await uno.generate(fixture5, { preflights: false })
+    expect(css).toMatchSnapshot()
+  })
+
+  test('fixture6', async () => {
+    const { css } = await uno.generate(fixture6, { preflights: false })
     expect(css).toMatchSnapshot()
   })
 
--- a/packages/preset-attributify/src/extractor.ts
+++ b/packages/preset-attributify/src/extractor.ts
@@ -8,7 +8,7 @@ const strippedPrefixes = [
 ]
 
 const splitterRE = /[\s'"`;]+/g
-const elementRE = /<[^>\s]*\s((?:'.*?'|".*?"|`.*?`|\{.*?\}|[^>]*?)*)/gs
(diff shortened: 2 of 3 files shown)


test/preset-attributify.test.ts
```diff
@@ -128,6 +128,11 @@ describe('attributify', () => {
       ]}
     >`
 
+  const fixture6 = `<div foo={\`\${5}\`} foo={"\${5}"} foo={'\${5}'} >
+  \`'"
+  <div bg="red" >
+  `
```
chu121su12, line 134: There's another backtick at the end as part of user string. Without the backtick, the bg-red token is properly generated. Maybe we could add more test with both test samples.

sibbng: I tested it on the playground, it works as expected. The first line of fixture already covers those situations.

antfu (approved):

PR #2255, opened 2023-02-24
fix(postcss): fix export from `unocss`

sibbng:
`unocss/postcss` was only exporting an object

unocss/postcss: `{ postcss: true } `
@unocss/postcss `[Function: unocss] { postcss: true }`


packages/unocss/src/postcss.ts
```diff
@@ -1 +1,2 @@
-export * from '@unocss/postcss'
+import postcss from '@unocss/postcss'
+export default postcss
```
sibbng, line 2: Main change is here, other changes didn't even have an effect. I just added them for consistency.

PR #2262, opened 2023-02-25
chore: update pnpm version with deps

sibbng:


scripts/pnpm.ts
```diff
@@ -0,0 +1,5 @@
+import { readFileSync, writeFileSync } from 'fs'
+
+const pkg = JSON.parse(readFileSync('package.json', 'utf-8'))
+pkg.packageManager = `pnpm@${pkg.devDependencies.pnpm}`
+writeFileSync('package.json', `${JSON.stringify(pkg, null, 2)}\n`, 'utf-8')
```
zyyv, line 5: Use `readJson` and `writeJson` from `fs-extra` is more easily, right?
https://github.com/jprichardson/node-fs-extra/blob/master/docs/writeJson.md

antfu:
Thanks. `taze` now supports bumping `packageManager`

PR #2267, opened 2023-02-26
refactor(postcss): perf and stability improvements

sibbng:
- fixes an issue that sometimes caused new styles to not appear
- prevents running on unrelated CSS files
- allows customizing `unocss` at-rule and plugin name. this is useful if you want to have separate CSS files and uno configs for different routes, for example, marketing site and dashboard.

### DIFF
--- a/packages/postcss/src/types.ts
+++ b/packages/postcss/src/types.ts
@@ -9,6 +9,7 @@ export interface UnoPostcssPluginOptions {
     apply: string
     screen: string
     theme: string
+    unocss: string
   }
   cwd?: string
   configOrPath?: string | UserConfig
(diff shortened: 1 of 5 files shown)


packages/postcss/src/types.ts
```diff
@@ -9,6 +9,7 @@ export interface UnoPostcssPluginOptions {
     apply: string
     screen: string
     theme: string
+    unocss: string
```
antfu, line 12: What is unocss directive?

sibbng: The placeholder: `@unocss` ,  `@unocss preflights` 

https://github.com/unocss/unocss/tree/main/packages/postcss#unocss

sibbng: > allows customizing unocss at-rule and plugin name. this is useful if you want to have separate CSS files and uno configs for different routes, for example, marketing site and dashboard.

For example:

```css
/* marketing.css */
@marketing;
````

```css
/* dashboard.css */
@dashboard;
````

I may remove this feature later, I'm not super sure about this. That's why I kept it undocumented. It still lacks VS Code support. I will later explore how we could support multiple configs in VS Code.

antfu (approved):

PR #1921, opened 2022-11-26
refactor(preset-attributify): improve ternary extraction

sibbng:
Fixes #1915

It seems we already support this in Vue templates. This PR adds support for ternary operators in JSX and eliminates some false positives of previous ternary extraction logic.

Issue #1915, opened 2022-11-25
about dynamic attributify mode


Handpear:
Hello guys,

I have a class need to use attributify mode,

`pos="!fixed"` it's working

`pos={isSideFixed ? '!fixed' : 'relative'}` it's not working

How do I get the corresponding css generated in advance? using the `safelist` ?

antfu:
We temporarily close this due to the lack of enough information.
Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) to reopen the issue.
Thanks.

> [Why reproduction is required](https://antfu.me/posts/why-reproductions-are-required)

Handpear:
![image](https://user-images.githubusercontent.com/40299372/203942123-26525723-5a76-4d14-8964-00c4c522fd4f.png)

`fixed` and `relative` has extracted but not working like other attributify

[minimal reproduction](https://github.com/Handpear/unocss-test)

Handpear:
It doesn't seem to be antd5 problem either

Handpear:
@zyyv

zyyv:
This does not affect anything, because unocss is a static analysis, and extracts the token to generate css.

Handpear:
I don't understand how this could be invalid, it's working in windicss..

sibbng:
@Handpear I made a PR to fix this issue: #1921

In the meantime, you could you this roughly written extractor:

[...]

PR #2271, opened 2023-02-27
feat(preset-attributify): improve ternary support

sibbng:


antfu:
I am not so sure about this - ideally, UnoCSS should be language/framework agnostic. In attrubutify mode we do specialize the HTML support. But having ternary support could be a framework specific thing. I think does not scale in the long term.

sibbng:
> In attributify mode we do specialize the HTML support.

Every framework out there that has HTML-like interface also supports ternaries and their users use them. Since most of our users use UnoCSS through a framework, we should also specialize in that part too.

- [elk/components/status/StatusPreviewCardNormal.vue](https://github.com/elk-zone/elk/blob/f3ad179c69700c504505a9a0fb366dc6e0b05eaf/components/status/StatusPreviewCardNormal.vue#L106)

- https://github.com/unocss/unocss/issues/1915

- https://github.com/unocss/unocss/issues/1921
   > It seems we already support this in Vue templates.

---

> UnoCSS should be language/framework agnostic

We couldn't be language agnostic without knowing what language we're dealing with. In its current form, `preset-attributify` seems to assume it's in an HTML zone. So, whenever this piece of code doesn't make sense for a language, we could disable it easily.

Initial ternary support [was aiming to reduce false positives](https://github.com/unocss/unocss/pull/1921#issuecomment-1328104261). You can see false positives in [this deploy preview](https://deploy-preview-1917--unocss.netlify.app/play/?html=DwZQLghmCuDOAKAnApgNwJbIO4GEKIBMBJAOwDMB7AAgC4AHAXgCIADATwBYqASAb3VggAjtHzIqAMglV2HFlQD8VAOQAPALQdltFVoC%2BLJlQD0APgBQ54AXSpzVWmGSqwzVQBtFKgKwftNFQAjCFhkZSZzMyA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6sfFJ1bX1enkwVjCJTqRwfZYD0Pa4pBhwNQBuGFDBowN99k5mqGqyIgqN4ZFyokoqPgCUssLyjCxsnDx8-OeXRwqS0n6HqHQS6xihAEYAHQAJheqAkaFweSqrDAjQA9Aj6CAgdUEZsWmoLuUkNikEA&options=N4XyA) of an older PR.

Initial ternary support wasn't perfect. When ternaries are nested inside a string literal, it ignores everything the string literal has, and only extracts the ternary's left and right sides. You can see that in main branch right now:
- [Playground (main)](https://uno.antfu.me/play/?html=DwZQLghmCuDOAKAnApgNwJbIO4GEKIBMBJAOwDMB7AAgC4AHAXgCIADATwBYqASAb3VggAjtHzIqAMglV2HFlQD8VAOQAPALQdltFVoC%2BLJlQD0APiA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcEWWwCGuOlmggMUAGrr2mmAHEoELmEoBfOFksg4Aci4qJdOi6ShIsRDgpABtoOjh7RwhnFwABdwhPOgB6EQUAWhA8YB8kDAAPf3hMLHUuYOLsPEJiMgAKZFQocvoALjgAbQ6XCS46GGj05uCMF3JAkOh2l1o0FwiAXQXZOgALaBhemDp2xtRXAB5e-sG1ja2XaZgCmHTr9WCxqls1VYw9XapUSagdxC-9jRQMZ%2BO0fnQAHQcYJcDCyfZwBhSXBoEFgiChX4QujBdTXeERWQvKhpRh-DoA0lKFR1ACUBKpLDYnB4fH4dIZ8nEyLoDQBqDoEgeGHaAEYIQAmAnfNC4aarVhgHbJZL0EDY1bJJ4I2z0qjLKgaLQ6KB6X7tDpG7S6fRGEy4cyWax0g26pBAA&css=HQMwNgpgHg3gJgSwM4AcwEMCeAuc0DcAvkA&options=N4IgTgpgzgDg9gOygSwG4RALgGYEMA2UEANCAC5i5LZxgC2WehEAvkA)

Fixing that was not as easy as today back then. It's required proper extraction and syntax highligting. In recent weeks, I adressed a bunch of edge cases in extraction and highlighting part. And we're here today:

- [This PR's Deploy Preview playground](https://deploy-preview-2271--unocss.netlify.app/play/?html=DwZQLghmCuDOAKAnApgNwJbIO4GEKIBMBJAOwDMB7AAgC4AHAXgCIADATwBYqASAb3VggAjtHzIqAMglV2HFlQD8VAOQAPALQdltFVoC%2BVVQA4WTKgHoAfAChrwAulTWqtMMlVhmqgDaKVAVh9tGhUAIwhYZGUmaysgA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6sfFJ1bX1enkwVjCJTqRwfZYD0Pa4pBhwNQBuGFDBowN99k5mqGqyIgqN4ZFyokoqPgCUssLyjCxsnDx8-OeXRwqS0n6HqHQS6xihAEYAHQAJheqAkaFweSqrDAjQA9Aj6CAgdUEZsWmoLuUkNikEA&css=Q&options=N4XyA)

Removing this feature was the easiest thing to do for me. I wanted to give it a try to see if this can work. While working on it I asked myself a few times; _"Why do we even support that?"_.

> I think does not scale in the long term.

I definitely agree with that. But this PR significantly improves an existing feature and it may not need to be scaled. It could stop there. Let's give it a try and see how far it goes 🤞

antfu (commented):
What I am worrying about is that - I do not understand the new regex and logic you introduced. I definitely trust you, but the problem is the complexity here. To demonstrate my concern, I have pushed a few negative test cases. I imagine our implementation will be more complex and more specific if we want to fix them all.

The fact is that JavaScript is so dynamic, and it's not possible to cover all the cases using RegExp.

I see "false-positive" as unavoidable as we can't possibly know the precise intention of users' code, or properly parse every language. The cost of "false-positive" is to ship a few extra unused CSS (that we try to minimize as possible). While on the other hand, "false-negtive" will be much more harmful as it might break users' app or their expectations.

I feel that I made a hasty decision to have the initial ternary support - I actually considering we'd better remove the ternary handle as a whole to use the simple split way.

PR #2287, opened 2023-03-01
fix(vite): svelte-scoped: search attributify candidates only on template

### LINKED ISSUE #2284: SvelteKit application doesn't build after updating uno from 0.50.1 to 0.50.2
### UnoCSS version

0.50.2

### Describe the bug

With UnoCSS 0.50.2, I'm getting the following error trying to build my SvelteKit application:
image

This does not happen with Uno 0.50.1 and as soon as I downgraded to 0.50.1 again, the error disappears.

### Reproduction

0.50.2 seems to cause incompatibility with [this](https://github.com/rgossiaux/svelte-headlessui) library. Don't have time to create a repro right now and was hoping you'd know what's going on since this was just a patch version change - but if needed, I can try making a reproduction later on.

### System Info

_No response_

### Validations

sibbng:
Fixes #2284

Is transforming node_modules in svelte-scoped mode something we expected?

### DIFF
--- a/packages/vite/src/modes/svelte-scoped/transform.ts
+++ b/packages/vite/src/modes/svelte-scoped/transform.ts
@@ -145,7 +145,10 @@ export async function transformSvelteSFC(code: string, id: string, uno: UnoGener
     s.overwrite(start, start + match[1].length, `${className}={${token}}`)
   }
 
-  const { matched } = await uno.generate(code, { preflights: false, safelist: false, minify: true })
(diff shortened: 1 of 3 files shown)


antfu:
> Is transforming node_modules in svelte-scoped mode something we expected?

Yes I think so. It will be able to support unbundled components libs. Users can always manually exclude if they don't want it, but the other way around would make things harder to config.

PR #2293, opened 2023-03-02
feat(postcss): allow customizing single directive name

sibbng:


kirklin (changes requested):
```
  const {
    cwd = process.cwd(),
    directiveMap: userDirectiveMap,
    content,
    configOrPath,
  } = options

  const directiveMap = Object.assign({}, {
    apply: 'apply',
    theme: 'theme',
    screen: 'screen',
    unocss: 'unocss',
  }, userDirectiveMap)
```

PR #2300, opened 2023-03-04
chore(preset-mini): use `h` replace `handler` name

zyyv:


antfu:
What's the benefit of this? To me, an export should always be explicit (h might be confusing with render function etc.), while in the local usage, you can have some shorthand (thus, the use of `as h`).

PR #2308, opened 2023-03-05
chore: use node: protocol when importing built-in modules

code-ManL:
`
import fs from 'fs'
`

->

`
import fs from 'node:fs'
`


antfu:
Thanks. I have update the ESLint rule to force that instead.

code-ManL:
ok

PR #2337, opened 2023-03-08
feat(transformer-attributify-jsx-babel): new transformer

sibbng:
Adds experimental babel transformer strategy to `transformer-attributify-jsx`.

Usage:
```
npm i @unocss/transformer-attributify-jsx-babel -D
```

```diff
- import transformerAttributifyJsx from '@unocss/transformer-attributify-jsx'
+ import transformerAttributifyJsx from '@unocss/transformer-attributify-jsx-babel'
```


packages/transformer-attributify-jsx/package.json
```diff
@@ -40,6 +45,9 @@
     "@unocss/core": "workspace:*"
   },
   "devDependencies": {
+    "@babel/core": "^7.21.0",
+    "@babel/plugin-syntax-jsx": "^7.18.6",
+    "@babel/preset-typescript": "^7.21.0",
```
antfu, line 50: Maybe we should not bundle them?

And I just checked, and it seems that `@unocss/transformer-attributify-jsx` is included in `unocss`, which I don't really willing as deps on babel. If you think it's worth to have the babel transformations I guess maybe we should stop including them by default, or make another babel specific preset.

sibbng: I moved babel transformer to a new package and not included it on unocss. I updated readmes and configurations as well. It should be ready to release if I didn't miss anything.

PR #2344, opened 2023-03-09

chu121su12:
Export all built-in presets as individual preset key/function pair. The main purpose of this is to be able to use `presetIcons` online without needing to use bundler. With separate presets, each of them can be used with their respective options.

As for the existing cdns, we can consider deprecating all but `core` and `full`. `core` is the recommended one to be used for integration, while `full` is for showcase shortcut.

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -93,6 +93,7 @@ declare global {
   interface Window {
     __unocss?: UserConfig & { runtime?: RuntimeOptions }
     __unocss_runtime?: RuntimeContext
+    __unocss_presets?: Record<string, Function>
   }
 }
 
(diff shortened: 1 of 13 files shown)


packages/runtime/src/index.ts
```diff
@@ -93,6 +93,7 @@ declare global {
   interface Window {
     __unocss?: UserConfig & { runtime?: RuntimeOptions }
     __unocss_runtime?: RuntimeContext
+    __unocss_presets?: Record<string, Function>
```
antfu, line 96: Do you think it might be better to have it as `__unocss_runtime.presets`?

antfu:
Closing this due to inactive. Feel free to bring it back if you still think it's necessary. Thanks.

PR #2345, opened 2023-03-10

chu121su12:
Continuing #2344, this PR refactors preset-icons & preset-web-fonts so that most of the code can be used in runtime. However, this requires direct file import (`require '../../..'`), hence this is on separate PR.


antfu:
Closing this due to inactive. Feel free to bring it back if you still think it's necessary. Thanks.

PR #2364, opened 2023-03-17
feat(extractor-arbitrary-variants)!: make arbitrary extractor standalone

antfu:
I feel the current arbitrary variant support is a bit polluting the core too much. Ideally, the core should be agnostic. Also, I think arbitrary variants might not be needed for everyone, so making it a separate PR so ppl could opt-in when needed, and ppl don't use it do not need to pay the performance cost.

/cc @chu121su12 @sibbng let me know what you think, thanks!


test/pos.test.ts
```diff
@@ -36,16 +36,6 @@ describe('matched-positions', async () => {
...
-          [
-            23,
-            37,
-            "[border=\\"[&_span]:white\\"]",
-          ],
-          [
-            46,
-            65,
-            "[hover=\\"[&>span]:text-white\\"]",
-          ],
```
sibbng, line 48: Please include presetArbitrary on this test too. It tests highlighting arbitrary variants in attributify context.

test/pos.test.ts
```diff
@@ -143,11 +133,6 @@ describe('matched-positions', async () => {
             43,
             "hover:scale-100",
           ],
-          [
-            44,
-            63,
-            "[&>span]:text-white",
-          ],
```
sibbng, line 150: Same as above. This tests some edge cases related to `<` and `>` characters.

test/assets/preset-wind-targets.ts
```diff
@@ -394,7 +394,6 @@ export const presetWindTargets: string[] = [
 
   // variants experimental
   '@hover-text-red',
-  '@hover:[[open]_&]:text-blue',
```
sibbng, line 397: We should add a separate test for this in `preset-wind.test.ts`.

packages/preset-arbitrary/src/split-arbitrary.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  for (const match of code.matchAll(quotedArbitraryValuesRE))
+    result.add(match[0])
+
+  code
+    .split(defaultSplitRE)
+    .forEach((match) => {
+      if (isValidSelector(match) && !arbitraryPropertyCandidateRE.test(match))
+        result.add(match)
+    })
```
sibbng, line 26: This probably will not work as it was working previously in `split.ts`. The purpose of these lines is to reduce some false positives captured by `defaultSplitRE`. We remove those false positives on line 12.

If I remember correctly, this is capturing false positives like that:

```
hover="[&>span]:text-white"
```

Which is excluded from the tests currently. 

`split.ts` will capture `[&>span]:text-white` and presetAttributify will capture `hover="[&>span]:text-white"` and unocss will generate unused `[&>span]:text-white`.

If this is a thing, we can't merge this PR without resolving this regression.

sibbng (changes requested):

packages/preset-arbitrary/src/split-arbitrary.ts
sibbng: Maybe we should provide `code` variable as a MagicString instance and remove the matched utilities from the code. That means this extractor should run before the default splitter (split.ts).

sibbng: Another solution is would be removing the default splitter from the extractors list since this extractor already does its job. This is not an ideal solution and we don't have access to uno config on here as of now.

sibbng (commented):
Can we have some benchmarking tests to see performance cost before/after?

test/assets/preset-wind-targets.ts
antfu: It has been moved to
https://github.com/unocss/unocss/blob/350fd1a38b0ae8c2be3de3f23a63db31e92235df/test/assets/preset-arbitrary-target.ts#L16-L17

antfu:
> Can we have some benchmarking tests to see performance cost before/after?

While actual performance is not my concern here. What I am concerned about is the "agnostic" part. The core should not be aware of how the utils are defined in presets, and split.ts should provide a minimal and transparent implementation for the possibility of other presets to use.

antfu:
Ran `nr bench` against this branch and the main, the delta cost of unocss is `140.01 ms` vs `150.79 ms`, I think it's not that significant but quite noticeable. I imagine the different could get higher when we have more files in the project. But again, I consider performance not a concern here.

sibbng:
> While actual performance is not my concern here. What I am concerned about is the "agnostic" part. The core should not be aware of how the utils are defined in presets, and split.ts should provide a minimal and transparent implementation for the possibility of other presets to use.

Recently added configResolved almost no-op when not using `presetAttributify({ strict: false })`. Because we don't include extractorSplit when an extractor is already defined.

https://github.com/unocss/unocss/blob/8554bf483da79bd51383a6f87f18160f90386a7d/packages/core/src/config.ts#L59-L62

And this already meets the "agnostic" part. So, instead of moving arbitrary extractor to a separate package, we should move it to preset-mini. This way, the core extractor can stay agnostic, and our preset can still process the possible candidates that they already target.

> Ran nr bench against this branch and the main, the delta cost of unocss is 140.01 ms vs 150.79 ms, I think (10ms, 7%) is not that significant but already quite noticeable. I imagine the different could get higher when we have more files in the project. But again, I consider performance not a concern here.

My test results: (windi, tailwind disabled, 50x)

main branch:
```
3/18/2023, 5:36:23 PM
1656 utilities | x50 runs (75% build time)

none                             36.26 ms / delta.      0.00 ms 
unocss       v0.50.6            401.40 ms / delta.    365.14 ms (x1.00)
```

feat/preset-arbitrary branch:
```
3/18/2023, 5:29:52 PM
1656 utilities | x50 runs (75% build time)

none                             37.53 ms / delta.      0.00 ms 
unocss       v0.50.6            384.21 ms / delta.    346.68 ms (x1.00)
```

feat/preset-arbitrary is 5.34459% faster.

OK, there are some performance wins here. But I don't think it's worth introducing a breaking change. Since performance is not a main concern here instead of introducing a breaking change, we should make it customizable as part of the presets, and people who don't need it can disable it. And the core extractor will stay agnostic. 

I see no point in moving this extractor to a separate package and making it default disabled by default.

---

Continuing the above discussion here:

> To me, it's actually fine as a little redundant in UnoCSS is invertible, we the rules merge it shouldn't have much impact.

Merging rules will only work if we used those false positives in class attribute. IMO even on a big size project, this is very unlikely to happen. And the real issue is not even that. The real issue is those false positives are highlighted on VS Code. pos.test.ts still have false positives. My guess is that you need to update shared-common/index.ts with preset-arbitrary is used.

> Having a "preprocessor" for extractors is a great idea and I have been thinking it for a while as well. As currently the pug and svelte extractor are coupled with the default extractor.

I was thinking of more unified pre/post processor API. IMO we should deprecate all preprocess/postprocess/configResolved options and have a hook option with a bunch of hooks that can take control on generating process, e.g.
- `pre:generate(hookSpecificContext, ctx)`
  If a hook returns `false` its internal step will be skipped and post hook will be called.
- `post:generate(hookSpecificContext, ctx)`
- `pre:extract(hookSpecificContext, ctx)`
- `post:extract(hookSpecificContext, ctx)`
- `pre:parseToken(hookSpecificContext, ctx)`
- `post:parseToken(hookSpecificContext, ctx)`
  Could be useful to modify declarations, for example: rem to px), We can also extract our caching mechanism to this hook.
- `pre:stringify(hookSpecificContext, ctx)`
  Similar to the current `postprocess` hook, but you have access to whole list of utilities. that means you can add/duplicate some utilities.
- `post:stringify(hookSpecificContext, ctx)`
- `pre:layer(hookSpecificContext, ctx)`
- `post:layer(hookSpecificContext, ctx)`

test/pos.test.ts
```diff
@@ -41,6 +47,11 @@ describe('matched-positions', async () => {
             37,
             "[border=\\"[&_span]:white\\"]",
           ],
+          [
+            46,
+            65,
+            "[&>span]:text-white",
+          ],
```
sibbng, line 54: As I mentioned in my last comment, these shouldn't be there.

test/pos.test.ts
```diff
@@ -15,12 +15,18 @@ describe('matched-positions', async () => {
...
     })
 
     expect(await match(uno, '<div border="b gray4 2 [&_span]:white" hover="[&>span]:text-white" border></div>'))
       .toMatchInlineSnapshot(`
         [
+          [
+            13,
+            14,
+            "b",
+          ],
```
sibbng, line 29: As I mentioned in my last comment, these shouldn't be there.

sibbng (changes requested):

packages/preset-arbitrary/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  },
+  multiPass: true,
+}
+
+export default function arbitraryPreset(options: PresetArbitraryOptions = {}): Preset {
+  return {
+    name: '@unocss/preset-arbitrary',
+    variants: [
+      variantVariables,
+    ],
```
sibbng, line 60: IMO `variantVariables` should stay in its original place in preset-mini and this package should be renamed to @unocss/extractor-arbitrary like #2368. 

variantVariables has preset-mini specific code and has nothing to do with extracting. Otherwise we have to move cssProperty rule here too and having preset-mini dependency in an extractor package doesn't make sense:

https://github.com/unocss/unocss/blob/8554bf483da79bd51383a6f87f18160f90386a7d/packages/preset-mini/src/_rules/variables.ts#L28-L33

sibbng (changes requested):

packages/preset-arbitrary/src/index.ts
sibbng: And by doing that we can include arbitrary extractor in preset-mini exctractor and avoid breaking change.

antfu:
> So, instead of moving arbitrary extractor to a separate package, we should move it to preset-mini

That sounds good to me. Initially, I was thinking of enabling arbitrary variants also for other presets. Directly embedding could also do.

> hooks

That's probably not a bad idea. But I think we won't implement all hooks you mentioned at once. Instead, we could implement them when we see the needs, so we don't over-engineer.

chu121su12:
> So, instead of moving arbitrary extractor to a separate package, we should move it to preset-mini

While at it, I think the variants on preset-mini may benefit when split on separate package

PR #2368, opened 2023-03-18
feat(svelte)!: move out `extractorSvelte` to a separate package

antfu:
Instead of `import { extractorSvelte } from 'unocss'`, now you need to install it via 

```bash
npm i -D @unocss/extractor-svelte
```

```diff
-import { extractorSvelte } from 'unocss'
+import extractorSvelte '@unocss/extractor-svelte'

export default defineConfig({
  extractors: [
-   extractorSvelte
+   extractorSvelte()
  ]
})
```


antfu:
Oh thanks! Updated

Issue #2312, opened 2023-03-06
Support configuration of `languageIds` via `vscode` settings.


ifiokjr:
### Clear and concise description of the problem

Currently, the languages that the `unocss` extension supports are hardcoded. This means that support sometimes pops up in languages, where it isn't wanted, and doesn't show in languages not hardcoded (like rust using wasm).

### Suggested solution

Make `languageId` configurable via **vscode** settings.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://uno.antfu.me).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

skelewir:
I see there is a setting for this in latest plugin, but it does not seem to work.
I've set 
```
    "unocss.languageIds": [       
        "rust"
    ],
```
and still no completion in rust files

antfu:
@skelewir please open a new issue with a minimal reproduction. Thanks

antfu:
Closing as the original issue is solved.

PR #2374, opened 2023-03-20
feat(vscode): support configuration of languageIds via vscode settings

wkeylin:
* #2312


packages/vscode/src/autocomplete.ts
```diff
@@ -21,8 +21,6 @@ const languageIds = [
   'ejs',
   'php',
   'svelte',
-  'typescript',
-  'typescriptreact',
```
antfu, line 25: Why do we remove them?

PR #2377, opened 2023-03-21
fix(astro)!: disable injected reset styles by default

Hongbusi:
When using unocss in Astro without installing @unocss/reset, an error will occur.

Therefore , the default value of injectReset was changed to false because I believe this choice should be left to the user.


zyyv:
`@unocss/reset` in dependencies, it should be installed in `node_modules`
https://github.com/unocss/unocss/blob/8554bf483da79bd51383a6f87f18160f90386a7d/packages/astro/package.json#L43-L47


> When using unocss in Astro without installing @unocss/reset, an error will occur.
I don't find any error in `example/astro`

Hongbusi:
Should be set `shamefully-hoist=true` to be installed in `node_modules`.

Because unocss is a monorepo repository, `@unocss/reset` is introduced in the top-level [package.json](https://github.com/unocss/unocss/blob/8554bf483da79bd51383a6f87f18160f90386a7d/package.json#L70), so you don't report an error in `example/astro`.

PR #2399, opened 2023-03-26
chore(cli): improve readme

sudongyuer:


antfu:
We probably don't need it with the new docs.

Issue #2392, opened 2023-03-26
Documentation Progress 


antfu:
- [x] Initial setup (thanks @lucpotage!)
- [x] Playground and Interactive deploy and redirects
- [ ] Complete team member profiles (@chu121su12, @sibbng, @ydcjeff, @sudongyuer, @QiroNT)
- [ ] Fulfill guides (@antfu)
- [ ] Pages marked with TODO:
- [ ] Remove descriptions of each package and link them to the docs page
- [ ] Ensure we advocate standalone config everywhere, instead of inline config (referencing to `/guide/config-file` page)
- [ ] API document for generator

Branch https://github.com/unocss/unocss/tree/docs-next
Preview https://preview.unocss.dev/

If you want to help, feel free to comment below and take the tasks you want. Thank you!

userquin:
missing links for tools folder?

I'm adding links and link texts to features in #2394

PR #2400, opened 2023-03-26
chore: typos and wordings

bennymi:
This is for the new doc site (#2392 ). I skimmed through all markdown files and updated some typos and wordings.


bennymi:
Also, there is a weird space in this `It's` on the landing page, but I'm not sure what's causing it:

image

Issue #12, opened 2021-10-27
Vendor prefixes


sanscheese:
Nice work on this. It's a great project to dig into!

Is there a plan or suggestion for handling vendor prefixes? Something like Autoprefixer, but that is for postcss.

This is a good run down of the current state of things for prefixing in CSS:
https://css-tricks.com/is-vendor-prefixing-dead/

PR #2408, opened 2023-03-27
feat(preset-mini): add webkit vendor prefix to `user-select`

sapphi-red:
[Safari requires `-webkit-` vendor prefix for `user-select`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility).

This PR was created in past (https://github.com/unocss/unocss/pull/1411) and was declined but given that

- this comment: https://github.com/unocss/unocss/issues/12#:~:text=Providing%20them%20in%20the%20first%20place%20could%20be%20the%20most%20well%20behaved%20and%20performant%20solution%20IMO
- other rules handles vendor prefixes
  - examples
    - https://github.com/unocss/unocss/blob/339f2b2c9be41a5505e7f4509eea1cf00a87a8d1/packages/preset-wind/src/rules/filters.ts#L48-L50
    - https://github.com/unocss/unocss/blob/339f2b2c9be41a5505e7f4509eea1cf00a87a8d1/packages/preset-mini/src/_rules/behaviors.ts#L22-L25

, I think it makes sense to add this.

### DIFF
--- a/packages/preset-mini/src/_rules/static.ts
+++ b/packages/preset-mini/src/_rules/static.ts
@@ -58,10 +58,10 @@ export const resizes: Rule[] = [
 ]
 
 export const userSelects: Rule[] = [
-  ['select-auto', { 'user-select': 'auto' }],
-  ['select-all', { 'user-select': 'all' }],
-  ['select-text', { 'user-select': 'text' }],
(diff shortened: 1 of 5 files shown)


antfu:
Safari would never let me down :)

Issue #2423, opened 2023-04-01
Feature: support linear-gardient percent


Simon-He95:
### Clear and concise description of the problem

<img width="642" alt="image" src="https://user-images.githubusercontent.com/57086651/229305728-3470ee74-55a8-494d-ae65-9b820db4ed4f.png">

### Suggested solution

support from-10%  to-10% via-10%
[Reference TailwindCss](https://tailwindcss.com/docs/gradient-color-stops#ending-color)

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://uno.antfu.me).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #2424, opened 2023-04-01
feat(preset-wind): support linear-gradient percent

Simon-He95:
close: #2423 

image


packages/preset-wind/src/rules/background.ts
```diff
@@ -22,6 +22,11 @@ const bgGradientColorValue = (mode: string, cssColor: CSSColorValue | undefined,
 
 const bgGradientColorResolver = (mode: 'from' | 'to' | 'via') =>
   ([, body]: string[], { theme }: RuleContext<Theme>) => {
+    if (/^[0-9]/.test(body)) {
+      return {
+        [`--un-gradient-${mode}-position`]: h.bracket(`[${body}]`)
```
chu121su12, line 27: You may want to do var reset since using this without stops now may generate unset variable.

Simon-He95: I have a problem the default value will cover the set value. How to resolve it?

image

Simon-He95: I fix it through add another rule

packages/preset-wind/src/rules/background.ts
```diff
@@ -77,7 +86,9 @@ export const backgroundStyles: Rule[] = [
   [/^(?:bg-gradient-)?from-op(?:acity)?-?(.+)$/, ([, opacity]) => ({ '--un-from-opacity': h.bracket.percent(opacity) })],
   [/^(?:bg-gradient-)?via-op(?:acity)?-?(.+)$/, ([, opacity]) => ({ '--un-via-opacity': h.bracket.percent(opacity) })],
   [/^(?:bg-gradient-)?to-op(?:acity)?-?(.+)$/, ([, opacity]) => ({ '--un-to-opacity': h.bracket.percent(opacity) })],
-
+  [/from-([0-9].+)$/, bgGradientPositionResolver('from')],
```
antfu, line 89:
```suggestion
  [/^(from|via|to)-([\d\.]+)%$/, bgGradientPositionResolver('from')],
```

Like this? Should we explicitly ask for it to be end with `%`?

packages/preset-wind/src/rules/background.ts
```diff
@@ -30,29 +30,38 @@ const bgGradientColorResolver = (mode: 'from' | 'to' | 'via') =>
     const { alpha, color, cssColor } = data
 
     if (!color)
-      return
+      return {}
```
antfu, line 33: Why do we return this?

packages/preset-wind/src/rules/background.ts
Simon-He95: Of course, I have adjusted

PR #2426, opened 2023-04-02
fix(preset-mini): split focus pseudo

chu121su12:
See #2416, #2425.

@sibbng I read your comment late, but I think #1973 should be addressed too.

I manage to isolate the mentioned issue to focus pseudo, but this is a temporary fix, since it still handle compound pseudo, say `focus:active:x`, wrong.


sibbng:
We should disable merging behavior for pseudos and remove that sort option for both `active` and `focus`. We already have order control in the array we define those pseudos. Since those pseudos represent a certain state, merging them with a rule up in the tree is a bad idea. I think, in general, merging rules across the whole css is a bad idea. There are some reasons to pursue that, but I already found better solutions for those reasons in #1973. I will look into it again ASAP. 

I was thinking of breaking #1973 into multiple parts. If we can fix the issues you mentioned by disabling merging behavior on pseudos that would be a good start.

PR #2434, opened 2023-04-04
docs: describe preset options

zyyv:


docs/presets/wind.md
```diff
@@ -48,11 +48,6 @@ import { presetWind } from 'unocss'
 ```
 :::
 
-## Rules
```
antfu, line 51: Can we keep this?

antfu (commented):
I guess you need to enable corepack or upgrade your pnpm. The lock is diffing

docs/presets/uno.md
```diff
@@ -60,7 +61,10 @@ For example, both `ml-3` (Tailwind), `ms-2` (Bootstrap), `ma4` (Tachyons), and `
 .mt-10px { margin-top: 10px; }
 ```
 
-## Rules
```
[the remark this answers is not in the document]
zyyv: I'm sorry, I'm based on an old version.😅

zyyv:
> I guess you need to enable corepack or upgrade your pnpm. The lock is diffing

In fact, that is the lock file after I upgraded.😅  
Now, I have downgraded the pnpm version same as `packageManager`.

docs/presets/icons.md
```diff
@@ -298,6 +294,101 @@ presetIcons({
...
+:::
+
+### unit
+
+- Type: `string`
+- Default: `'em'`
+
+Custom icon unit.
+
+### cdn
```
userquin, line 380: I need to review current state for runtime, maybe we need to add some more hints here

userquin: the PR LGTM, can be added later if necessary

userquin: https://github.com/unocss/unocss/pull/1555

PR #2437, opened 2023-04-04
fix(preset-wind): align with `line-clamp`

zyyv:
https://tailwindcss.com/docs/line-clamp


packages/preset-wind/src/rules/line-clamp.ts
```diff
@@ -7,11 +7,12 @@ export const lineClamps: Rule[] = [
...
     '-webkit-line-clamp': v,
-    'line-clamp': v,
   }), { autocomplete: ['line-clamp', 'line-clamp-<num>'] }],
 
   ...['none', ...globalKeywords].map(keyword => [`line-clamp-${keyword}`, {
+    'overflow': 'visible',
+    'display': 'block',
+    '-webkit-box-orient': 'horizontal',
     '-webkit-line-clamp': keyword,
-    'line-clamp': keyword,
```
antfu, line 15: I don't think we should remove this, for future propose.

antfu: I mean, it's ok to align with TW, but I think it's better to keep the non-prefixed one

antfu (approved):

PR #2439, opened 2023-04-04
docs(config): split config file into smaller files

alexzhang1030:
Splitting the `config` file into smaller files, could be convenient for future internationalization efforts. Separating the files in advance can help avoid potential conflicts that may arise if changes are made to the files later.

Will start translating docs to `zh-CN` after this PR is merged.


antfu:
Wait... Thanks for the intention, but I explicitly don't want to do translation for docs. Similar reasons to https://vueuse.org/why-no-translations.html

alexzhang1030:
> Wait... Thanks for the intention, but I explicitly don't want to do translation for docs. Similar reasons to https://vueuse.org/why-no-translations.html

Thanks for your reply!

PR #2449, opened 2023-04-06
docs(fix): navbar links not working on mobile

userquin:
On desktop play and interactive docs links work, but on mobile result in 404, added full url to navbar entries


antfu:
I noticed it too - but it should be a VitePress bug right?

PR #2451, opened 2023-04-06
docs(fix): code blocks on light theme

userquin:
I'm talking with brc-dd to try to fix this on VP... we can remove/update styles added in this PR once we've a way to configure light code theme on VP


docs/.vitepress/theme/style.css
```diff
@@ -115,3 +120,41 @@
   transition: color 0.25s;
 }
 
+.vp-doc th,
```
antfu, line 123: Thanks! But this feels a bit too much. I rather wait for VP to fix it.

userquin: I'll try to simplify it a bit, I just copy styles from pwa docs, we are not using a lot of cases (code blocks inside details)

antfu: Can we keep the complexity in VP? I think we should try to avoid duplications across repos

userquin: Sure

userquin: Anthony, simplified to minimal changes

PR #2456, opened 2023-04-07
docs: link rich integrations to vite

Alfred-Skyblue:
Link rich integrations to vite, otherwise page navigation will result in a 404 error.


userquin (approved):

antfu:
Acutally it's the page I forgot to write, but it's ok we do this way as well for now.

PR #2469, opened 2023-04-08
pref(preset-wind): merge similar rules

Simon-He95:


packages/preset-wind/src/rules/background.ts
```diff
@@ -114,26 +110,25 @@ export const backgroundStyles: Rule[] = [
     }
   }, { autocomplete: ['bg-gradient-shape', `bg-gradient-shape-(${Object.keys(positionMap).join('|')})`, `shape-(${Object.keys(positionMap).join('|')})`] }],
   ['bg-none', { 'background-image': 'none' }],
-
-  ['box-decoration-slice', { 'box-decoration-break': 'slice' }],
-  ['box-decoration-clone', { 'box-decoration-break': 'clone' }],
+  [/box-decoration-(slice|clone)/, ([, body]) => ({
```
antfu, line 113: Please revert these. They are on purpose, as static utils are much more efficient than dynamic ones.

antfu (changes requested):

PR #2470, opened 2023-04-08
docs: fix navbar style

leedom92:
When window innerWidth is not enough but greater than `768px`, there is the problem in navbar.

|  before  |  after  |
|  ------  |  -----  |
| unocss1 |unocss2|


antfu:
Can you send a PR to VitePress as well? Thanks

leedom92:
> Can you send a PR to VitePress as well? Thanks

Thanks for your reminder, I had sent the PR to VitePress.

PR #2477, opened 2023-04-09
Add section to notify devs on how to use astro's add for integrations.

louiss0:
The changes below are a way of telling people how to install unocss as an integration. I'm doing this so that people will know this and do things correctly. The next time you want to do an integration remember this https://docs.astro.build/en/reference/integrations-reference/#allow-installation-with-astro-add.

### DIFF
--- a/docs/integrations/astro.md
+++ b/docs/integrations/astro.md
@@ -9,6 +9,23 @@ The UnoCSS integration for [Astro](https://astro.build/): `@unocss/astro`. Check
 
 ## Installation
 
+
+**Using astro add **
+
+::: code-group
+  ```bash [pnpm]
+  pnpm  astro add -D @unocss/astro
+  ```
+  ```bash [yarn]
+  yarn astro add -D @unocss/astro
+  ```
+  ```bash [npm]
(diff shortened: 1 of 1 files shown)


antfu:
While I am not against it, I just feel it does not provide enough value - you need to read and learn how to use it in order to save two lines of code. I guess later if we support better out-of-box integration we could start to recommend that way.

louiss0:
> While I am not against it, I just feel it does not provide enough value - you need to read and learn how to use it in order to save two lines of code. I guess later if we support better out-of-box integration we could start to recommend that way.

I disagree with this I


I don't think you understood what I was saying. You can use the `astro add` command to add unocss to astro. What an integration does is add the integration to the config.  This pull request is about letting devs know that they can do that for unocss. The last sentence was to let you know that you can allow this to happen for next time you want to add an integration for astro  with a different library.  

**Please do this for developers**

I was trying to install unocss by using astro add but didn't know the name. I assumed it was unocss. I typed `pnpm astro add unocss`. I should've typed `pnpm add @unocss/astro`. There are other things you could do like prompt for a config file.

PR #2478, opened 2023-04-09
Update astro.md

louiss0:
I'm doing this again because I don't think you understand the point of the previous pull request. 
The point of it is to inform developers that they can install unocss by using `astro add`. 
They do this properly by using `astro add @unocss/astro -D`. 
I wrote all of the code that is necessary for you to do just that.

Integrations automatically setup the importing of the file and places it at the right place making life easier for developers. 
If this is not documented time will not be saved people will struggle to type the right name for `astro add` and assume it's just `astro add unocss`.  

The last sentence was to tell you how to do it again if you needed to do another integration.

### DIFF
--- a/docs/integrations/astro.md
+++ b/docs/integrations/astro.md
@@ -9,6 +9,23 @@ The UnoCSS integration for [Astro](https://astro.build/): `@unocss/astro`. Check
 
 ## Installation
 
+
+**Using astro add **
+
+::: code-group
+  ```bash [pnpm]
+  pnpm  astro add -D @unocss/astro
+  ```
+  ```bash [yarn]
+  yarn astro add -D @unocss/astro
+  ```
+  ```bash [npm]
+  npx astro add -D @unocss/astro
+  ```
+:::
+
+**Individually **
+
 ::: code-group
   ```bash [pnpm]
   pnpm add -D unocss
@@ -32,6 +49,7 @@ export default defineConfig({
   ],
 })
 ```
+
 
 Create a `uno.config.ts` file:
 
--- a/docs/integrations/astro.md
+++ b/docs/integrations/astro.md
@@ -10,7 +10,7 @@ The UnoCSS integration for [Astro](https://astro.build/): `@unocss/astro`. Check
 ## Installation
 
 
-**Using astro add **
+**Using astro add**
 
 ::: code-group
   ```bash [pnpm]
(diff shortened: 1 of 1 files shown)


docs/integrations/astro.md
```diff
@@ -24,7 +24,7 @@ The UnoCSS integration for [Astro](https://astro.build/): `@unocss/astro`. Check
   ```
 :::
 
-**Individually **
+**Individually**
```
louiss0, line 27: To tell users how to install the package standalone.

docs/integrations/astro.md
```diff
@@ -10,7 +10,7 @@ The UnoCSS integration for [Astro](https://astro.build/): `@unocss/astro`. Check
 ## Installation
 
 
-**Using astro add **
+**Using astro add**
```
louiss0, line 13: To tell users how to use the astro add command.

antfu:
Thanks for contributing, but again, I don't think it is worth it at the current stage.

`astro add @unocss/astro -D` is only doing halfway, where you still need to create your `uno.config.ts` and do some other stuff. Before we have the official integration like `astro add unocss`, having two ways of installing the integration creates confusion that doesn't justify its complexity. If you want to help with it, you could try to propose to the astro team to have the official integration, then I'd be happy to include it as a recommended way in the docs.

louiss0:
> Thanks for contributing, but again, I don't think it is worth it at the current stage.
> 
> `astro add @unocss/astro -D` is only doing halfway, where you still need to create your `uno.config.ts` and do some other stuff. Before we have the official integration like `astro add unocss`, having two ways of installing the integration creates confusion that doesn't justify its complexity. If you want to help with it, you could try to propose to the astro team to have the official integration, then I'd be happy to include it as a recommended way in the docs.

Integrations allow you to create a config file as well. Is that wjhat you are missing. If you believe there is little gain because of that I can create a pull request for the config file. In fact I was about to do just that. This page should help if you want to automate the creation of the unocss config file. [here](https://github.com/withastro/astro/blob/main/packages/integrations/tailwind/src/index.ts)

**P.S.**
Oh by the way I'm not asking for unocss to be the name I was asking for `@unocss/astro` to be the name.

antfu:
I am looking forward to the integration PR if you are going to create one. Otherwise I prefer to keep things as is. Thanks.

PR #2481, opened 2023-04-10
fix(postcss): ignore dot folders

### LINKED ISSUE #2473: When unocss is used as a postcss plugin, it causes "wait - compiling..." for nextjs.
### UnoCSS version

0.50.8

### Describe the bug

I tried postcss plugin for UnoCSS according to the [documentation](https://unocss.dev/integrations/postcss).

* UnoCSS only works at startup. HMR gets stuck with "wait - compiling...".
* And in globals.css, Tips "Unknown at-rule ".

### Reproduction

https://stackblitz.com/edit/nextjs-dfibap

### System Info

NextJS version: `13.3.0`
Node version: `v18.12.1`
OS: `Windows 11 22H2`

### Validations

sibbng:
Fixes #2473

In this case, the `.next` folder had a few large `js` files causing the slowdown. IMO ignoring all dot folders by default is fine.

### DIFF
--- a/packages/postcss/src/index.ts
+++ b/packages/postcss/src/index.ts
@@ -118,7 +118,7 @@ function unocss(options: UnoPostcssPluginOptions = {}) {
           cwd,
           dot: true,
           absolute: true,
-          ignore: ['**/{.git,node_modules}/**'],
+          ignore: ['**/{.*,node_modules}/**'],
           stats: true,
         }) as unknown as { path: string; mtimeMs: number }[]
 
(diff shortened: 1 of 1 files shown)


packages/postcss/src/index.ts
```diff
@@ -118,7 +118,7 @@ function unocss(options: UnoPostcssPluginOptions = {}) {
           cwd,
           dot: true,
```
antfu, line 119: Should we just update this?

PR #2485, opened 2023-04-10
pref(core): exclude splite to base64

zyyv:


antfu:
I don't think this usage should be advocated. Changing the regex for this very nitch case is a bit overkill. For those I would suggest using custom rule/shortcut, or a custom extractor.

zyyv:
But tailwind supports base64 image type. It happens in the extractor stage, so custom rules have no effect, do you think we should have it originly?

PR #2487, opened 2023-04-11
docs: Update vscode.md to show array syntax

Bobakanoosh:
Didn't realize at first that this could take an array.


zyyv:
> Didn't realize at first that this could take an array.

Its type is `string | string[]`, i think we can introduce its type in describe.

PR #2503, opened 2023-04-14
feat(core)!: add `content` option

sibbng:
TODO: 
- explain the motivation
- add tests

Breaking changes:
- postcss!: `content` option removed
- core: `content` option added (similar to https://tailwindcss.com/docs/content-configuration)
- core!: `include` and `exclude` options moved to content option
- core!: `extraContent` removed, functionality moved to `content` option

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -1,5 +1,5 @@
+import { clone, defaultExclude, defaultInclude, defaultIncludeGlobs, isStaticRule, mergeDeep, normalizeVariant, toArray, uniq } from './utils'
 import type { Preset, ResolvedConfig, Rule, Shortcut, ToArray, UserConfig, UserConfigDefaults, UserShortcuts } from './types'
-import { clone, isStaticRule, mergeDeep, normalizeVariant, toArray, uniq } from './utils'
 import { extractorSplit } from './extractors'
 import { DEFAULT_LAYERS } from './constants'
 
@@ -132,12 +132,27 @@ export function resolveConfig<Theme extends {} = {}>(
   if (!separators.length)
     separators = [':', '-']
 
+  const content = (config.content || {}) as ResolvedConfig['content']
+  content.plain ||= []
+  const defaultPipeline = {
+    include: defaultInclude,
+    exclude: defaultExclude,
+  }
+  content.pipeline ||= defaultPipeline
+  content.pipeline.include ||= defaultPipeline.include
+  content.pipeline.exclude ||= defaultPipeline.include
+
+  // @ts-expect-error private
+  if (config.__postcss)
+    content.filesystem ||= defaultIncludeGlobs
+
   const resolved: ResolvedConfig<any> = {
     mergeSelectors: true,
     warn: true,
     blocklist: [],
     sortLayers: layers => layers,
     ...config,
+    content,
     presets: sortedPresets,
     envMode: config.envMode || 'build',
     shortcutsLayer: config.shortcutsLayer || 'shortcuts',
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -313,6 +313,39 @@ export type Postprocessor = (util: UtilObject) => void
 export type ThemeExtender<T> = (theme: T) => T | void
 
 export interface ConfigBase<Theme extends {} = {}> {
+  content?: {
+    /**
+     * Glob patterns to extract th classes from.
+     *
+     * In Vite plugin, this patterns only used to extract classes
+     * from the files that is not part of the build pipeline.
+     *
+     * In PostCSS plugin, files to be scanned are determined by
+     * the glob patterns defined in this option.
+     */
+    filesystem?: string[]
+
+    /**
+     * Plain text to be extracted
+     */
+    plain?: (string | { content: string; extension: string })[]
+
+    /**
+     * Options used by Vite to filter processing files.
+     */
+    pipeline?: {
+      /**
+       * Patterns that filter the files being extracted.
+       */
+      include?: FilterPattern
+
+      /**
+       * Patterns that filter the files NOT being extracted.
+       */
+      exclude?: FilterPattern
+    }
+  }
+
   /**
    * Rules to generate CSS utilities.
    *
@@ -571,7 +604,7 @@ export interface CliOptions {
 }
 
 export interface UnocssPluginContext<Config extends UserConfig = UserConfig> {
-  ready: Promise<LoadConfigResult<Config>>
+  ready: Promise<LoadConfigResult<ResolvedConfig>>
   uno: UnoGenerator
   /** All tokens scanned */
   tokens: Set<string>
@@ -590,7 +623,7 @@ export interface UnocssPluginContext<Config extends UserConfig = UserConfig> {
   filter: (code: string, id: string) => boolean
   extract: (code: string, id?: string) => Promise<void>
 
-  reloadConfig: () => Promise<LoadConfigResult<Config>>
+  reloadConfig: () => Promise<LoadConfigResult<ResolvedConfig>>
   getConfig: () => Promise<Config>
   onReload: (fn: () => void) => void
 
@@ -598,7 +631,7 @@ export interface UnocssPluginContext<Config extends UserConfig = UserConfig> {
   onInvalidate: (fn: () => void) => void
 
   root: string
-  updateRoot: (root: string) => Promise<LoadConfigResult<Config>>
+  updateRoot: (root: string) => Promise<LoadConfigResult<ResolvedConfig>>
   getConfigFileList: () => string[]
 }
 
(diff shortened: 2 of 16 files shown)


antfu:
I am ok with this refactoring to make options more scoped. But I think we need to make it clear that UnoCSS works differently from Tailwind, that we read "content" from the build pipeline, not from the filesystem manually. So I'd suggest we drop the shorthand of providing an array directly, and name `files` as `filesystem` or so.

```ts
{
  content: {
    filesystem: ['**/foo/**.vue'],
    pipeline: {
       include: [],
       exclude: []
    }
  }
}
```


```ts
{
  content: {
    filesystem: {
      glob: ['**/foo/**.vue'],
      exclude: [] // tho I doubt if this is necessary
    },
    pipeline: false // disable reading from pipeline, relying on fs solely,
    text: [`some content to extra`] 
  }
}
```

sibbng:
> UnoCSS works differently from Tailwind, that we read "content" from the build pipeline, not from the filesystem manually

This is only true for Vite plugin, `extraContent` option and `PostCSS` plugin read files from the filesystem. And I don't think end users have to know that until they really need to.

> But I think we need to make it clear...

We should make it clear only on documentation. I don't think end users will ever need to know that until they have to deal with query parameters that passed along paths. For example:

- `example.vue?macros=true`
- `example.svelte?svelte&type=style&lang.css`

These patterns will not be matched by any glob patterns. Because glob patterns compiled to something like that: `/^.+\.vue$/`. This is the only scenario I can think of when people will ever need to use `include` and `exclude` options.

Also, I would like the mention one thing I did that may not be obvious from the changes. Glob patterns also passed to `include` filter and I removed `string` type from `type FilterPattern` to encourage users to use `files` option instead. This makes clear `include` & `exclude` options are for advanced use cases, and most of the time used by us rather than users.

> So I'd suggest we drop the shorthand of providing an array directly

I'm strongly against that suggestion. Most of the time people will not ever need to configure this option. When they have to, simply providing an array of glob patterns seems pleasing to me as a user. And Tailwind team thinking of making `content` optional on v4. We shouldn't worry too much about that.

My goals with these changes were:
- Unify content options, so Vite plugin and PostCSS can rely on the same options
- Make migration from/to Tailwind easier
- Cleaner content options: Tailwind team nailed that. It is super clear from their documentation how content extraction work. I have seen a lot of issues/discussions from people that confused about how extraction work.

antfu:
> This is only true for Vite plugin

Actually, this is part of the design phoilshpy of UnoCSS, is that we could do that whenever possible. I understand it's not possible for PostCSS plugin, and I think that makes sense. But in that sense, it's actually "This is only **false** for PostCSS plugin".

> Unify content options, so Vite plugin and PostCSS can rely on the same options

So you see I am trying to make the interface generic for both cases. You could say `content.pipeline` is not available in PostCSS, where `content.filesystem` works for both.

> drop the shorthand of providing an array directly > Most of the time people will not ever need to configure this option

If we say most of ppl don't need it, then it actually makes less sense to me to make it "easier". My question is that if we put `content: ['somefile/*.ts']`, what does that mean to a Vite plugin? Is that saying we:

- only extract from that file
- apply it as a `include` filter to the pipeline
- or adding the file on top of the pipeline extracting

Either way, I think they will have different semantic meanings across Vite and PostCSS.

I admit that "extraContent" can be a bit not fair to PostCSS, so I agree with this renaming to make it general. 

> In UnoCSS project I have seen a lot of issues/discussions from people that confused about how extraction work.

Maybe we are biased? People would ask simple questions without looking into the docs anyway, they would keep asking why `bg-${color}` does work, even if it's the same limitation of Tailwind or Windi.

sibbng:
@antfu I updated PR with your suggestions. There is still some work to do. I just want to finalize the API. 

Here is the current type of `content` option:

```ts
export type FilterPattern = ReadonlyArray<string | RegExp> | string | RegExp | null

export interface ConfigBase<Theme extends {} = {}> {
  content?: {
    filesystem?: string[]
    plain?: (string | { content: string; extension: string })[]
    pipeline?: {
      include?: FilterPattern
      exclude?: FilterPattern
    }
  }
}
 ```

File: packages/core/src/config.ts [hunk 1]
```diff
@@ -132,12 +132,27 @@ export function resolveConfig<Theme extends {} = {}>(
...
+    include: defaultInclude,
+    exclude: defaultExclude,
+  }
+  content.pipeline ||= defaultPipeline
+  content.pipeline.include ||= defaultPipeline.include
+  content.pipeline.exclude ||= defaultPipeline.include
+
+  // @ts-expect-error private
+  if (config.__postcss)
+    content.filesystem ||= defaultIncludeGlobs
```
antfu, line 147: Could this be outside of the core? Or maybe the entire `content` config could be outside and up to the integrations to resolve

packages/core/src/utils/defaults.ts
```diff
@@ -0,0 +1,9 @@
+import type { FilterPattern } from '../types'
+import { cssIdRE } from './helpers'
+
+// regex patterns, used with rollup's createFilter
+export const defaultExclude: Exclude<FilterPattern, null> = [cssIdRE]
+export const defaultInclude: Exclude<FilterPattern, null> = [/\.(vue|svelte|[jt]sx|mdx?|astro|elm|php|phtml|html)($|\?)/]
+
+// micromatch patterns, used in postcss plugin and createFilter
+export const defaultIncludeGlobs = ['./**/*.{html,js,ts,jsx,tsx,vue,svelte,astro,elm,php,phtml,mdx,md}']
```
antfu, line 9: I feel they are better to live in shared-integrations

packages/core/src/types.ts
```diff
@@ -313,6 +313,39 @@ export type Postprocessor = (util: UtilObject) => void
...
+     */
+    filesystem?: string[]
+
+    /**
+     * Plain text to be extracted
+     */
+    plain?: (string | { content: string; extension: string })[]
+
+    /**
+     * Options used by Vite to filter processing files.
```
antfu, line 334:
```suggestion
     * Filter to determine whether to extract certain modules from the build tools' transformation pipeline
```

packages/core/src/types.ts
```diff
@@ -313,6 +313,39 @@ export type Postprocessor = (util: UtilObject) => void
...
+  content?: {
+    /**
+     * Glob patterns to extract th classes from.
+     *
+     * In Vite plugin, this patterns only used to extract classes
+     * from the files that is not part of the build pipeline.
+     *
+     * In PostCSS plugin, files to be scanned are determined by
+     * the glob patterns defined in this option.
+     */
```
antfu, line 325:
```suggestion
    /**
     * Glob patterns to extract from the file system.
     */
```

packages/core/src/config.ts
sibbng: Yes, I will move it to the PostCSS plugin.

packages/core/src/utils/defaults.ts
sibbng: I was getting an error during the production build. This is why I moved them there. I will give it a look again.

packages/core/src/types.ts
sibbng: IMO we should clarify its behavior in Vite. Otherwise, people will be confused about why transform plugins are not working on those files.

antfu: What I am thinking it that the `content` itself indicates multiple sources to read from. We could say `pipeline` source is only supported in Vite and Webpack but not in PostCSS, etc.

packages/core/src/types.ts
```diff
@@ -313,6 +313,39 @@ export type Postprocessor = (util: UtilObject) => void
...
+    filesystem?: string[]
+
+    /**
+     * Plain text to be extracted
+     */
+    plain?: (string | { content: string; extension: string })[]
```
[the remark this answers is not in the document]
sibbng: I wouldn't suggest doing that. It would be confusing in many ways. Let's say we have 2 plain content objects with `id` property. What `id` should be? A path, relative or absolute, should it exist in fs or just an extension? And most importantly `id`s should be unique, This will force people to think of a unique name for their `id`s. We somewhat forcing them to overthink. `extension` name clearly shows all we need is an extension.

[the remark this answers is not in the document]
antfu: It's sometimes about extensions. Extractors can check the module id and apply conditionally based on it (under certain folders, or conventions). For example, in Vue we will have

- `foo.vue`
- `foo.vue?vue&type=template&lang=pug`
- `foo.vue?vue&type=style&lang=css`

If I want to apply transformation before passing the extractor, I will need to compile the pug template to HTML, in this case, `extension` can be limited.

Maybe this would not be used often, but being able to pass the full path (it's always named as `id` in other places) would make it extendable in the future and more consistent across the pipeline.

If you are concerned about the understanding, I think we could document to explain it in jsdocs or the docs site.

antfu:
I would love to help moving this forward, let me how I can help on it.

sibbng:
> I would love to help moving this forward, let me how I can help on it.

Todos:
- Resolve content options in the Nuxt package 
- Remove config.__postcss
- Add runtime warnings for deprecated APIs

By extracting the changes in the `packages/core/src/config.ts` to `shared-integration` package we can resolve default values for nuxt and postcss packages. This may also eliminate the required changes related to `ResolvedConfig` type (check`packages/shared-integration/src/context.ts`).

PR #2505, opened 2023-04-15
fix(svelte-scoped): handle expressions in class attribute correctly

### LINKED ISSUE #2504: Expressions in `class` does not get extracted in `svelte-scoped` mode
### UnoCSS version

0.51.4

### Describe the bug

Expressions in `class` attribute such as `class={clsx('contents', loading && 'hidden')}` does not get extracted in `svelte-scoped` mode. They should be converted to corresponding svelte component scoped class, but instead they just inserted as-is. This works in React when used like `className={clsx('text-red-500')}`, so it seems like a bug.

Example code:
```svelte
<div class="text-red-500">This should be red</div>
<div class={clsx("text-red-500")}>This should be red</div>
<div class={clsx(flag && "text-red-500")}>This should be red</div>
```

Result:
image

### Reproduction

https://stackblitz.com/edit/unocss-unocss-nfuydr?file=src/routes/+page.svelte

### System Info

All latested versions

### Validations

devunt:
Fixes #2504.

Corresponding tests are also added.

Now UnoCSS correctly transforms
```svelte
<div class={clsx(flag && "text-red-500")}>This should be red</div>
```
into
```svelte
<div class={clsx(flag && "uno-3as8ej")}>This should be red</div>
```
in the `svelte-scoped` mode.

### DIFF
--- a/packages/vite/src/modes/svelte-scoped/transform.ts
+++ b/packages/vite/src/modes/svelte-scoped/transform.ts
@@ -3,10 +3,11 @@ import { type UnoGenerator, attributifyRE, escapeRegExp, expandVariantGroup } fr
 import { wrapSelectorsWithGlobal } from './wrap-global'
 import { hash } from './hash'
 
-const classesRE = /class=(["'\`])([\S\s]+?)\1/g // class="mb-1"
+const classesRE = /class=(["'\`])([^\{][\S\s]*?)\1/g // class="mb-1"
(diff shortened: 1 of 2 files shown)


devunt:
@antfu since this got merged, can we get a version release? I usally don't rush on open source maintainers (appreciate for your amazing effort on open source projects, I really do admire it 🙇‍♀️) but this bug keeping us from building resuable components in svelte with unocss now and no workaround is available immediately.

Issue #2515, opened 2023-04-18
Bug: skew-


Simon-He95:
### UnoCSS version

0.51.4

### Describe the bug

skew-50 is not supported like scale-50
it should transform to --skew-x:50deg;--skew-y:50deg

### Reproduction

https://unocss.dev/play/

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2516, opened 2023-04-18
pref(preset-mini): imporve `skew` rule

Simon-He95:
close: #2515


test/preset-mini.test.ts
```diff
@@ -110,6 +110,7 @@ describe('preset-mini', () => {
       'text-opacity-50',
       'text-red',
       'scale-100',
+      'skew-50deg',
```
antfu, line 113: Emm. this case shouldn't be put here. Go the the preset-mini-targets instead.

Simon-He95: done!

PR #2525, opened 2023-04-20
fix(preset-wind): remove `--un-gradient-to`

shamokit:
remove `--un-gradient-to` at `case 'via'`


shamokit:
# Check

## class attributes
`bg-gradient-linear from-test-from via-test-via to-test-to`

 ## config add
`  theme: {
    colors: {
      test: {
        from: '#ff0',
        via: '#f0f',
        to: '#0ff'
      }
    }
  }`

sibbng:
This change would fix the issue for your use case but it will the change the behavior of gradients for people who don't use `from` or `to`. I will take care of that. Thanks for the PR 🙏

PR #2539, opened 2023-04-24
Update eslint.md

agenordebriat:
Noticed that `@unocss` didn't work but `@unocss/eslint-config` did.
I followed the installation instructions.


antfu:
No, it should work. See https://github.com/antfu/vitesse/blob/main/.eslintrc

agenordebriat:
Switched .eslintrc.json to .eslintrc and it worked. Switched it back to .eslintrc.json and it still works now. 🥸

PR #2548, opened 2023-04-28
chore: fix typo of wrong file exported in preset-tagify.ts

jacob-8:


zyyv (approved):

jacob-8:
@zyyv does @antfu do all the merging even when another team member reviews a PR or do I go ahead and merge?

zyyv:
I think if the change of this pr is very small and does not involve the need for multiple people to review and make comments, it can be merged at any time. This way we can also save antfu time.

PR #2552, opened 2023-04-28
feat(svelte-scoped)!: introduce`@unocss/svelte-scoped` package

### LINKED ISSUE #2379: Can't import preflights in seperate Svelte component
### UnoCSS version

0.50.6

### Describe the bug

I'm trying to add preflights to a `svelte-scoped` project. Since my `App.svelte` uses dark mode, I tried adding them to a different component. `vite-plugin-svelte` throws a warning (shown below), and the preflights aren't applied.
```
[vite-plugin-svelte] /src/components/StyleWrapper.svelte:2:1 No scopable elements found in template. If you're using global styles in the style tag, you should move it into an external stylesheet file and import it in JS. See https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#where-should-i-put-my-global-styles.
```

### Reproduction
https://stackblitz.com/edit/vitejs-vite-jkyxue

### System Info

Arch Linux/Fedora Kinoite, Firefox 111.0.1, Node 19.8.0

### Validations

### LINKED ISSUE #2420: Element with empty class attribute breaks UnoCSS classes in next element in SvelteKit scoped
### UnoCSS version

0.50.6

### Describe the bug

Element with empty class attribute breaks UnoCSS classes in next element in SvelteKit scoped.

For example if I have a div with an empty class attribute like `<div class=""></div>` the next element gets messed up. Instead of real UnoCSS SvelteKit scoped class names like 
`<p class="_text-4xl_lnitps s-y_bCXRrkrYfP">This is some text</p>` 
it turns into 
`<p class="text-4xl s-y_bCXRrkrYfP">This is some text</p>`
which doesn't work.

### Reproduction

https://stackblitz.com/edit/unocss-unocss-mn6o3t?file=src/routes/+page.svelte

I added `<p class="text-5xl">This is some text</p>` because one time I loaded the stackblitz two elements were bugged instead of just one.

### System Info

_No response_

### Validations

### LINKED ISSUE #2433: Backdrop filters broken in Svelte-scoped from 0.50.2
### UnoCSS version

0.50.2 to latest

### Describe the bug

Backdrop filters broken in Svelte-scoped from 0.50.2. They apparently worked in 0.50.1 but something changed in 0.50.2 that made them stop working. 
This thing is missing: 
opera_03_22-29-19QT7JsNtx65

### Reproduction

Broken: (0.50.2) https://stackblitz.com/edit/unocss-unocss-byrlxk?file=src/routes/+page.svelte
Working: (0.50.1) https://stackblitz.com/edit/unocss-unocss-bijr9g?file=src/routes/+page.svelte

### System Info

_No response_

### Validations

jacob-8:
See discussion starting at https://github.com/unocss/unocss/pull/2530#issuecomment-1518971105 for context. 

Adds `@unocss/svelte-scoped` with two options:
- `@unocss/svelte-scoped/vite`
- `@unocss/svelte-scoped/preprocess`

Status:
- [x] review all `svelte-scoped` related issues as things got a little stale in the intermediary branch away from the project and now a return back
- [x] use vitest.workspace.ts to allow for colocated tests within packages/svelte-scoped/*
- [x] copy in UnocssSveltePreprocess & update tests to use .toMatchFileSnapshot where helpful
- [x] create a modified version of transformerDirectives
  - only pays attention to `apply` directives
  - better suited for use in the Svelte style preprocessor pipeline
  - smartly wraps the Uno added parent, sibling, and child class selectors with `:global()` but not the original class name so that the Svelte compiler won't strip out the class it would otherwise see as unused
- [x] break transformClasses into smaller, more readable, testable parts
  - [x] Ignore #2505 - that is how things are done in React, in Svelte use the framework's `class:mb-1={flag}` syntax.
  - [x] Review #2530 and integrate improvements, closes #2530
  - [x] handle `class=""` bug, fixes #2420 
- [x] add preflights/safelist option to place inside a local component for component libraries, closes #2379
- [x] copy across svelte-scoped-uno Vite plugin (all pieces not related to Svelte component transformation), fixes #2433
- [x] remove the general Vite plugin's `svelte-scoped` mode with an info notice
- [x] update docs & examples
  - [x] point those who'd like to use attributify mode to https://github.com/MellowCo/unplugin-attributify-to-class (update include option to .svelte)
  - [x] add svelte-preprocess example


antfu:
Thanks a lot for linking the issues. Looking forward to it!

jacob-8:
@sibbng this is ready for your review.

@antfu, are the names of packages good? 

Svelte conventions would have me name the preprocessor as `svelte-preprocess-unocss` but with the `@unocss` prefix I feel it's a little silly to write unocss twice, as in `@unocss/svelte-preprocess-unocss`, is why I settled on `@unocss/svelte-preprocess`. Also I don't think it makes sense to say "scoped" in the preprocessor's package name because everything a preprocessor does is just pertaining to one component. There is no global vs scoped context when it comes to Svelte preprocessors. But if you have a better package name, I don't care if you change it.

Also I made the assumption you don't want these packages included in the base `unocss` package, but please change that if I assumed wrongly.

alias.ts
```diff
@@ -29,6 +29,8 @@ export const alias: Record<string, string> = {
   '@unocss/shared-common': r('./packages/shared-common/src/'),
   '@unocss/shared-docs': r('./packages/shared-docs/src/'),
   '@unocss/shared-integration': r('./packages/shared-integration/src/'),
+  '@unocss/svelte-preprocess': r('./packages/svelte-scoped/preprocessor/src/'),
+  '@unocss/svelte-scoped-vite': r('./packages/svelte-scoped/vite/src/'),
```
antfu, line 33: Can we flat these nested packages, or we publish one package with sub modules, like `@unocss/svelte-scoped/preprocess` / `@unocss/svelte-scoped/vite`?

vitest.workspace.ts
```diff
@@ -0,0 +1,4 @@
+export default [
+  'vitest.config.ts',
+  'packages/svelte-scoped/*/vitest.config.ts',
+]
```
antfu, line 4:
```suggestion
import { defineWorkspace } from 'vitest/config'

export default defineWorkspace([
  'vitest.config.ts',
  'packages/svelte-scoped/*/vitest.config.ts',
])
```

packages/svelte-scoped/vite/src/resets/tailwind-reset.css
```diff
@@ -0,0 +1,364 @@
+/* Tailwind Reset */
```
antfu, line 1: Can we reuse `@unocss/reset` and avoid duplication here?

packages/svelte-scoped/preprocessor/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineProject } from 'vitest/config'
+
+export default defineProject({
+  test: {
+    globals: true,
```
antfu, line 5: I would like to avoid have `globals` if possible

packages/svelte-scoped/vite/src/resets/tailwind-reset.css
jacob-8: Sure, though I need help to do that. SvelteKit provides no options to import styles in a specific location in the sense that we would normally place the resets at the top of some `main.ts` like file. So I'm placing them at the start of the global css that will be injected into the head above SvelteKit styles.

In `global.ts` in my vite plugin I've tried all these iterations with no success:

image

They all incur this error:

```bash
error when starting dev server:      
C:\dev\unocss\packages\reset\tailwind.css:6
*,
^

SyntaxError: Unexpected token '*'    
    at Object.compileFunction (node:vm:360:18)
```

packages/svelte-scoped/preprocessor/vitest.config.ts
jacob-8: Sure, I've removed that out.

alias.ts
antfu: https://github.com/unocss/unocss/blob/ab86ae7c7b1441ac1257fac8864a2f4f706f3c68/packages/preset-mini/package.json#L28-L37
https://github.com/unocss/unocss/blob/ab86ae7c7b1441ac1257fac8864a2f4f706f3c68/packages/preset-mini/build.config.ts#L4-L11

Here are the references

packages/svelte-scoped/vite/src/resets/tailwind-reset.css
antfu: Do we need it to be a string? Can we directly `import '@unocss/reset/tailwind.css'`?

packages/svelte-scoped/preprocessor/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      if (checkForApply) {
+        const transformedApplies = await transformApply({
+          code: content,
+          uno,
+          applyVariables: options.applyVariables,
+        })
+
+        if (transformedApplies)
+          return { code: preflightsSafelistCss + transformedApplies }
```
dominikg, line 54: just came across this and wanted to mention that svelte preprocessors can (and should) return a sourcemap for the transforms they did. 

https://svelte.dev/docs#compile-time-svelte-preprocess:~:text=Preprocessor%20functions%20should%20additionally%20return%20a%20map%20object%20alongside%20code%20and%20dependencies%2C%20where%20map%20is%20a%20sourcemap%20representing%20the%20transformation.

It might not be needed today as there are more issues with vite css sourcemaps, but if you have the maps ready, it would be great to return them here.

alias.ts
jacob-8: Ok, done

packages/svelte-scoped/vite/src/resets/tailwind-reset.css
jacob-8: Yes, it needs to be a string that I can inject directly into the head along with preflights/safelist (in dev it's inlined, in build it's a turned into a separate hashed css file) because there is no place in SvelteKit into which we can directly import CSS such that it will come first. If user's manually import into it, say into their root `+layout.svelte` file, the reset could end up being placed into the `head` tag anywhere amongst other imported styles.

So it looks like the only way we can reuse `@unocss/reset/tailwind.css` is if a Vite plugin can take a CSS file and give us a string to use in our function. I don't know how to do that.

The only alternative is to just let users handle resets on their own, separate from Uno. I really, really like the convenience of Svelte Scoped easily providing a reset.

packages/svelte-scoped/preprocessor/src/index.ts
jacob-8: Good catch @dominikg! 

Yes, I was under the impression that CSS sourcemaps gave no benefit, but that makes sense to future proof it. In this context is it important to add the filename as the source?

`map: s.generateMap({ hires: true, source: filename || '' }),` in https://github.com/unocss/unocss/pull/2552/commits/dce1a8085c669c7e9d9673683729660f3d434004

dominikg: i think svelte would add it but to be safe, add it. vite-plugin-svelte currently postprocesses the maps to ensure relative paths https://github.com/sveltejs/vite-plugin-svelte/blob/5e9518b3a5e887efbfada483ec0fcdbbc875a35c/packages/vite-plugin-svelte/src/utils/sourcemaps.ts#L41

you can use vite css.devSourceMap option to test it. eg like https://github.com/sveltejs/vite-plugin-svelte/tree/main/packages/e2e-tests/css-dev-sourcemap

jacob-8: Ok, I'll leave it in.

packages/svelte-scoped/vite/src/resets/tailwind-reset.css
fehnomenal: I also like the simplicity for adding a reset. Couldn't you just read the file from the filesystem to get the contents?

antfu: We have been asking users to add the reset on their own since the beginning. The reasons I am pushing this back are:

- Duplication, harder to maintain/update
- Coupled with Tailwind/preset-wind. UnoCSS is designed to be unopinionated, and I wish it can be welcome for users having different/custom conventions, that are not necessarily bound to Tailwind's reset.
- Asking them to do also give them more awareness of what has been included in their bundle, as resets sometime can conflicts with other stylesheet/frameworks they have.

If there is no strong technical reason I would prefer to have it separate.

jacob-8: - I completely agree on not wanting duplication. It was leftover as a result of copying the plugin in from my personal project and me not taking the time to think of a better solution. So thanks for the pushback :)
- Having Tailwind reset as the only choice was also a result of my copying from personal usage and not improving to fit the broader UnoCSS context. If we are able to settle upon a good way to add resets via this plugin, we would provide all 5 options: `'normalize' | 'sanitize' | 'eric-meyer' | 'tailwind' | 'tailwind-compat'` matching the normal offerings in https://unocss.dev/guide/style-reset
- The way the plugin currently adds the reset option follows the principle seen in https://unocss.dev/guide/style-reset: "UnoCSS does not provide style resetting or preflight **by default** for maximum flexibility and does not populate your global CSS." It's opt-in and users must use the `addReset` option to add it:
  image

So then as I understand, we have two options:

1) Do not provide an optional reset method when using Svelte Scoped, but since the regular `import '@unocss/reset/_______.css'` option is not available in SvelteKit, add a little info to the docs clueing users in to some of their options such as adding this to their `head` tag:
  `<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/normalize.css@8.0.1/normalize.css" />`

2) Read from the filesystem to avoid duplication in some manner like this:

```ts
import { readFileSync } from 'node:fs'
import { dirname, resolve } from 'node:path'
import { fileURLToPath } from 'node:url'
import type { UnocssSvelteScopedViteOptions } from './types'

const _dirname = typeof __dirname !== 'undefined'
  ? __dirname
  : dirname(fileURLToPath(import.meta.url))

export function getReset(addReset: UnocssSvelteScopedViteOptions['addReset']): string | undefined {
  if (addReset === 'eric-meyer')
    return readFileSync(resolve(_dirname, '../node_modules/@unocss/reset/eric-meyer.css'), 'utf-8')
    // Method 1: Include @unocss/reset as a dependency of svelte-scoped. This works in my machine with pnpm but will it always work with npm/yarn?

  if (addReset === 'normalize')
    return readFileSync(resolve(_dirname, '../node_modules/@unocss/reset/normalize.css'), 'utf-8')

  if (addReset === 'sanitize')
    return readFileSync(resolve(_dirname, '../node_modules/@unocss/reset/sanitize/sanitize.css'), 'utf-8')
    // this one is bit more complicated with a base reset and additional resets for specific areas https://unocss.dev/guide/style-reset#sanitize-css. Do we provide the base and point people to https://github.com/csstools/sanitize.css#usage for direct CDN imports if they want more than the base?

  if (addReset === 'tailwind')
    return readFileSync(resolve(_dirname, 'reset/tailwind.css'), 'utf-8')
    // Method 2: this filepath works if we copy resets into svelte-scoped/dist/reset at build/stub time

  if (addReset === 'tailwind-compat')
    return readFileSync(resolve(_dirname, '../node_modules/@unocss/reset/tailwind-compat.css'), 'utf-8')
}
```

That 2nd option would allow users to opt-in using the aforementioned `addReset` option without code-duplication. If you like this one, please comment on my two different filepath methods mentioned in the code comments above (using @unocss/reset as a dependency vs copying resets into dist at build/stub time)
  
Which do you want me to implement, @antfu?

dominikg: What is the goal here in terms of dx? 
While sveltekit does not support the `transformIndexHtml` hook, it does allow developers to import css files.
A common place to do it for global css is in the script block of the root layout, `/src/routes/+layout.svelte` see eg https://sveltekit.new  which is where i'd add the `import "virtual:uno.css"` too.

 Importing a reset before that would work. Or am I missing something here?

jacob-8: I'll answer my own question. Conduitry said this style ordering inconsistency is "expected"  https://github.com/sveltejs/kit/issues/7751#issuecomment-1322630570

> I am imaging to rename it to injectReset and allow custom module resolving
In that case @antfu, I can update the code to handle this, but could you just write a few scratch notes to point me in the right direction? I'm not sure all that would be involved... if it starts with a period then assume the path is relative to the vite.config.ts file? If it doesn't then assume it's in node_modules?

jacob-8: > that's a bug that should be fixed in sveltekit

@dominikg sorry for posting twice. I think we overlapped in our timing. Please see my previous message where I mentioned https://github.com/sveltejs/kit/issues/7751 was filed as a bug report that turned into something which is expected behavior that needs documented. Their are two good repros in there.

For a little more understanding on how and why Svelte Scoped works the way it does please read https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped, especially the [usage](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#usage) section.

> would lead to a lot of duplication

There certainly is a lot of duplication of the simple `mb-1` type styles, but svelte-scoped is very helpful in terms of being able to freely add countless variations of styles and icons without leading to a rapidly ballooning out of control global stylesheet. This is important in the case of large applications that are used for many years. I've worked on long-running projects (15+ years) where the global stylesheet turns into a behemoth. I don't like developing like that, there's too much [pain](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#scoped-utility-classes-unleashes-creativity)). What I want is the ability to just add utility styles wherever needed with the knowledge my global stylesheet is not getting bigger and the styles will seemlessly be handled as needed. 

In many ways, the conceptual model is similar to how each compiled Svelte component is a bit larger than it needs. There's some duplication. But in the end, it's [worth it](https://github.com/halfnelson/svelte-it-will-scale) not to have everything used in the entire app placed in one runtime.js, for which many parts are not needed in many different routes.

dominikg: > I'll answer my own question. Conduitry said this style ordering inconsistency is "expected" [sveltejs/kit#7751 (comment)](https://github.com/sveltejs/kit/issues/7751#issuecomment-1322630570)

That comment was made in regards to importing conflicting global styles from different routes. The root +layout.svelte is used on all routes, so there would be no conflict, the css is always loaded and always before other css from child layouts afaik (import order).

Yes, users could import a different reset on some routes to make this break, but this would be on them, not unocss or sveltekit.

dominikg: This PR is probably not the right place to continue discussing about the usefulness of this mode.

From a technical aspect importing the reset in the root layout should be deterministic/working as long as the user does not import conflicting global css in a child AND does a clientside nav to that route during dev? 

I don't see that as a problem big enough to prevent recommending users to add their style import to the root layout and instead go out of your way to implement a different way to add resets. But ymmv. Happy to see unocss embracing svelte and using it myself.

jacob-8: @antfu custom module resolution for `injectReset` is implemented: https://github.com/unocss/unocss/blob/915a68d43707a5e4489cb0b416d3bd061e0c1640/packages/svelte-scoped/src/vite/types.d.ts#L9-L24 https://github.com/unocss/unocss/blob/2b2096fcc869d1002b29de2f47b7e93e9be87e9c/packages/svelte-scoped/src/vite/getReset.ts#L9-L29.

If that looks good, this review note can be resolved.

@dominikg, please forgive me if I've come across as argumentative. I admire your contributions to Svelte (literally every time I use the Svelte inspector) and am just happy to be in conversation with you and others who make great tools. 

The svelte-scoped mode as it's seen here is the result of me using various iterations of wind+svelte-scoping for almost 2 years. I started with https://github.com/windicss/svelte-windicss-preprocess, then added a preprocessor to run before that preprocessor to overcome some missing features I wanted that are available out of the box with a normal Tailwind setup. I still didn't have feature parity however, and when I came upon UnoCSS, it finally gave a platform where we could build something that had feature parity. 

Due to the various complexities of what we are trying to accomplish, I struggle with how to properly explain how it works. I gave another attempt by adding a bit on [how it works](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#how-it-works) to the docs and hope that gives more clarity. If not, all I can say is that this is the best way to accomplish utility styles in a distributed fashion across Svelte components that I've used. I've also not seen any other solution that supports all the [use cases](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#context-aware). It's the only way I've found to be able to do extensive styling in a large and complex app without feeling any friction. It's a wonderful DX in my opinion, and I hope others will like it too.

dominikg: don't worry, I like to argument a lot, and it seems you have put a lot of thought into it. I still believe there should be better, less duplicating ways to achieve unocss codesplitting (from what I understand is that you want to contain rarely used tokens to separate .css files that are only loaded on some routes. eg. some specific icons or these complex custom things that tend to creep into projects over time.

As long as there is a way to use unocss+svelte without this svelte-scoped mode and unocss sees a value in it, i'm ok with this feature.

For the technical aspects there might be some misconceptions going on either on my part or on yours. The way you are using :global to escape sveltes own scoping is not really recommended.

:global is meant as the occasional escape hatch and it is recommended to only use it with a prefix like `div :global(.something)` so that it is still scoped to the component, but applies to anything inside.

So would there be a way where you use svelte's own scoping instead? This should also avoid the ordering issue you described (apart from users putting imports in the wrong order)

sibbng: > So would there be a way where you use svelte's own scoping instead? This should also avoid the ordering issue you described (apart from users putting imports in the wrong order)

Our PostCSS plugin will be doing exactly that. From my comment at https://github.com/unocss/unocss/pull/2530#issuecomment-1518991854

> I'm working on some sort of scoped feature for our [PostCSS plugin](https://unocss.dev/integrations/postcss#postcss-plugin) that can also be supported by the Vite plugin in the future. I have already created a POC locally. No transformation, no Vite plugin, vite-plugin-svelte does the scoping and hashing for us. It works great. I will make a proposal about it ASAP.

jacob-8: > I have already created a POC locally. No transformation, no Vite plugin, vite-plugin-svelte does the scoping and hashing for us. It works great. I will make a proposal about it ASAP.

Thanks for adding in. I'm looking forward to it. 

In your proposal, please address the ability to [pass classes to children components](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#passing-classes-to-child-components) for which I've found no solution in Svelte's default hashing. Here's the problem:

To allow for this usage you will need to detect when the `class="..."` attribute is part of a Component as opposed to an element. Then you will need to generate the css for those classes separately and wrap them in `:global()` to allow them to have effect when placed into the child component.

That will then lead to bugs in a situation where A passes classes to B that conflict with the classes C passes to D. In this scenario, if C loads later then B's intended behavior will be flipped:

A > B
.hidden .sm:block

C > D
.block .sm:hidden

The only 2 ways around this are to place all styles in a global stylesheet (what everyone does) or to compile the class names into unique names for each component so they don't conflict (what Svelte Scoped does).

So our choices are:
1) Don't compile class names. Use Svelte's default hashing and adjust Uno's generated CSS to have :global() wrappers only around the parts of rule selectors which are added in generation (e.g. `[dir="rtl"]`). There is still a good amount of footwork required for this as seen in the [apply-directives feature](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped#apply-directives). This means cutting out the passing classes to children feature.
2) Do all of the above + compile class names just when they are passed to children to be able to keep the feature.
3) Compile all class names, let them be global, and inject anything that needs to come first, like resets, into the head before sveltekit styles as this PR is doing.

Some like this passing classes feature, and some don't like it, but for my usage Svelte Scoped does not have feature parity with regular global styles setups without it. Who knows, maybe I'm the odd duck though. 😳

BlankParticle:
As `svelte-scoped` is being moved to a new package so I am commenting here instead of creating a new issue.
`@unocss/preset-web-fonts` doesn't work with `svelte-scoped`, I have tried all providers and tried in both `dev` and `preview` mode. It works with `@unocss/extractor-svelte`. According to my investigation font imports are not added to the generated css.
Can you please check and update me on the issue

jacob-8:
> `@unocss/preset-web-fonts` doesn't work with `svelte-scoped`

Yes, `@unocss/preset-web-fonts` adds to the preflights which did not have a robust solution previously. This PR solves that issue as you can see in the [stackblitz example](https://stackblitz.com/fork/github/jacob-8/svelte-scoped-uno/tree/main/examples/sveltekit-vite-plugin) of [svelte-scoped-uno](https://github.com/jacob-8/svelte-scoped-uno/blob/main/packages/svelte-scoped-uno/README.md). That package is the same code as this PR and is where the code found in this PR was cultivated. The only difference being the package name and global styles placeholder strings so if you want to try out this PR's features in advance you can follow the [install instructions](https://github.com/jacob-8/svelte-scoped-uno/blob/main/packages/svelte-scoped-uno/README.md) of that package and learn more about usage by using this PR's [docs preview](https://deploy-preview-2552--unocss.netlify.app/integrations/svelte-scoped).

antfu:
Merging this for now to move forward, thanks a lot for the great work!

PR #2560, opened 2023-04-29
docs: link next.js integration guide in sidebar and header

moinulmoin:


antfu:
Thanks. I didn't do it because I consider it not finished. Currently, we only list integrations that we directly supported. I think we need to overhaul the guides to provide more "guide" for each framework, but no only for Next. Thanks.

PR #2571, opened 2023-05-04
fix(vite): fix a broken issue when used with @vitejs/plugin-legacy

Dunqing:
fix: #2563


test/fixtures.test.ts
```diff
@@ -47,6 +47,18 @@ describe.concurrent('fixtures', () => {
     expect(js).contains('uno-tacwqa')
   })
 
+  it('vite legacy', async () => {
+    const root = resolve(__dirname, 'fixtures/vite-legacy')
+    await fs.emptyDir(join(root, 'dist'))
+    await execa('npm', ['run', 'build'], { cwd: root })
```
Dunqing, line 53: ### Additional Information
I try using `await build({...})` first. but the test case not work and `../src/uno.svg referenced in assets/index-!~{001}~.js-unocss-hash.css didn't resolve at build time, it will remain unchanged to be resolved at runtime` error appears

And then i use `await execa('npm', ['run', 'build'], { cwd: root })` instead of it just work fine.

Dunqing: fixes: https://github.com/unocss/unocss/pull/2571/commits/dac2bd5acfaf524a8341938c994c9262721002a8

antfu:
Windows is failing. If you can confirm it's a non-issue but more like timeout thing, we could skip the fixture test on Windows

Dunqing:
> Windows is failing. If you can confirm it's a non-issue but more like timeout thing, we could skip the fixture test on Windows

Yes, it's just a timeout thing

PR #2577, opened 2023-05-05
feat(transformer-compile-class): support for custom class names

Techassi:
This PR adds support for custom compile class names. The current behavior looks like this:

```html
<div class=":uno: font-bold text-white">Foo</div>
```

Which generates something like this:

```css
.uno-hpseh7{font-weight:700;--un-text-opacity:1;color:rgba(255,255,255,var(--un-text-opacity));}
```

---

The PR adds support to customize the compiled class name. It is still possible to customize the trigger (`:<TRIGGER>:`) and the prefix (default `uno-`). The new trigger syntax is: `:<TRIGGER>(-NAME):`. This allows the following behavior:

```html
<div class=":uno-foo: font-bold text-white">Foo</div>
```

```css
.uno-foo{font-weight:700;--un-text-opacity:1;color:rgba(255,255,255,var(--un-text-opacity));}
```

---

The custom compiled class name is added to the `tokens` set. If a duplicate is encountered, an error is thrown.

#### Note

Unrelated to this PR, but still worth a mention: the playground doesn't pass `tokens` in the `UnocssPluginContext`. It is `undefined` and results in errors when trying to run the compile class transformer. (I guess this also applies to other transformers)


Techassi:
Added test. PR should now be ready for review.

packages/transformer-compile-class/src/index.ts
```diff
@@ -34,12 +34,12 @@ export interface CompileClassOptions {
 
 export default function transformerCompileClass(options: CompileClassOptions = {}): SourceCodeTransformer {
   const {
-    trigger = ':uno:',
+    trigger = 'uno',
```
antfu, line 37: Why do we change this? The idea was allowing ppl to have custom trigger other than forcing to use `:xxx:`

test/transformer-compile-class.test.ts
```diff
@@ -34,22 +34,26 @@ describe('transformer-compile-class', () => {
...
     `.trim())
     expect(result.code.trim()).toMatchInlineSnapshot(`
-      "<div class=\\"uno-pe1esh\\">
-      <div class=\\"foo bar\\">
+      "<div class=\\"uno-: bg-red-500 text-xl font-bold border border-gray-200 dark:hover:bg-green-500 transform scale-5\\">
+      <div class=\\":uno: foo bar\\">
 
-      <div class=\\"uno-cbgd7b foo\\">
-        <div class=\\"uno-s9yxer\\"/>
+      <div class=\\":uno: text-center sm:text-left foo\\">
```
antfu, line 40: This seems to be an regression

packages/transformer-compile-class/src/index.ts
antfu: Maybe we could have the trigger to accept a RegExp and provide the first group to be the generated name?

test/transformer-compile-class.test.ts
Techassi: This should be fixed in f8493bf.

packages/transformer-compile-class/src/index.ts
Techassi: The trigger is still customizable, but will now require wrapping it in `:`. Possible values are: `:uno:`, `:mytrigger:` or `:mytrigger-foo:`. But we could also switch to a Regex like you suggested so that users have to most flexibility.

antfu: This could be a breaking change and will lose the flexibility of the choices of conventions, that I would like to avoid.

Techassi: Then we have to move to a user-provided Regex. The requirements would be: We should only accept regular expressions with one group. This group will then be used for the custom compile class name. If the group match is empty (undefined), the transformer will use the hashed name like it currently does.

Let me know if this makes sense, and if yes, I will go ahead and include the required changes in this PR.

antfu: Yeah I think so, we could use a named regex group. As long as we properly document it I think it should be fine.

Techassi: I now converted the trigger to a user-provided Regex. PR is ready for review.

packages/transformer-compile-class/src/index.ts
```diff
@@ -1,12 +1,33 @@
...
+   *
+   * This regular expression matches `:uno-MYNAME:` and uses `MYNAME` in
+   * combination with the class prefix as the final class name, for example:
+   * `.uno-MYNAME`. It should be noted that the regex literal needs to include
+   * the global flag `/g`.
+   *
+   * @default `/(["'`]):uno:\s([^\1]*?)\1/g`
    */
-  trigger?: string
+  trigger?: RegExp
```
antfu, line 30:
```suggestion
  trigger?: string | RegExp
```

To keep backward compatibility, we could allow strings that do extract match (without the capability if custom name)

antfu:
Can we have some tests for it, as well as the custom regex? Thanks

packages/transformer-compile-class/src/index.ts
Techassi: Changed in c00c5e5

Techassi:
Added tests in abc2bac.

antfu (approved):

PR #2589, opened 2023-05-08
feat: support `DEFAULT` type in colors

Dunqing:


Shinigami92:
@Dunqing I extracted #2592 util helpers to `@unocss/core` so that these can be individually applied to custom colors when defining them manually

packages/preset-mini/src/_theme/types.ts
```diff
@@ -9,7 +9,7 @@ export interface ThemeAnimation {
 }
 
 export interface Colors {
-  [key: string]: Colors | string
+  [key: string]: Colors & { DEFAULT?: string } | string
```
antfu, line 12:
```suggestion
  DEFAULT?: string
  [key: string]: Colors | string
```

This would do the same, right?

Dunqing: A little bit different.

### { [key: string]: Colors & { DEFAULT?: string } | string }
``` ts
colors: {
  red: {
     DEFAULT: ''
  },
  DEFAULT: '' // ts incorrect
}
```

###   { DEFAULT?: string; [key: string]: Colors | string }
``` ts
colors: {
  red: {
     DEFAULT: ''
  },
  DEFAULT: '' // ts correct
}
```

Issue #2588, opened 2023-05-08
Support DEFAULT for theme colors


Shinigami92:
### Clear and concise description of the problem

I'm configuring the preset for my company

```ts
function presetCompany(options: PresetCompanyOptions = {}): Preset<Theme> {
  const uno = presetUno(options);
  return {
    ...uno,
    name: "@company/preset-company",
    theme: {
      ...uno.theme,
      colors: {
        ...uno.theme?.colors,
        primary: {
          50: "#......",
          100: "#......",
          200: "#......",
          300: "#......",
          400: "#......",
          500: "#......",
          600: "#......",
          700: "#......",
          800: "#......",
          900: "#......",
          950: "#......",
          DEFAULT: "#......",
        },
      },
    },
  } as Preset<Theme>;
}
```

However, as we can see, I need to configure a DEFAULT like e.g. possible in TailwindCSS: https://tailwindcss.com/docs/customizing-colors#color-object-syntax

This does not seem to work 🙁 
It always takes the `...-500` as default

How can I change the default so `text-primary === text-primary-700`?

### Suggested solution

support `DEFAULT` like TailwindCSS

### Alternative

Document how it is possible

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://uno.antfu.me).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Dunqing:
`unocss` already support `DEFAULT`. You can see https://github.com/Dunqing/unocss/blob/main/packages/preset-mini/src/_theme/colors.ts#L360-L371

Shinigami92:
> `unocss` already support `DEFAULT`. You can see [Dunqing/unocss@`main`/packages/preset-mini/src/_theme/colors.ts#L360-L371](https://github.com/Dunqing/unocss/blob/main/packages/preset-mini/src/_theme/colors.ts?rgh-link-date=2023-05-08T12%3A48%3A40Z#L360-L371)

That workes!

Thx, but is there a way so that I don't need to manually copy this

```ts
// assign default color, and color shortcuts
Object.values(colors).forEach((color) => {
  if (typeof color !== "string") {
    color.DEFAULT = color.DEFAULT || color[400];
    Object.keys(color).forEach((key) => {
      const short = +key / 100;
[...]

PR #2592, opened 2023-05-08
feat(util): shorts and default helper

Shinigami92:
references #2588


packages/core/src/utils/helpers.ts
```diff
@@ -28,3 +28,26 @@ export function notNull<T>(value: T | null | undefined): value is T {
...
 export function noop() {}
+
+export interface Colors {
+  [key: string]: Colors | string
+}
+
+export function assignDefaultToColors(colors: Colors): void {
+  Object.values(colors).forEach((color) => {
+    if (typeof color !== 'string' && color !== undefined) 
+      color.DEFAULT = color.DEFAULT || color[400]
```
antfu, line 39: I wonder if we could go recursive to support deep colors

```ts
const colors = {
    red: {
      50: '#fef2f2',
      100: '#fee2e2',
      200: '#fecaca',
      deep: {
        100: '#...',
        DEFAULT: '#...'
      }
    }
  }
```

We didn't do that because it was only for the built-in colors object

Shinigami92: IMO we could do this in a separate PR

Shinigami92:
@antfu it's now ready for review

I also fixed that it does not override short values if they already exists

packages/core/src/utils/helpers.ts
antfu: Then I don't actually see the value of exposing those utilities. This code was not meant to be used as utilities, they are just to keep the bundle small and make the codebase more maintainable (as all are static). Do you want to use them externally to your own theme object?

antfu:
Closing this for now until we see really and general use cases. Thanks for bringing this up.

Issue #2513, opened 2023-04-17
Duplicate styles injected when using Astro integration


markmals:
### UnoCSS version

0.51.4

### Describe the bug

When using the UnoCSS Astro integration, two different sets of the same UnoCSS styles are injected into the `<head>`, one with the attribute `data-astro-dev-id="{path/to}/__uno.css"` and another with the attribute `data-vite-dev-id="/__uno.css"`

### Reproduction

https://stackblitz.com/edit/github-u8fq8k?file=src/pages/index.astro

### System Info

macOS 13.3.1, Chrome 112.0.5615.121, Astro 2.3.0

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

jsonleex:
This problem still exists.

https://stackblitz.com/edit/github-u8fq8k-s1lqwe?file=package.json

<img width="232" alt="image" src="https://github.com/unocss/unocss/assets/25883665/ba178fb3-6c01-4075-935d-0c516760f220">

<img width="830" alt="iShot_2023-05-29_07 56 28" src="https://github.com/unocss/unocss/assets/25883665/01825ad7-8d63-4ffa-a79f-ddd7913d4d10">

PR #2593, opened 2023-05-09
fix(astro): inject duplicate style

Dunqing:
fix: #2513

Avoid inject styles in non-ssr

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -29,14 +29,8 @@ function AstroVitePlugin(options: AstroVitePluginOptions): Plugin {
       if (id.endsWith(UNO_INJECT_ID))
         return injects.join('\n')
 
-      if (!options?.ssr && id.includes(UNO_QUERY_KEY)) {
-        return {
-          code: '',
-          map: {
-            mappings: '',
-          },
-        }
-      }
+      if (!options?.ssr && id.includes(UNO_QUERY_KEY))
+        return ''
     },
   }
 }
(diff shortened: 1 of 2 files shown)


packages/vite/src/modes/global/dev.ts
```diff
@@ -146,7 +148,14 @@ export function GlobalModeDevPlugin({ uno, tokens, tasks, flushTasks, affectedMo
...
         },
       },
-      resolveId(id) {
+      resolveId(id, importer) {
+        if (
+          isAstro
+          && importer === 'astro:scripts/page-ssr.js'
+          && id.includes('?uno')
+        )
+          return `${id}?uno`
```
antfu, line 157: Emm. I don't think this is a good approach to have Astro-specific logic inside the Vite plugin.

antfu (changes requested):

packages/vite/src/modes/global/dev.ts
Dunqing: I think so too, but I haven't thought of a better way.

packages/astro/src/index.ts
```diff
@@ -4,6 +4,41 @@ import type { AstroIntegration } from 'astro'
 import type { VitePluginConfig } from '@unocss/vite'
 import VitePlugin from '@unocss/vite'
 import type { UserConfigDefaults } from '@unocss/core'
+import type { Plugin } from 'vite'
+
+const UNO_INJECT_ID = 'uno-astro'
+const UNO_QUERY_KEY = 'uwa'
```
antfu, line 10: What does uwa mean? Can we use a longer and more descriptive name?

Issue #2608, opened 2023-05-11
Support Tailwind CSS 3.3 variables without `var()`


AaronBeaudoin:
### Clear and concise description of the problem

Tailwind CSS 3.3 allows you to write `bg-[--brand-color]` instead of `bg-[var(--brand-color)]`:
https://tailwindcss.com/blog/tailwindcss-v3-3#css-variables-without-the-var

I've added support for this into my own project with the following code, inspired by [`@unocss/preset-rem-to-px`](https://github.com/unocss/unocss/blob/5d693ec7d2713a1421f8208cc87b6273b09405bf/packages/preset-rem-to-px/src/index.ts#L20):

```js
export default defineConfig({
  ...
  postprocess: (util) => {
    util.entries.forEach((item) => {
      const value = item[1];
      if (typeof value !== "string") return;
      if (/^--/g.test(value)) item[1] = `var(${value})`;
    });
  }
});
```

Since Tailwind/Windi compatibility seems to be one of the current goals of UnoCSS, and this feature is very nice for convenience, is there any reason why it shouldn't be officially included? At least perhaps as a `@unocss/preset-short-var` preset or something?

### Suggested solution

See description above.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://uno.antfu.me).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
> is there any reason why it shouldn't be officially included

No. I am open to that. PR welcome to directly support in preset-mini/preset-wind.

PR #2616, opened 2023-05-13
feat(preset-mini): support css variables without `var()`

Dunqing:
close: #2608


zyyv:
@Dunqing I think we didn't need to convert output css in `postprocess ` hook, and i created a new PR in #2618. You can review it.

PR #2635, opened 2023-05-18
docs: update expected type in variants example

Phojie:
The expected type that comes from the [variant](https://github.com/unocss/unocss/blob/main/packages/core/src/types.ts#LL334C5-L334C5) is `Array` not `Object`


zyyv (approved):

zyyv:
Nice catch, thanks.

Issue #2646, opened 2023-05-22
0.52.0 broke Astro integration


Hrishikesh-K:
### UnoCSS version

0.52.0

### Describe the bug

Running `astro dev` in a project with UnoCSS v0.52.0 gives the error:

```
Failed to load url /uno.css?uno-with-astro-key (resolved id: /uno.css?uno-with-astro-key) in uno-astro. Does the file exist?
14:20:42 [vite] Error when evaluating SSR module uno-astro: failed to import "/uno.css?uno-with-astro-key"
|- Error: Failed to load url /uno.css?uno-with-astro-key (resolved id: /uno.css?uno-with-astro-key) in uno-astro. Does the file exist?
    at loadAndTransform (file://file:///home/projects/unocss-2646/node_modules/vite/dist/node/chunks/dep-4d3eff22.js:53428:21)

14:20:42 [vite] Error when evaluating SSR module astro:scripts/page-ssr.js: failed to import "uno-astro"
|- Error: Failed to load url /uno.css?uno-with-astro-key (resolved id: /uno.css?uno-with-astro-key) in uno-astro. Does the file exist?
    at loadAndTransform (file://file:///home/projects/unocss-2646/node_modules/vite/dist/node/chunks/dep-4d3eff22.js:53428:21)

14:20:42 [vite] Error when evaluating SSR module /home/projects/unocss-2646/src/pages/index.astro: failed to import "astro:scripts/page-ssr.js"
|- Error: Failed to load url /uno.css?uno-with-astro-key (resolved id: /uno.css?uno-with-astro-key) in uno-astro. Does the file exist?
    at loadAndTransform (file://file:///home/projects/unocss-2646/node_modules/vite/dist/node/chunks/dep-4d3eff22.js:53428:21)

 error   Could not import `/uno.css?uno-with-astro-key`.
  Hint:
    This is often caused by a typo in the import path. Please make sure the file exists.
  File:
    /home/projects/unocss-2646/file:/file:/home/projects/unocss-2646/node_modules/vite/dist/node/chunks/dep-4d3eff22.js:53428:21
  Stacktrace:
Error: Failed to load url /uno.css?uno-with-astro-key (resolved id: /uno.css?uno-with-astro-key) in uno-astro. Does the file exist?
    at loadAndTransform (file://file:///home/projects/unocss-2646/node_modules/vite/dist/node/chunks/dep-4d3eff22.js:53428:21)

 error   Could not import `/uno.css?uno-with-astro-key`.
[...]

PR #2648, opened 2023-05-22
fix(astro): incorrect path and build failure

Dunqing:
close: #2646

The real file path should be used here, but I don't know why it works in the unocss example.
I have tested it in the broken repo and that work.

And the build failure was also fixed...

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -19,11 +19,16 @@ function AstroVitePlugin(options: AstroVitePluginOptions): Plugin {
     name: 'unocss:astro',
     apply: 'serve',
     enforce: 'pre',
-    resolveId(id, importer) {
+    async resolveId(id, importer) {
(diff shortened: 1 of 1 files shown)


packages/astro/src/index.ts
```diff
@@ -15,15 +15,25 @@ interface AstroVitePluginOptions {
...
-    resolveId(id, importer) {
+    configResolved(_config) {
+      config = _config
+    },
+    async resolveId(id, importer) {
       if (id === UNO_INJECT_ID)
         return id
-      if (importer?.endsWith(UNO_INJECT_ID))
-        return `${id}${id.includes('?') ? '&' : '?'}${UNO_QUERY_KEY}`
+      if (importer?.endsWith(UNO_INJECT_ID) && config && config.command === 'serve') {
```
Dunqing, line 30: need remove `&& config` here @antfu Can you help me?

PR #1963, opened 2022-12-05
feat(vite, webpack): add `extraContent` option

sibbng:
Closes #1944
Fixes #1937 // astro islands
Fixes #1734 // astro islands
Fixes #1947 // nuxt content
Fixes #2005 // nuxt content

> **Note**
> At the moment nuxt module is not pre-configured. I will add it soon.

I wanted to emphasize the _external_ word by naming it `watchExternal`. Otherwise calling it something more generic might confuse people who migrate from other tools.

We're applying transformer plugins to files captured by this filter. But not really transforming them. Since those files are not processed by Vite, it's better to not transform them at all. I far as I see, our CLI package directly modify **source files** which is something really unintuitive.

In the case of Astro, files captured by this filter will go through transformer plugins. Because client-only components in Astro will be processed by Vite during client build. Since we already applied transformers on the SSR side manually, the required classes are already generated in bundled CSS.

Issue #419, opened 2022-01-05
[@unocss/webpack] CSS generation is inconsistent with persistent caching enabled


chrissantamaria:
## Problem

👋 noticed an issue with `@unocss/webpack` and Webpack 5's new [persistent caching](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#persistent-caching) feature. It looks like `@unocss/webpack`'s approach of using loaders to extract code for class parsing can break with persistent caching - when a module is considered a cache hit, loaders are skipped entirely. This has the end result of subsequent builds having a potentially incomplete set of unocss classes included.

I've created a [minimal reproduction repo](https://github.com/chrissantamaria/unocss-bug-persistent-caching) to help isolate this issue - instructions are included in the README.

## Potential solution

One approach could be to perform the code extraction (i.e. pushing to `tasks`) in the `NormalModuleFactory` stage as all modules (cached or otherwise) pass through this hook when being added to the dependency tree. Unfortunately, this requires manually reading the file contents rather than pulling what the loader has already read. Fortunately, `compiler.inputFileSystem.readFile` can be used which (from what I can tell) has [internal caching](https://github.com/webpack/webpack/blob/main/lib/node/NodeEnvironmentPlugin.js#L44) to make the impact of repeated reads fairly negligable.

I've loosely tested this approach and it seems to solve the original issue - the second cached build succeeds and produces the same output. It's unfortunately a bit hacky and doesn't compose cleanly with `unplugin` but could at least be used as a starting point for a cache-friendly setup. Happy to post a snippet or open a rough PR.

However, for some reason the third (also cached) build seems to crash with or without the fix:

<details>
  <summary>Build output</summary>

```
➜  unocss-bug-persistent-caching git:(main) ✗ yarn build
yarn run v1.22.11
warning package.json: No license field
$ webpack
assets by status 131 KiB [cached] 3 assets
Entrypoint main = main.css main.js
cached modules 134 KiB (javascript) 238 bytes (css/mini-extract) [cached] 13 modules
orphan modules 39 bytes [orphan] 1 module
./_virtual_uno.css 39 bytes [built] [1 error]

ERROR in ./_virtual_uno.css (./_virtual_uno.css.webpack[javascript/auto]!=!./node_modules/css-loader/dist/cjs.js!./_virtual_uno.css)
Module build failed (from ./node_modules/css-loader/dist/cjs.js):
[...]

PR #2652, opened 2023-05-23
feat(webpack): add scan-only mode to  adapt webpack5 persistent cache 

GrapevineLin:
**fix**：https://github.com/unocss/unocss/pull/419

**Description:**
This issue is related to webpack5 persistent cache because @unocss/webpack using loaders to extract code . Inspired by Windicss's pre-scanning, I think we can do it similarly. Fortunately there is a solution already: https://github.com/unocss/unocss/pull/1963. 

Therefore, I added an option `scanModeOnly`  to only use file-sacnning to extract code, which can solve the issue of persistent caching.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -648,6 +648,12 @@ export interface ExtraContentOptions {
    * Plain text to be extracted
    */
   plain?: string[]
+
+  /**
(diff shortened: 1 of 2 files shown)


antfu:
https://github.com/unocss/unocss/pull/2503 should solve this better

PR #2657, opened 2023-05-23
docs(svelte-scoped): change "server hooks" to "hooks.server.js" in error message and docs

henrikvilhelmberglund:
The docs say `server.hooks.js` instead of `hooks.server.js` so fixing that, also updated the error message to be more clear about what the file name should be.


henrikvilhelmberglund:
Thanks for fixing the title! 😊

PR #2664, opened 2023-05-24
fix(transformer-compile-class): duplicate class name

nguyentuevuong:
fix: #2667

When using the transformer class, class names defined by the developer are not allowed to have duplicates, but with classes generated by hashFn, duplicates are completely normal. In the latest update, there was confusion in filtering these classes, resulting in unnecessary duplication errors.


antfu:
Can you provide more context and also some tests? Thanks

nguyentuevuong:
> Can you provide more context and also some tests? Thanks
How can I provide context and test cases? I apologize, this is my first time creating a PR, and I am not familiar with the process.

antfu:
I would suggest you create an issue first, describing that is expected vs happening. The best to provide a reproduction, or a failing test to convince the maintainer it's a bug.

nguyentuevuong:
> I would suggest you create an issue first, describing that is expected vs happening. The best to provide a reproduction, or a failing test to convince the maintainer it's a bug.

I created new issue: #2667

antfu:
Seem to related to https://github.com/unocss/unocss/pull/2577, @Techassi do you mind to have a look? Thanks

Techassi:
Yes, I introduced the check to prevent multiple custom class names. Just as the `FIXME` says, we ideally should also compare the hashed class name. If the hash value matches, we know the styles are the same and in this case we allow duplicate class names.

The fix proposed in this PR fixes the issue, but I don't think this is the right approach. It doesn't allow defining the same custom class names multiple times even if the styles are the same. This is a valid use-case and thus should be possible.

nguyentuevuong:
@Techassi I have just added a new commit that thoroughly resolves the issue you mentioned above without needing to change the token structure from Set to Map. Please check if you can utilize this approach.

Techassi:
Sorry for the delayed response. I will look into it!

PR #2692, opened 2023-05-31
docs(preset-web-fonts): modify the usage of customFetch in the documentation

lainbo:
### What issue did I discover
In the current documentation, when using `proxy-agent` in the `uno.config.ts`, you will encounter an error message saying `proxyAgent.default is not a constructor`.

image
---
### What did I modify
I have updated the documentation to use `https-proxy-agent`, and in actual testing with Clash and v2rayN proxies, it functions correctly without any error messages

### Screenshot of the test results
**clash**
image

---
**v2rayN**
image

### DIFF
--- a/docs/presets/web-fonts.md
+++ b/docs/presets/web-fonts.md
@@ -66,6 +66,17 @@ PR welcome to add more providers. 🙌
 ### Custom fetch function
 
 Use your own function to fetch font source.
+::: code-group
+  ```bash [pnpm]
+  pnpm add -D https-proxy-agent
+  ```
+  ```bash [yarn]
+  yarn add -D https-proxy-agent
+  ```
+  ```bash [npm]
+  npm install -D https-proxy-agent
+  ```
+:::
 
 ```ts
 // uno.config.ts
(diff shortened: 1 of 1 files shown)


lainbo:
You are right. Perhaps I can remove the related comments and explanations. At the very least, we should ensure that the content in the documentation is accurate, shouldn't we?

lainbo:
> Thank you, but no.
> 
> * Not everyone needs a proxy
> * For ppl who need proxy, this should be a piece of common/general knowledge when working with Node etc, and it's not related to UnoCSS itself. It shouldn't be our responsibility to teach this.

You are correct. Perhaps I can remove the related comments and explanations. At the very least, we should ensure that the content in the documentation is accurate, shouldn't we

antfu:
I would say we should even remove the mentioning about proxy at all, to avoid confusing. They should refer to axios or other fetching lib to see how to do so. PR welcome if you want to do so.

lainbo:
> I would say we should even remove the mentioning about proxy at all, to avoid confusing. They should refer to axios or other fetching lib to see how to do so. PR welcome if you want to do so.

okay I understand. Thank you for your patient explanation

PR #2693, opened 2023-05-31
feat(preset-icons): fallback on env does not support NodeLoader

Fatpandac:


Fatpandac:
I'm writing an extension similar to [UnoCSS Interactive Docs](https://unocss.dev/interactive/) for Raycast. 
Raycast doesn't support [Node Loader](https://github.com/Fatpandac/unocss/blob/e82c5fd106301ce396d6ade20a2e39e3c94877a2/packages/preset-icons/src/index.ts#L9-L19), just like the UnoCSS VScode extension, so I'm adding this condition.

image

Fatpandac:
I am find out this way, it may not be correct. 🤕

PR #2694, opened 2023-06-01
feat(preset-mini): pseudo-class function for custom arguments

enpitsuLin:
~~Add `has-*`,`group-has-*`, and `peer-has-*` support for the [`:has()` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:has)~~

This PR implements both `:not()`/`:is()`/`:where()`/`:has()` pseudo-class function and its parent variant with custom arguments.

### Only pseudo-class

```html
<div class="has-[[data-orange]]:c-orange">
  <div data-orange/>
<div>

<!-- Ouput -->
<style>
  .has-\[data-orange\]\:c-orange:has([data-orange]) {
    --un-text-opacity: 1;
    color: rgba(251, 146, 60, var(--un-text-opacity));
  }
</style>
```

### /w parent variant

Work with both `.group`/`.parent`/`.peer`/`.previous`

```html
<div class="group">
  <div class="group-has-[[data-orange]]:c-orange"><div>
  <div data-orange/>
</div>

<!-- Ouput -->
<style>
  .group:has([data-orange]) .group-has-\[\[data-orange\]\]\:c-orange {
    --un-text-opacity: 1;
    color: rgba(251, 146, 60, var(--un-text-opacity));
  }
</style>
```


packages/preset-wind/src/variants/has.ts
```diff
@@ -0,0 +1,74 @@
...
+      const bracketValue = h.bracket(match)
+      if (!bracketValue)
+        return
+
+      if (match) {
+        return {
+          matcher: reset,
+          handle: (input, next) => next({
+            ...input,
+            selector: `.group:has(${bracketValue}) ${input.selector || ''}`,
```
antfu, line 45: We have an option to toggle `.group` style: https://github.com/unocss/unocss/blob/5c9fc9942fcb564acc063ac4d18926a2ae14a344/packages/preset-mini/src/_variants/pseudo.ts#L254

packages/preset-wind/src/variants/has.ts
```diff
@@ -0,0 +1,74 @@
...
+      const bracketValue = h.bracket(match)
+      if (!bracketValue)
+        return
+
+      if (match) {
+        return {
+          matcher: reset,
+          handle: (input, next) => next({
+            ...input,
+            selector: `.peer:has(${bracketValue}) ~ ${input.selector || ''}`,
```
antfu, line 68: Same for `.peer`

packages/preset-wind/src/variants/has.ts
enpitsuLin: I have done this and moved to the `preset-mini` package, had to add some code because the previous implementation might work not perfectly with custom arguments (and I'm kinda confused with it).

And implement both four pseudo-class functions with it (so change the pr title

packages/preset-mini/src/_variants/pseudo.ts
```diff
@@ -138,23 +139,39 @@ function taggedPseudoClassMatcher(tag: string, parent: string, combinator: strin
...
+      `${parent}${escapeSelector(label)}${pseudo}`,
+    ]
+  }
+
   return {
     name: `pseudo:${tag}`,
     match(input, ctx) {
       if (!(splitRE && pseudoRE && pseudoColonRE)) {
         splitRE = new RegExp(`(?:${ctx.generator.config.separators.join('|')})`)
         pseudoRE = new RegExp(`^${tag}-(?:(?:(${PseudoClassFunctionsStr})-)?(${PseudoClassesStr}))(?:(/\\w+))?(?:${ctx.generator.config.separators.join('|')})`)
```
enpitsuLin, line 162: confused me on this, it seems can implement custom arguments at the beginning, but didn't

don't know if there has some edge case that I just add code

PR #2695, opened 2023-06-01
fix(svelte-scoped): global wrap preflights and safelist classes when used in component libraries

jacob-8:
Trying to use presetTypography's `.prose` class in a component library exposed the fact that it was not possible. As seen in [prose/OutputDev.svelte](https://github.com/unocss/unocss/compare/improve-preflights-safelist-usage-in-component-libraries?expand=1#diff-e429f5b7f0e6095695ca61697d534c7ecbc9acaa29051d4d72bc3b555ce0036d), we can now use `.prose` in a component assuming we've placed the needed class in our safelist (I set the selector to `.uno-prose` in the test's Uno config)

This was made possible by also wrapping preflights placed into a component with `:global()` to keep them from being stripped out. The matching pattern ignores the default preflights and just applies to specially added ones (starting with a period or bracket), like the large list seen in presetTypography.

This also led me to the realization that `<style uno:safelist></style>` is absolutely useless as safelisted styles, if they are needed in the safelist for the normal reason that they only exist at runtime, will always get stripped out. This was forcing us to provide the advice to use `<style uno:safelist global>`, but that creates dependence on the old Svelte preprocessor (Svelte has jumped on board with using Vite's default Preprocessor) so I realized there is no reason not to just do the global part ourselves. Now users can just write `<style uno:safelist>` and it will work as expected. Component library consumers won't be tied down to the old Svelte preprocessor.

Lastly, I fixed a bug whereby preflights/safelists would not be output into the `style` block if there was no `apply` or `theme`.


antfu:
Can you help resolve the conflicts? Thanks

antfu (approved):

PR #2714, opened 2023-06-03
fix(preset-mini)!: fix pseudo variants generation order, close #2713

### LINKED ISSUE #2713: Hover being triggered when disabled even with specified disabled: variants
### UnoCSS version

0.52.7

### Describe the bug

In tailwind and their product tailwind ui hover states are not triggered when hovering over an item that is disabled when disabled styles are specified.

In unocss when an item is disabled specifies disabled: classes, they are still overridden by hover on hover.

this breaks feature parity from moving from tailwind to uno when using presetUno

uno from what i can tell provides no configuration or speicifc default behaviour when dealing with this

### Reproduction

https://stackblitz.com/edit/unocss-unocss-rxp1nt?file=src%2FApp.vue for uno

https://play.tailwindcss.com/9eEGbmAT2c tailwind (for reference)

### System Info

Macos
arc
Vite + Vue 3

happens in the playgrounds specified

### Validations

antfu:
fix #2713

- Add `sort` based on the order of variants list
- Disable CSS merge for pseudo variants
- Align the list order of Tailwind


sibbng (commented):
We may not need the `sort` option here anymore: https://github.com/unocss/unocss/blob/5c9fc9942fcb564acc063ac4d18926a2ae14a344/packages/preset-mini/src/_variants/pseudo.ts#L81-L95

Since the `sort` value varies based on their index in the array now, this is not safe too.

But, I would suggest keeping the `noMerge` option here or we could disable merging for pseudos in general. In #1973 I explained why merging could be so dangerous.

Issue #2103, opened 2023-01-21
NextJS hmr not working


nxy7:
Hello, I'm trying out UnoCSS (great project) and it seems like nextjs HMR is not working even with cache set up as in example

https://github.com/unocss/unocss/blob/bb321caea95dd0a6e0cc44c22d897a9dc96cb6c9/examples/next/next.config.js#L6-L12

I think what's happening is that CSS file is updated, but doesn't get loaded - i see loading spinner but need to refresh the page to see UnoCSS styles updated.

eramdam:
I'm seeing the same behavior with a NextJS 13.x project. I ended up using the CLI similar to what [this article](https://thundermiracle.com/blog/en/2022-06-07-next-with-unocss/) describes 😅

RealTong:
I have the same mistake. (NextJS hmr not working )
```json
    "next": "13.3.0",
    "@unocss/preset-attributify": "^0.50.8",
    "@unocss/preset-uno": "^0.50.8",
    "@unocss/preset-web-fonts": "^0.50.8",
    "@unocss/reset": "^0.50.8",
    "@unocss/webpack": "^0.50.8",
    "unocss": "^0.50.8"
```
```javascript
// next.config.js
const UnoCSS = require('@unocss/webpack').default;

/** @type {import('next').NextConfig} */
const nextConfig = {
    reactStrictMode: true,
    webpack: (config) => {
        config.cache = false
        config.plugins.push(UnoCSS({ presets: [] }));
        return config
    }
}

module.exports = nextConfig

```

```javascript
// unocss.config.js
import {
    defineConfig,
    presetAttributify,
    presetUno,
} from 'unocss';

export default defineConfig({
    presets: [
        presetUno(),
        presetAttributify(),
    ]
})
```

sibbng:
FYI we also have a PostCSS plugin: https://unocss.dev/integrations/postcss

If you don't use any [transformer](https://unocss.dev/presets/#transformers) or only use [transformer-directives](https://unocss.dev/transformers/directives) I would suggest you give it a try. Features of [transformer-directives](https://unocss.dev/transformers/directives) already included in PostCSS plugin, so you don't need to have it in config.

RealTong:
> FYI we also have a PostCSS plugin: https://unocss.dev/integrations/postcss
> 
[...]

PR #2724, opened 2023-06-05
docs: Examples Next 13 app Dir

iamhectorsosa:


jacob-8:
Link to test current code on codesandbox: https://codesandbox.io/s/github/iamhectorsosa/unocss/tree/docs-examples-next/examples/next

jacob-8:
Our convention is to use links for local packages to enable maintainers to easily test their changes against all examples. Like this:

```json
{
  "devDependencies": {
    ...
    "@unocss/postcss": "link:../../packages/postcss",
    "@unocss/reset": "link:../../packages/reset",
    "unocss": "link:../../packages/unocss"
  }
}
```

Then Stackblitz will change those to `latest` and work fine. I can't find any way to modify files in Codesandbox before it installs packages, so I've just set them to latest. The closest I could come with Codesandbox docs and Copilot help was to create a `.codesandbox/Dockerfile` like this:

```
FROM node:18

# Set working directory
WORKDIR /app

# Copy custom script to container
COPY .codesandbox.sh /app

# Make custom script executable
RUN chmod +x .codesandbox.sh

# Copy application code to container
COPY . /app

# Rename links to latest
CMD ["./.codesandbox.sh"]
```

And then a script like this:

```bash
#!/bin/sh

updateDependencyLinksToLatest() {
  local filename="$1"
  local contents=$(cat "$filename")
  local updatedContent=$(sed 's/"link:...+"/"latest"/gi')
  echo "Updated content: $updatedContent"
  echo "$updatedContent" > "$filename"
}

updateDependencyLinksToLatest "/app/package.json"
```

But that sed part is broken and my skills in this area aren't good enough to know how to do it. If someone else wants to pick this up and make it work - great. Otherwise I guess we can settle for the opposite - to just have maintainers update the packages to links if they ever want to run this example locally against changes.

jacob-8:
Thanks for this @iamhectorsosa. I made a few changes to make it more parallel to the way the rest of our examples run, and some little stuff. I can look at it again when you have it finished. And if you know any solution to the problem mentioned in https://github.com/unocss/unocss/pull/2724#issuecomment-1579681609 - feel free to add that. 

Taking another example as a starting point is helpful, but do feel free to deviate as much as it helps people use UnoCSS and not just copy Tailwind. For example, you added the icons preset but you're still using SVG icon files which are not needed if you use `@iconify-json/logos` - maybe you could try swapping those out. Also I didn't see you take advantage of attributify mode, so remove that preset and the shims for it if you aren't demoing it. Alternatively you can demo it too.

jacob-8:
Fixes #2103

iamhectorsosa:
@jacob-8 Huge huge thanks for taking a look at this! I've revised the work and did the following:

- Demo'd more presets but kept it simple (I referenced them in the Readme)
- Added the steps to guide users in creating this example from scratch in the Readme.

Let me know if there is anything else I can help with! Once the CI is done, I'll open this for review.

jacob-8:
This looks good to me. 

@antfu is it just an oversight that `docs\guide\packages.md` is not included in the public-facing docs? I've referenced it so many times and it would have helped @iamhectorsosa know there was no need to instruct the installation of `@unocss/reset`. Not that it's going to hurt anyone to over-install, but still I've found that Packages list to be very helpful many times.

iamhectorsosa:
@jacob-8 @antfu this makes sense, thanks for the observation. I can update references if needed. I see we still do mention of the reset package in the readme and it is still imported in the global css.

Can you tell me a bit more how this works, where is this reset included? Is it part of the main 'unocss'?

jacob-8:
> Can you tell me a bit more how this works, where is this reset included? Is it part of the main 'unocss'?

You still need to **import** the reset just as you have done (e.g. `import '@unocss/reset/normalize.css'`), but my meaning was that you don't need to **install** (`npm install -D @unocss/reset`). Although as you've pointed out, the [docs](https://unocss.dev/guide/style-reset#installation) do instruct users to install. So then I defer to @antfu on his preference on whether to instruct users to install the reset package.

docs/.vitepress/config.ts
```diff
@@ -172,19 +172,19 @@ const SidebarGuide: DefaultTheme.SidebarItem[] = [
         link: '/integrations/',
       },
       ...Integrations,
+      {
+        text: 'Examples',
+        link: 'https://github.com/unocss/unocss/tree/main/examples',
```
antfu, line 177:
```suggestion
        link: '/integrations/#examples',
```

Maybe we should stop directly linking to the source

examples/next/public/next.svg
antfu, line 1: Can we avoid including those static assets? Thanks.

examples/next/src/app/favicon.ico
antfu, line 1: This as well.

antfu:
I hear that the app dir style is a bit infamous in the community, should we keep those two styles as separate examples?

iamhectorsosa:
@antfu alrighty! I did the following:

- Removed the static assets from the directory and also from the components. I replaced them with just text. This isn't relevant for the example itself so it should be perfectly fine.
- I updated both the top navigation and the side navigation of the Docs to point to the docs instead of the external source for Examples.

> _I heard that the app dir style is a bit infamous in the community, should we keep those two styles as separate examples?_

In my opinion, the `app` directory is the, already default, recommended and stable for production, way of building Next applications. If we keep two examples, we might end up having to remove the `pages` dir at some point. Next still provides support but [recommends](https://nextjs.org/docs/pages) using `app`.

itzomen:
Nice, awesome work @iamhectorsosa 

I'm also setting up a blank project with UnoCss and Next 13 app directory.

Do you plan on leveraging next/font ? Because I'm stuck there, I can't seem to find how to pass the font variables to uno css just like with Tailwind ( https://nextjs.org/docs/pages/building-your-application/optimizing/fonts#with-tailwind-css )

Any insights would be helpful, thanks

antfu (approved):

Issue #2717, opened 2023-06-03
Reset priority in Astro


shishkin:
### UnoCSS version

0.52.7

### Describe the bug

I might be missing something, but I wonder why astro integration loads reset styles after the main `__uno.css` style. This prevents me from overriding some defaults with normal means without increasing specificity or using `!important`.

In the reproduction example I'm using `tailwind-compat.css` reset and am trying to make links underlined again from a preflight. I've read docs about resets, preflights and layers, but nothing seems to help since resets apparently don't use layers.

If there is another solution to that I'd be grateful to learn it.

### Reproduction

https://stackblitz.com/edit/github-mrekxn-qfbrtc?file=astro.config.ts

### System Info

Astro

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

shishkin:
What I also observe in the same repro is that resets and uno styles come _after_ the styles on the page itself. So if page contains generic CSS styles they get reset.

shriharip:
Would love to see it merged with the main :)

PR #2726, opened 2023-06-05
fix(astro): ensure the reset css load first

Dunqing:
fix: #2717 
re-fix: #2513, broken in #2656

Astro doesn't provide any way to control the import order, but we can manually control which file is loaded first via promise.


antfu:
I am not very confident about the complexity added here. Do you think it make sense to send a request to Astro upstream?

Dunqing:
> I am not very confident about the complexity added here. Do you think it make sense to send a request to Astro upstream?

Yes, we can send a request to Astro upstream. But I think we can merge it first, to avoid perpetuating this problem because this is caused by the processing mechanism of `astro`, it may take a long time, or `astro` does not consider solving it.

PR #2736, opened 2023-06-06
fix(preset-mini): support comma-separated grid template arbitrary values

Waleed-KH:
tailwind support grid template arbitrary values to be comma separated

e.g. `grid-cols-[200px,minmax(90px,1fr),auto]`:

- tailwind output: https://play.tailwindcss.com/h7OBzXndk2 🐌
```css
.grid-cols-\[200px\2c minmax\(90px\2c 1fr\)\2c auto\] {
  grid-template-columns: 200px minmax(90px,1fr) auto;
}
```
- unocss output: [https://unocss.dev/play/?html=DwEwlgbg...](https://unocss.dev/play/?html=DwEwlgbgBAxgNgQwM5ILwCIDmAnMIo54C0MA9nEkQNoBMADHQA4AeANALZgB27CzAFAE4mbAIwAzbAEpWCAK4AXUgF10APgBQUKKEixEKDACNMRI3DkBTIgDYGUABZEAzDfXAA9OAibtu6PDIaOgmRNiWILb2Tq7uXpC%2BOt76QcamOJaWXFF0ji42UADuLjRx3pqe5RpAA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6sfFJ1bX1enkwVjCJTqRwfZYD0Pa4pBhwNQBuGFDBowN99k5mqGqyIgqN4ZFyokoqPgCUssLyjCxsnDx8-OeXRwqS0n6HqHQS6xihAEYAHQAJheqAkaFweSqrDAjQA9Aj6CAgdUEZsWmoLuUkNikEA&css=Q&options=N4XyA) 🏎️
```css
.grid-cols-\[200px\,minmax\(90px\,1fr\)\,auto\]{grid-template-columns:200px,minmax(90px,1fr),auto;}
```

in this PR I just added a test unit that fails for the above case, I'll try to dive in and work on a fix


zyyv:
You should use `grid-cols-[200px_minmax(90px,1fr)_auto]`. https://unocss.dev/interactive/?s=grid-cols-[200px_minmax(90px,1fr)_auto]

Waleed-KH:
@zyyv yes I know.
I used comma separated grid template arbitrary values (e.g. `grid-cols-[auto,1fr]`) everywhere while working with tailwind with no problem, so I opened this PR to ensure that it's also supported with unocss and make it easier for anyone migrating from tailwind.

Waleed-KH:
I have committed a fix that replaces the arbitrary value commas with spaces while skipping the commas in functions (between brackets `()`), so `200px,minmax(900px,1fr),auto` will become `200px minmax(900px,1fr) auto`

I didn't use regex like `/,(?![^\(]*\))/g` because it fails to handle all cases especially nested functions, so I wrote a small algorithm that solve the problem, I call it **The Replace Jumper** 😁, because it uses some tricks to jump to where it should do the replace without iterating over every character in the input string (well, technically it does, because it uses `indexOf`).

I also would like to suggest to support it in more places (e.g. `m-[20px,15px]`), I think comma-separated looks nicer than underscore-separated

zyyv:
Thanks, but I don't think some logic should be handled for a rule alone. UnoCSS placesholders for spaces is `_`, which is different from tailwind. If you want to smooth out the differences between them, I think you can create your own private preset using.

Waleed-KH:
I understand, but I would like for UnoCSS to support comma-separated values, because when I write class like this `grid-cols-[auto,1fr]` its like passing 2 values to the `grid-cols` class (I don't think of the comma as a placeholder for space) and I hope it would be supported with more rules (margin, padding...), I could move the code to the value handler chain so it could be used with more rules

Waleed-KH:
I will close this for now, feel free to reopen it if you would consider supporting comma-separated arbitrary values for margin, padding, grid template ...

PR #2755, opened 2023-06-11
fix(preset-mini): rgb and hsl are not comma separated functions

nothendev:
This pull request fixes preset-mini assuming that `rgb` and `hsl` functions' values should be comma separated (`rgb(189, 73, 120)`), and not space-separated with optional opacity after a slash (`hsl(224 48% 3% / 100%)`).


nothendev:
Oh, I completely forgot about tests

antfu:
Close due to inactive. Feel free to bring it up if you could include tests. Thanks.

Issue #2682, opened 2023-05-29
Add utilities usage info to UnoCSS Inspector 


enkot:
### Clear and concise description of the problem

In WindiCSS Analyzer it was very convenient to see many times each utility is used, so developers can perform additional optimisations (create shortcuts, use inline styles etc.).

<img width="1110" alt="Screenshot 2023-05-29 at 12 39 41" src="https://github.com/unocss/unocss/assets/10506522/8806fe97-e425-4c29-8981-df334e1c02b6">

### Suggested solution

Add another "Analyzer" page (or tab to the "Overview" page) to UnoCSS Inspector with this info:
- top utilities
- utilities usage
- suggested shortcuts
- color palette 

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://uno.antfu.me).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
I currently don't have bandwidth to do so, but PR welcome :)

enkot:
I have started working on it, but currently have one problem. I'm trying to get `currentSelector` value to properly group selectors in Analyzer, but it looks like that `parseToken` produces different result in different contexts. 
Here I printed output of `parseToken()` in `generate()`  function (`packages/core/src/generator/index.ts`):
1. Left - `nr interactive`
2. Right - `nr inspector`

The `RuleContext` is missing in the output from `nr inspector`.

<img width="967" alt="Screenshot 2023-06-01 at 00 46 43" src="https://github.com/unocss/unocss/assets/10506522/7638aa9f-821e-414f-afef-46d664c0cfbb">

Can you please help/give some direction? For sure I'm missing something 🤔 

Another option I see - is to call `matchVariants()` for each selector in the `packages/inspector/node/index.ts`. But it looks like redundant operations, because it already done in the `generate()`.

antfu:
For the additional info you need to turn on https://github.com/unocss/unocss/blob/0d66262d3851db010f7ad6eb8f1fef6e3c9ffaab/interactive/uno.config.ts#L13

enkot:
Hidden gems :) 
Thank you! Thats exactly what I need 👍

PR #2762, opened 2023-06-13
feat(inspector): analyzer

enkot:
Implements #2682

This PR adds utilities usage info to UnoCSS Inspector (under additional `Analyzer` tab), the same way it was done for WindiCSS.
Screenshot 2023-06-14 at 00 18 26
Screenshot 2023-06-14 at 00 12 26
Screenshot 2023-06-14 at 00 12 08

#### Tradeoffs:
- Because UnoCSS uses `Set` as main data structure to add/store tokens (in the core, presets, extractors etc.), new `CountableSet` data structure was introduced. The main idea is that besides actual value it stores count number linked to the value, which increments each time the same value is added to the `CountableSet`.
It is based on the `Map` and has the same API as `Set`, but with additional `getCount` and `setCount` methods. Because of performance reasons `CountableSet` is opt-in and utilised only for Inspector needs. 
☝️  But currently there is a problem that extractors should also "know" about what data structure to use: `Set` or `CountableSet`, which looks like breaking change.
Alternative solution I see is to implement separate extractors and `generate` method only for Analyzer needs. But it feels like an overkill.

#### Problems:
- Because Analyzer uses `parseColor` util from `@unocss/preset-mini/utils` - Inspector package should build after `preset-mini` package, which currently is not the case. The fast workaround was to do this:
```sh
"build": "rimraf packages/*/dist && esno scripts/copy-files.ts && pnpm -r --filter=./packages/* --filter=!./packages/inspector run build && nr build:inspector && pnpm -r run build-post",
``` 
I know it doesn't look good and for sure there is more correct solution 🙈 
Also the design/icons is topic to discuss :)

I would be grateful for a review!


packages/core/src/utils/set.ts
```diff
@@ -0,0 +1,66 @@
+export class CountableSet<K> {
```
antfu, line 1: How about extending `Set`, override the `add` function to count, and introduce an additional function to get the counts?

enkot: Do you mean something like this, right?
E.g:
```ts
export class CountableSet<K> extends Set {
  _map = new Map<K, number>()

  add(key: K) {
    super.add(key)
    this._map.set(key, (this._map.get(key) ?? 0) + 1)
    return this
  }

  getCount(key: K) {
    return this._map.get(key) ?? 0
  }
}
``` 
But in this case Set's values with be duplicated as Map's keys. Not sure if it is a problem 🤔

enkot: and we also need `delete` method:
```ts
delete(key: K) {
  super.delete(key)
  return this._map.delete(key)
}
```

antfu: Yes :)

enkot:
@antfu Reimplemented `CountableSet` to extend `Set`, but there is weird error on `build-post`:

Screenshot 2023-06-17 at 22 44 23

Sometimes `this._map` is `undefined`. It feels like something tries to replace context for `Set` method calls, not sure. 
Trying to investigate that, but will be grateful for your help 🕵️

enkot:
@antfu Maybe lets stick to the initial implementation of `CountableSet`?

packages/core/src/utils/set.ts
avinean: Can we use a simple getter instead of overriding Set methods ?

```ts
class MySet extends Set {
    get lastValue() {
        return [...this][this.size - 1]
    }
}
```

enkot:
@antfu @zyyv Could you please share your thoughts on extending `Set`? - https://github.com/unocss/unocss/pull/2762#issuecomment-1595845904

antfu:
The design looks neat! Thanks. I will need some more time to review the changes to the core and how things are categorized

enkot:
Thanks! 
Currently categories should be specified explicitly. The more universal solution I like more - is adding another `category` field to the `RuleMeta` object.

zyyv:
> Thanks! Currently categories should be specified explicitly. The solution I like more - is adding another `category` field to the `RuleMeta` object.

Agree, this is a simple but complicated work, (such as `autoComplete`), we can use file management temporarily, and then we can send pr separately for update.

antfu:
Screenshot 2023-07-23 at 17 49 57

It seems there is something wrong since the latest pushes, the usage does not show the correct count I think. Would be appreciated if you can have a look.


In addition, I think it could be nice to list which modules are contributing those utilities by providing a popup instead of directly going to the interactive search, similar to this:

image

enkot:
@antfu Done :) Added the popup which shows the list of modules where target utility/color/shortcut is used.
It also allows to copy the name or go to the docs.

Screenshot 2023-08-03 at 22 10 15
Screenshot 2023-08-03 at 22 10 31
Screenshot 2023-08-03 at 22 12 34

antfu:
Thanks a lot, great work! Sorry for taking so long

antfu (approved):

enkot:
Thanks! Great news 🥳 
But aren't we planning to have suggested shortcuts?

packages/core/src/utils/countable-set.ts
```diff
@@ -4,6 +4,8 @@ export class CountableSet<K> extends Set<K> {
   constructor(values?: Iterable<K>) {
     super(values)
     this._map ??= new Map()
+    for (const value of values ?? [])
```
enkot, line 7: @antfu Why do we need this loop? `add()` method will be called twice for each value, because `super()` already does this under the hood.

antfu: Because super does not add to `_map`

antfu:
I feel suggested shortcuts might not be as useful as I originally implemented in Windi, while adding a bunch of complexity. I guess it's better to define those shortcuts from a Design System rather that from the usages as well.

packages/core/src/utils/countable-set.ts
enkot: But it calls our `add()` method which do this, isn't it?

enkot: @antfu 
Screenshot 2023-08-14 at 12 14 52

antfu: We could add some test and remove it if not necessary

Issue #2574, opened 2023-05-05
presetWebFonts/bunny provider unreliable


lmeysel:
### UnoCSS version

0.51.8

### Describe the bug

When using bunny fonts provider in presetWebFonts in some cases the fonts are not downloaded. Switching to google works. As far as I can see it has something to do with the order and/or whitespaces.

### Reproduction

The following works:
```typescript
presetWebFonts({
  provider: 'bunny',
  fonts: {
    sans: ['Acme'],
    mono: ['Roboto Slab'],
  },
});
```

Switching the fonts, however does not work:
```typescript
presetWebFonts({
  provider: 'bunny',
  fonts: {
    sans: ['Roboto Slab'],
    mono: ['Acme'],
  },
});
```

According to the devtools/Network traffic the Acme font is not downloaded, other than Roboto Slab.

### System Info

Browser is Google Chrome 112.0.5615.138
Node 18.16.0

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2766, opened 2023-06-15
fix(preset-web-fonts): fix creation of import URL for `bunny`

arunanshub:
Fixes the issue #2574.

### DIFF
--- a/packages/preset-web-fonts/src/providers/bunny.ts
+++ b/packages/preset-web-fonts/src/providers/bunny.ts
@@ -1,3 +1,26 @@
-import { createGoogleCompatibleProvider } from './google'
+import type { Provider, WebFontsProviders } from '../types'
 
-export const BunnyFontsProvider = createGoogleCompatibleProvider('bunny', 'https://fonts.bunny.net')
+export function createBunnyFontsProvider(
+  name: WebFontsProviders,
+  host: string,
+): Provider {
+  return {
+    name,
+    getImportUrl(fonts): string {
+      const fontFamilies = fonts.map((font) => {
+        const { name, weights = [400], italic } = font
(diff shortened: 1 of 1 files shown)


antfu:
Can you explain what is happening and why this fix would need? I guess it should be bunny's responsibility to align with Google's params?

arunanshub:
> Can you explain what is happening and why this fix would need?

@antfu As highlighted in the linked issue, it simply fixes how the import URL is created. Earlier, it used to create the URL in a Google Fonts compatible manner. However, it seems like Bunny has decided to change how the fonts are loaded. Hence, the Google Fonts compatible 

> I guess it should be bunny's responsibility to align with Google's params?

It does work with Google's params **when executed in a browser**, but strangely refuses to work with UnoCSS when configured via `webFonts.provider`. It might also be related to caching:

https://github.com/unocss/unocss/blob/698a7e68d6dd2be0cc9e4fa7c335fd940776012c/packages/preset-web-fonts/src/index.ts#L58-L74

However I have yet to test out this hypothesis.

The important point here is that it only fails with **Bunny**, not Google. Hence if both are equally compatible with each other, and Google's loader works but Bunny's loader fails, it can be safely said there is something wrong with Bunny's loader.

antfu:
It does sound like a bunny issue isn't it? I am still not very sure if we should keep the complexity on us.

arunanshub:
> It does sound like a bunny issue isn't it? I am still not very sure if we should keep the complexity on us.

While the issue indeed lies with Bunny fonts, it is essential to note that the proposed fix is relatively straightforward and can be implemented within approximately 10 lines of code.

Plus, we should see Bunny as a separate font provider service and a [privacy-respecting competitor to Google Fonts](https://css-tricks.com/bunny-fonts/).

PR #2769, opened 2023-06-16
feat(vscode): support autocomplete with fuzzy search.

chizukicn:
### 1. Support autocomplete with fuzzy search.
Just like [Windicss Intellisense](https://github.com/windicss/windicss-intellisense)

https://github.com/unocss/unocss/assets/16945858/d1c06dcb-94fb-42b6-a3be-3775c8e0dd95

this feature need set match type to be **'fuzzy'**

image

### 2. Add  autocomplete max items configure.
image

For better performance during fuzzy search, we need add a setting for the maximum number of items


packages/autocomplete/src/utils.ts
```diff
@@ -18,3 +18,32 @@ export function searchAttrKey(content: string, cursor: number) {
...
+
+/**
+ * fuzzy search
+ * eg: tct -> [t]ext-[c]en[t]er
+ * @param list
+ * @param keyword
+ * @returns
+ */
+export function searchFuzzy(list: string[], keyword: string) {
+  const regex = new RegExp(keyword.split('').join('.*'))
```
antfu, line 30: Creating a new regex for every change does not sound very efficient to me. Can we use `fzf` or `Fuse.js` to do it?

chizukicn: nice! Fzf can dot it better.

antfu:
Look good! We could have it for now, and later if we make the ordering handles better, we could enable it by default.

For example, this should list `gap-` first
image

antfu (approved):

PR #2773, opened 2023-06-18
fix(preset-icons): add `try/catch` wrap `iconLoader`

### LINKED ISSUE #2670: When using className of preset-icon in React component props, cli unexpectedly exits
### UnoCSS version

0.51.13

### Code
```ts
// test.tsx
import type { FC } from 'react';

type AnyProps = {
  [p: string]: any;
}

const SomeNode: FC<AnyProps> = () => null;

const Test: FC = () => {
  return (
    <SomeNode
      className="w-100px h-100px bg-red"
      icon={ <span className="i-icon:search"/> }
    />
  )
}

export default Test;
```

### config
```ts
// unocss.config.ts
import { defineConfig, presetUno, presetAttributify } from 'unocss';
import presetIcons from '@unocss/preset-icons';
import * as fs from "fs";

export default defineConfig({
  presets: [
    presetAttributify(), // working fine after remove this preset
    presetIcons({
      autoInstall: true,
      collections: {
        'icon': name => fs.readFileSync(`./src/images/icon/${name}.svg`, 'utf-8'),
      }
    }),
  ],
});

```

### command
`npx unocss "src/pages/test.tsx" --out-file ./uno.css --watch`

### Reproduction

https://stackblitz.com/edit/node-uuidvm?file=package.json,README.MD,test.tsx,unocss.config.ts

### System Info

OS: macOS 13.3.1 22E261 arm64
NodeJS：16.20.0
uno: 0.51.13

### Validations

zyyv:
close #2670

### DIFF
--- a/packages/preset-icons/src/core.ts
+++ b/packages/preset-icons/src/core.ts
@@ -66,19 +66,25 @@ export function createPresetIcons(lookupIconLoader: (options: IconsOptions) => P
           iconLoader = iconLoader || await lookupIconLoader(options)
 
           const usedProps = {}
-          if (body.includes(':')) {
-            [collection, name] = body.split(':')
-            svg = await iconLoader(collection, name, { ...loaderOptions, usedProps })
-          }
-          else {
-            const parts = body.split(/-/g)
-            for (let i = COLLECTION_NAME_PARTS_MAX; i >= 1; i--) {
-              collection = parts.slice(0, i).join('-')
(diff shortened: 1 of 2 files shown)


antfu:
I don't think we should shallow the error. Ideally they should be handled upper level.

zyyv:
> I don't think we should shallow the error. Ideally they should be handled upper level. 

Ok, thanks for your reply.

PR #2774, opened 2023-06-18
docs: add support for transformerAttributifyJsx

valcosmos:


zyyv:
I don't know how this helps to improve? I guess the usage instructions for `transformerAttributifyJsx` are enough in the [docs](https://unocss.dev/transformers/attributify-jsx).

examples/next/package.json
```diff
@@ -18,6 +18,7 @@
     "@types/react": "18.2.8",
     "@types/react-dom": "18.2.4",
     "@unocss/postcss": "latest",
+    "@unocss/preset-attributify": "latest",
```
zyyv, line 21: You are use it import from `unocss`, why need install it again?

valcosmos: It is quoted here, but it is not installed, so it does not need to be installed?

image

valcosmos:
> I don't know how this helps to improve? I guess the usage instructions for `transformerAttributifyJsx` are enough in the [docs](https://unocss.dev/transformers/attributify-jsx).

u are right.

PR #2787, opened 2023-06-23
fix(vscode): color preview with !important

### LINKED ISSUE #2708: vscode extension provides wrong color in indicator when class has ! prefix
### UnoCSS version

0.51.12

### Describe the bug

It seems that the vscode extension is showing wrong color in the little indicator, when the ! prefix is added to a unocss class:

image

vs

image

Both applies the right styling, although when the important prefix is present, the vscode extension seems to use a default color for the indicator.

### Reproduction

I think the previous images can describe the problem clearly, although if more information is needed, please let me know!

### System Info

Windows 10, 16GB RAM, i7 6th Gen, WSL2 Ubuntu, Chrome, React + ViteJS

### Validations

vargasmesh:
fix #2708

# Problem

The problem was happening because of the regex in `getCssVariables`

```js
const regex = /(?<key>--\S+?):\s*(?<value>.+?);/gm
```

For the given `str`:
```
/* layer: default */
.\!text-amber {
  --un-text-opacity: 1 !important;
  color: rgba(251, 191, 36, var(--un-text-opacity)) !important;
}
```

When `match.groups?.key` is `--un-text-opacity`, its `match.groups?.value` would be `1 !important`. Finally, the resulting `colorString` is `rgba(251, 191, 36, 1 !important)`, which breaks the CSS and wrongly previews the color:

image

# Solution

Use the following regex to stop the capturing group until it finds either an `!` or `;`
```js
const regex = /(?<key>--\S+?):\s*(?<value>.+?).*[!]{0,1}.*;/gm
```


vargasmesh:
Another solution would be to create a sanitizer/cleaner for `match.groups?.value`. Let me know what is best.

antfu:
I am ok with the regex fix. Can you help adding some unit test for that function? Thanks

vargasmesh:
@antfu done! The function `getCssVariables` is not exportable, so I tested `getColorString` instead.

I had to update `alias.ts` and `tsconfig.json` as well to test it from `tests/utils.test.ts`. Let me know if there is anything else you wanna me to do.

vargasmesh:
Also, I have updated the regex to make it more readable. From
```ts
/(?<key>--\S+?):\s*(?<value>.+?).*[!]{0,1}.*;/gm
```
to
```ts
/(?<key>--\S+?):\s*(?<value>.+?)\s*[!;]/gm
```

antfu (approved):

PR #2789, opened 2023-06-23
perf(autocomplete): better order with fuzzy search.

chizukicn:
image

**Through second search, we can have better ordering**


antfu:
Awesome!

PR #2810, opened 2023-06-29
fix(eslint-plugin): modify sort by alphabetical order instead of code

### LINKED ISSUE #2780: [eslint-plugin] order rule false positive
### UnoCSS version

0.53.1

### Describe the bug

The following vue code:

```vue
<template>
  <a
    class="bg-u-black!"
    disable-resize-watcher
    disable-route-watcher
    :elevation="0"
    permanent
    rail
  >
  </a>
</template>
```

Will report an error `UnoCSS attributes are not ordered (@unocss/order-attributify)`. However, once you add a space and delete it, the error goes away.

### Reproduction

👆

### System Info

ESLint 8.43.0

### Validations

Tzyito:
# Problem

to fixed #2780 
As seen by issue, when we use the eslint plugin and the `unocss.config.js` configures the `presetUno` preset and `shortCuts`, it may cause an abnormal error message in the eslint plugin: `warning uno css attributes are not ordered`.
The source code is sorted according to `attributes code`. As a result, when the user sets short cuts, there will be problems. for example
```javascript
// unocss.config.js
{
    shortCuts: [{
      'center': 'items-center flex',
      'disblock': 'block',
      'bg-my-white': 'bg-white',
    }]
}
// App.vue
<template>
   <div bg-my-white center disblock items-center></div>
</template>
```
In our common sense, it should be sorted alphabetically. But at this time, after eslint fix, it is sorted as `disblock center items-center bg-my-white`(the code list is [640, 'bg-my-white'],[107, 'center'],[95, 'disblock'],[362, 'items-center']). so we get confused

# Suggestion
We always get the sortedlist first, then sort by string. so i made this pr. 
I only see the sort rules method in the eslint plugin, is there anything else that I have missed that needs to be modified? please help me point it out
let me know if I'm missing any information or it's unclear.


antfu (changes requested):
Can we have them in the test instead of a fixture that requires manual triggering?

You can put it in https://github.com/unocss/unocss/blob/19d360defb637071016f55015788d8a3a206ca0a/test/sort-rules.test.ts

Tzyito:
> Can we have them in the test instead of a fixture that requires manual triggering?
> 
> You can put it in https://github.com/unocss/unocss/blob/19d360defb637071016f55015788d8a3a206ca0a/test/sort-rules.test.ts

oh, sure thing @antfu I am very grateful for your reply, is there anything else I need to modify?

test/sort-rules.test.ts
```diff
@@ -18,19 +25,24 @@ describe('sort rules', () => {
     expect(await sort('pt-2 p-4 foo'))
       .toMatchInlineSnapshot('"foo p-4 pt-2"')
     expect(await sort('hover:focus:p1 hover:mt1 hover:m2 pt-2 p-4'))
-      .toMatchInlineSnapshot('"p-4 pt-2 hover:m2 hover:mt1 hover:focus:p1"')
+      .toMatchInlineSnapshot('"hover:focus:p1 hover:m2 hover:mt1 p-4 pt-2"')
```
antfu, line 28: These do not look ideal me. Variants should be after normal rules.

Tzyito: If we sort the codes corresponding to the variants in the preset, will it be more difficult for us to judge how to write class in order?
```javascript
// sort-rules.ts
let sorted = result
    .filter(notNull)
    .sort((a, b) => {
      let result = a[0] - b[0]
      if (result === 0)
        result = a[1].localeCompare(b[1])
      return result
    })
    .map(i => i[1])
    .join(' ')
```
when I set up `shortCuts`, I don't know what the code corresponding to my class is. As a result, I cannot actively modify the sorting, so is it better to sort alphabetically?

antfu: I think the goal of the eslint plugin is to sort the rules by their definition order, so rules with similar scopes are placed together. Otherwise, if we sort alphabetically, you might not need this config to be UnoCSS-specific.

To tackle this, I think we should have a higher sort by the CSS layers, so shortcuts are separate from utilities classes. In the shortcuts layer, we might be able to sort alphabetically.

Tzyito: > I think the goal of the eslint plugin is to sort the rules by their definition order, so rules with similar scopes are placed together. Otherwise, if we sort alphabetically, you might not need this config to be UnoCSS-specific.
> 
> To tackle this, I think we should have a higher sort by the CSS layers, so shortcuts are separate from utilities classes. In the shortcuts layer, we might be able to sort alphabetically.

thank you for your reply. I may not understand what you said. Do you mean we need to pay attention to the writing order in short cuts ?

PR #2814, opened 2023-06-30
fix(icons): loader does not exist is displayed in case of undefined

TalexDreamSoul:
# Problem

image
Please take a look at the image above.

The `uno.config.ts`

``` typescript
import { defineConfig } from "unocss";
import presetIcons from '@unocss/preset-icons'

export default defineConfig({
  presets: [
    ...
    presetIcons({
      ...
    }),
  ],
});
```

When I started the project in the `electron-vite` ([Here](https://github.com/electron-vite)) template repository (Additionally: UnoCSS added), an error occurred!

image

# Solution

I don't know what the problem exactly is, but when I ignore to load `undefined loader` it works and has no influence on functions.

### DIFF
--- a/packages/preset-icons/src/core.ts
+++ b/packages/preset-icons/src/core.ts
@@ -124,7 +124,8 @@ export function createPresetIcons(lookupIconLoader: (options: IconsOptions) => P
 export function combineLoaders(loaders: UniversalIconLoader[]) {
   return (async (...args) => {
     for (const loader of loaders) {
-      if ( !loader ) continue;
+      if (!loader)
+        continue
       const result = await loader(...args)
       if (result)
         return result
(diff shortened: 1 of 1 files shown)


zyyv:
@TalexDreamSoul Can you provide a **minimal** reproduction repo about this issue?

packages/preset-icons/src/core.ts
```diff
@@ -124,6 +124,7 @@ export function createPresetIcons(lookupIconLoader: (options: IconsOptions) => P
 export function combineLoaders(loaders: UniversalIconLoader[]) {
   return (async (...args) => {
     for (const loader of loaders) {
+      if ( !loader ) continue;
```
antfu, line 127:
```suggestion
      if (!loader)
        continue
```

PR #2815, opened 2023-06-30
fix(svelte-scoped): respect passed options

fehnomenal:
While testing why the svelte scoped mode does not work with svelte 4 I noticed two things:
1. The class prefix differs from the documented default.
2. `combine` is forced during `vite build` to be `true` regardless of the passed config option.

This PR fixes both.


jacob-8 (approved):
Thanks, @fehnomenal! You can always ping me to save @antfu time.

@antfu, not sure why the Windows test is having trouble with finding rimraf on build.

PR #2838, opened 2023-07-06
feat(nuxt): Allow the usage of a custom reset for preflight

Kloen:
Following the behavior that's already implemented for Astro, this changes allows the current `preflight` option on Nuxt to define a different reset import.

> Main motivation for this change is that I have a project that was impacted by #2127 and, even though I could use the reset of tailwind-compact from the .vue files instead of having it on the config, I think that having equivalent configuration options between frameworks would be better.

### DIFF
--- a/examples/nuxt3/app.vue
+++ b/examples/nuxt3/app.vue
@@ -11,5 +11,9 @@
 </template>
 
 <style>
+/**
+ * Alternatively, `preflight: true` or `preflight: "@unocss/reset/tailwind.css"` can be used on `nuxt.config.ts`
+ * to achieve the same behaviour
+ */
 @import '@unocss/reset/tailwind.css';
 </style>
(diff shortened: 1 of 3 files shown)


zyyv:
I don't think we need this kind of compatibility processing, we can't guarantee the correctness of string. 
You should install `@unocss/reset` package to use.

Kloen:
> we can't guarantee the correctness of string.

I mean, not like you can do it right now in the `@unocss/astro` package either... Could always be validated I suppose?

> You should install `@unocss/reset` package to use.

Just saying that package is installed when you install `@unocss/nuxt` otherwise the `preflight: true` you can set right now wouldn't work at all, the current default reset for preflight is `@unocss/reset/tailwind.css`, useful and all, but useless option when it comes down to the issue I pointed out on the PR description.

PR #2844, opened 2023-07-08
chore: fix types

chizukicn:
see https://github.com/vueuse/vueuse/commit/4d7577ee


chizukicn:
I want to do some optimization on the playground, but I found a lot of various bugs, I am thinking whether to complete it in one PR, or divide it into a separate PR to submit😊 @antfu

antfu:
@chizukicn Thanks! It's better to do in separate PRs so they could be review and merged separately

PR #2845, opened 2023-07-08
fix(preset-web-fonts): sort weights as string using localeCompare

arunanshub:
Previously, the weights were converted to Number before sorting. This would lead to conversion of string to NaN.

### DIFF
--- a/packages/preset-web-fonts/src/index.ts
+++ b/packages/preset-web-fonts/src/index.ts
@@ -27,7 +27,11 @@ export function normalizedFontMeta(meta: WebFontMeta | string, defaultProvider: 
   if (typeof meta !== 'string') {
     meta.provider = resolveProvider(meta.provider || defaultProvider)
     if (meta.weights)
-      meta.weights = [...new Set(meta.weights.map(Number).sort((a, b) => a - b))]
+      meta.weights = [
+        ...new Set(meta.weights.sort((a, b) => 
+          a.toString().localeCompare(b.toString(), 'en', { numeric: true }))
+        )
+      ]
     return meta as ResolvedWebFontMeta
   }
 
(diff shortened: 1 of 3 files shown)


zyyv (approved):

zyyv:
@arunanshub  Is there any downside to using `map(Number).sort((a,b)=>a-b)`, We are missing a step to judge whether it is a number.

zyyv:
@arunanshub I changed the code a bit, feel free to optimize it. By the way can you add a test case for bunny. Need your review.

arunanshub:
@zyyv In [UnoCSS docs for Web fonts preset](https://unocss.dev/presets/web-fonts#fonts) we should ideally be able to selectively choose the font weights along with their italicized versions. This is currently possible using this configuration:

```ts
{
  // ...
  fonts: {
    sans: 'Roboto',
    mono: ['Fira Mono:400,700i'],
  }
}
```

However, this method is sub-optimal. It is cleaner if we specify the weights like this:

```ts
{
  // ...
  fonts: {
    sans: 'Roboto',
    mono: {
      name: 'Fira Code',
      weights: [400, '700i'],
    },
  }
}
```

However, the current weight sorting function converts all the weight values to numbers, which converts the italicized weights to NaN.

zyyv:
I see, I will revert my changes. Thank you for your guide.

arunanshub:
> @arunanshub I changed the code a bit, feel free to optimize it. By the way can you add a test case for bunny. Need your review.

@zyyv The removal of italicized font weights undermines the objectives of the pull request. Our intention is to sort all font weights, **including** the italicized versions.

zyyv:
I didn't know the weight would carry the italics, thanks again.

zyyv (approved):

PR #2852, opened 2023-07-10
chore: add npm homepage link

warmthsea:
The npm homepage has already displayed the content of README.md, you can add the official website address to the homepage of npm .

image


antfu:
Thanks, but I don't think it's the right place to update it.

Issue #1166, opened 2022-06-25
Webstorm Intellisense support


SnowingFox:
vscode is a nice ide, but there  has a lost of people using others like webstorm, and webstorm only has `tailwindcss` Intellisense, i noticed that unocss has `tailwindcss-preset`, it means i can use tailwindcss synatx to use unocss.
So it's possible to use unocss in webstorm with tailwindcss intellisense?

panzhangguo:
同  现在的问题不只是智能提示，很多unocss的写法直接报错。例如：<i i-icon-park-outline:search></i>；after:flex-auto after:content-none；等

antfu:
I don't use WebStorm myself. Counting on the community to come up with a WebStorm plugin. Later we could consider making it official or so.

Xwudao:
> I don't use WebStorm myself. Counting on the community to come up with a WebStorm plugin. Later we could consider making it official or so.

Webstorm implement Tailwindcss Intellisense because of it's LSP, does unocss has it's language server?

shtse8:
I am in need too. Anyone could provide this so that I don't need to check the interactive doc always.

EvanMeek:
> > 我自己不使用 WebStorm。指望社区想出一个 WebStorm 插件。稍后我们可以考虑将其正式化。
> 
> Webstorm 实现 Tailwindcss Intellisense 因为它提供 LSP，unocss 有语言服务器吗？

我赞成提供一个语言服务器，这样各大编辑器可以写一段简单的json配置即可

antfu:
I am also open to refactoring it into an LSP. PR also welcome on that direction.

SnowingFox:
> I am also open to refactoring it into an LSP. PR also welcome on that direction.

Could you provide the url about unocss LSP package?

agileago:
+1

chz:
+1

meritozh:
@SnowingFox it's a monorepo. https://github.com/unocss/unocss/tree/main/packages/vscode

ouuan:
> it's a monorepo. [`main`/packages/vscode](https://github.com/unocss/unocss/tree/main/packages/vscode?rgh-link-date=2022-10-11T03%3A43%3A48Z)

I think that's a VS Code extension, not an LSP server. No one said there is an LSP package. But there's an issue for LSP #869, which has been closed by the stale bot.

adoin:
[...]

PR #2861, opened 2023-07-12
docs: add jetbrains ide plugin doc

re-ovo:
Add webstorm plugin for unocss

Close https://github.com/unocss/unocss/issues/1166

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -40,7 +40,7 @@ const Integrations: DefaultTheme.NavItemWithLink[] = [
   { text: 'PostCSS', link: '/integrations/postcss' },
   { text: 'ESLint', link: '/integrations/eslint' },
   { text: 'VSCode extension', link: '/integrations/vscode' },
-  { text: 'WebStorm plugin', link: '/integrations/webstorm' },
+  { text: 'JetBrains IDE Plugin', link: '/integrations/jetbrains' },
 ]
 
 const Presets: DefaultTheme.NavItemWithLink[] = [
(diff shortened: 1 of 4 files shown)


zyyv:
Maybe we can temporarily put it in the `community integration` and listen to the feedback.
Add a tab named `community integration` in https://unocss.dev/presets/community.

re-ovo:
> Maybe we can temporarily put it in the `community integration` and listen to the feedback. Add a tab named `community integration` in https://unocss.dev/presets/community.

I know what you mean, but it doesn't seem like this has much to do with "presets"?

docs/.vitepress/config.ts
```diff
@@ -40,6 +40,7 @@ const Integrations: DefaultTheme.NavItemWithLink[] = [
   { text: 'PostCSS', link: '/integrations/postcss' },
   { text: 'ESLint', link: '/integrations/eslint' },
   { text: 'VSCode extension', link: '/integrations/vscode' },
+  { text: 'WebStorm plugin', link: '/integrations/webstorm' },
```
userquin, line 43: It should work in any JetBrains IDE, I'm using IntelliJ Ultime Edition and it is also working. Maybe we can switch the text and the md file to `JetBrains IDE Plugin` and `jetbrains` respectivelly

userquin: ?

re-ovo: ok

userquin: I mean, it should work with any JetBrains IDE since share the logic between IDE, and it should work.

Check Vitest ide entry https://vitest.dev/guide/ide.html#jetbrains-ide, in the history and you'll see the community plugin: https://github.com/vitest-dev/vitest/commit/30306eec3a1ed361f12a3d032cff1517c11d0873 (IntelliJ/WebStorm)

re-ovo: done

userquin (approved):

zyyv (approved):

PR #2862, opened 2023-07-12
made NextJS docs visible in the sidebar

NyllRE:
since the doc is done and it's working I thought why not display it for users so it increases adoption

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Read the [documentation](https://unocss.dev/) for more details.
 
 - [Vite](https://unocss.dev/integrations/vite)
 - [Nuxt](https://unocss.dev/integrations/nuxt)
+- [Next.js](https://unocss.dev/integrations/next)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,6 +51,7 @@ Read the [documentation](https://unocss.dev/) for more details.
 
 - [Vite](https://unocss.dev/integrations/vite)
 - [Nuxt](https://unocss.dev/integrations/nuxt)
+- [Next.js](https://unocss.dev/integrations/next)
```
antfu, line 54: The reason we don't put it here is that unlike the others, we don't provide direct integrations with Next.js. It's more like a guide than an integration

NyllRE: makes sense, well would there be a place that fits a guide for next?

antfu: I think we could have a `Guide` section under https://unocss.dev/integrations/#frameworks-tools

PR #2902, opened 2023-07-22
fix(preset-mini): supplement rule autocomplete

chizukicn:
This PR will have these changed.

1. Supplement rule autocomplete, like `absolute` and `w-full`
image


packages/preset-mini/src/_rules/position.ts
```diff
@@ -1,9 +1,26 @@
...
 
 export const positions: Rule[] = [
-  [/^(?:position-|pos-)?(relative|absolute|fixed|sticky)$/, ([, v]) => ({ position: v })],
+  [/^(?:position-|pos-)?(relative|absolute|fixed|sticky)$/, ([, v]) => ({ position: v }), {
+    autocomplete: [
+      makeGroupAutocomplete([
+        ['position', 'pos'],
+        positionValues,
+      ]),
+      makeGroupAutocomplete([positionValues]),
```
antfu, line 21: I'd actually prefer to have plain string there. For the position values, we could put it in https://github.com/unocss/unocss/blob/995cc6270c167537c7ab70c94696022e5220e813/packages/autocomplete/src/parse.ts#L6

chizukicn: Now we can add a custom shorthands in preset-mini

antfu (approved):

PR #2903, opened 2023-07-22
feat(autocomplete): support custom shorthands

chizukicn:
This PR will support custom shorthands on autocomplete.

```
import { defineConfig } from 'unocss'

export default defineConfig({
  autocomplete: {
    shorthands: {
      size: ['xs', 'sm', 'md', 'lg', 'xl', '2xl', '3xl', '4xl', '5xl', '6xl', 'full'],
    },
  },
})


```


antfu:
I am not sure if I understand. Can you explain a bit what's this is trying to do?

chizukicn:
> I am not sure if I understand. Can you explain a bit what's this is trying to do?

I think since autocomplete can customize extractors and templates, shorthands should also be able to extend and cover.

We can customize shorthands in preset. Although both theme and extractor can achieve this effect. But the theme may interfere with the runtime, and the extractor is too troublesome, which can be more of an option

chizukicn:
eg:
image
image
We can easily customize autocomplete. @antfu

chizukicn:
We can also customize the shorthand in the preset.
```ts
 export default {
    name: "preset-palette",
    autocomplete: {
      shorthands: {
        opacity: `(${Array.from({ length: 21 }, (_, i) => i * 5).join("|")})`
      }
    },
   rules:[
      [
        /^text-(\w+)\/(\d{1,2})$/,
        colorResolver("color"),
        {
          autocomplete: "text-$colors/<opacity>"
        }
      ],
  ]
}
```

Autocomplte:
image

antfu:
OK now I see. Can you also add some documentation for this? Thanks.

chizukicn:
> OK now I see. Can you also add some documentation for this? Thanks.

done~

antfu (approved):

PR #2909, opened 2023-07-24
fix(web-fonts): hoist `@import` in single layer

zyyv:
When building with this warning, i think we can sort the css ahead of time.

```ts
// uno.config.ts
presetWebFonts({
  provider: 'google',
  fonts: {
    ...
  },
  inlineImports: false,
}),
```
image


zyyv:
I don't know if it's better to handle it in `core` or `vite plugin`. 
What do you think ? Or we don't need it at all.

packages/core/src/generator/index.ts
```diff
@@ -298,6 +298,18 @@ export class UnoGenerator<Theme extends object = object> {
       getLayers,
       getLayer,
     }
+
+    function sortImport(css: string) {
```
antfu, line 302:
```suggestion
    function hoistAtImport(css: string) {
```

packages/core/src/generator/index.ts
```diff
@@ -280,15 +280,15 @@ export class UnoGenerator<Theme extends object = object> {
       }
 
       const layerMark = minify ? '' : `/* layer: ${layer} */${nl}`
-      return layerCache[layer] = css ? layerMark + css : ''
+      return layerCache[layer] = sortImport(css ? layerMark + css : '')
     }
 
     const getLayers = (includes = layers, excludes?: string[]) => {
-      return includes
+      return sortImport(includes
```
antfu, line 287: But honestly, I don't think this is the best way to do so, as this would affect every generation for a relevantly rare use case.

I think it might be better to have a separated layer for the `@import` to be generated to

zyyv:
Done. 
Extracte css with `@import` preflight's  and put it in a new layer named `@imports`.

packages/core/src/generator/index.ts
```diff
@@ -177,18 +177,35 @@ export class UnoGenerator<Theme extends object = object> {
...
         await Promise.all(Array.from(preflightLayerSet).map(
           async (layer) => {
             const preflights = await Promise.all(
               this.config.preflights
                 .filter(i => (i.layer || LAYER_PREFLIGHTS) === layer)
-                .map(async i => await i.getCSS(preflightContext)),
+                .map(async (i) => {
+                  const css = await i.getCSS(preflightContext)
+                  return css ? hoistAtImport(css) : undefined
+                }),
```
antfu, line 208: Should put in the rule, not the core, as it, again, coupled and affect every generation

packages/core/src/constants.ts
```diff
@@ -1,8 +1,10 @@
 export const LAYER_DEFAULT = 'default'
 export const LAYER_PREFLIGHTS = 'preflights'
 export const LAYER_SHORTCUTS = 'shortcuts'
+export const LAYER_IMPORTS = '@imports'
 
 export const DEFAULT_LAYERS = {
+  [LAYER_IMPORTS]: -1000,
```
antfu, line 7:
```suggestion
  [LAYER_IMPORTS]: -200,
```

packages/core/src/constants.ts
```diff
@@ -1,8 +1,10 @@
 export const LAYER_DEFAULT = 'default'
 export const LAYER_PREFLIGHTS = 'preflights'
 export const LAYER_SHORTCUTS = 'shortcuts'
+export const LAYER_IMPORTS = '@imports'
```
antfu, line 4:
```suggestion
export const LAYER_IMPORTS = 'imports'
```

Maybe in the context, `imports` are already explicit enough?

packages/core/src/generator/index.ts
zyyv: You mean, when writing them, put them into the import layer actively?
```ts
prelights: [
 {
   layer: 'imports',
   getCSS: ()=> `@import ....`
 }
]
```

> coupled and affect every generation

Although it is a bit coupled, uno can correct it in the final stage by Regexp, and extract the correct import css to the layer. Make it a default behavior? 😅

antfu:
Pushed the changes. In some CSS parsers, `@import` can be everywhere. And when a rule emits a raw string, UnoCSS assumes the syntax error etc are handled correctly by the rule itself. To not baby-sitting too much is one of the factors that keep UnoCSS fast.

If the change looks good to you, we could merge it and proceed.

antfu (approved):

zyyv:
> To not baby-sitting too much is one of the factors that keep UnoCSS fast.

Thanks your guide, keep it in mind.🥰

PR #2922, opened 2023-07-27
fix(svelte-scoped): avoid race conditions in shortcuts extending

### LINKED ISSUE #2818: Styles sometimes don't apply with Svelte Scoped and Svelte 4
### UnoCSS version

0.53.4

### Describe the bug

There are some edge cases when using Svelte 4 where styles are not generated for some components. I have yet to find the source of the issue but managed to provide a repro showcasing the problem; *extracted and stripped down from the real app*.

Currently, a workaround is to downgrade Svelte to 3.59.2.

But here comes more weird part. The diffs below are a few examples of changes that make the issue disappear in the repro.

```diff
diff --git a/src/routes/technical-studies.section.svelte b/src/routes/technical-studies.section.svelte
index b40e081..27c676c 100644
--- a/src/routes/technical-studies.section.svelte
+++ b/src/routes/technical-studies.section.svelte
@@ -6,9 +6,6 @@
 		<h4 class="text-primary text-lg font-bold tracking-wide">
 		</h4>
 
-		<h3 class="my-4 text-4xl font-bold text-gray-100">
-		</h3>
-
 		<h4 class="my-4 text-xl leading-relaxed text-gray-100">
 		</h4>
 
```

```diff
diff --git a/src/routes/+page.svelte b/src/routes/+page.svelte
index fa56246..4e581e4 100644
--- a/src/routes/+page.svelte
+++ b/src/routes/+page.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 	import EnvironmentalRegulationSection from "./environmental-regulation.section.svelte";
-	import TechnicalStudiesSection from "./technical-studies.section.svelte";
 </script>
 
 <EnvironmentalRegulationSection />
```

### Reproduction

https://github.com/scheatkode/unocss-weirdness-repro

#### Expected

image

#### Actual

image

### System Info

_No response_

### Validations

scheatkode:
Closes #2818.


jacob-8:
Thanks for the patience, will review soon though should be a simple check. That line was needed originally a few iterations back and very well may be completely unnecessary now.

scheatkode:
No worries. This is but a simple fix that I'm concerned may introduce performance issues, noticeably in larger apps because of the accumulated memory and added lookup times.

Going the immutable route should prevent such issues, something along these lines:

```diff
diff --git a/packages/svelte-scoped/src/_preprocess/index.ts b/packages/svelte-scoped/src/_preprocess/index.ts
index 21559d92..47ae5e21 100644
--- a/packages/svelte-scoped/src/_preprocess/index.ts
+++ b/packages/svelte-scoped/src/_preprocess/index.ts
@@ -12,12 +12,9 @@ export function UnocssSveltePreprocess(options: UnocssSveltePreprocessOptions =
   if (!options.classPrefix)
     options.classPrefix = 'usp-'
 
-  let uno: UnoGenerator
-
   return {
     markup: async ({ content, filename }) => {
-      if (!uno)
-        uno = await getGenerator(options.configOrPath, unoContextFromVite)
+      const uno = await getGenerator(options.configOrPath, unoContextFromVite)
 
       if (isViteBuild && options.combine === undefined)
         options.combine = isViteBuild()
@@ -49,8 +46,7 @@ export function UnocssSveltePreprocess(options: UnocssSveltePreprocessOptions =
       if (!changeNeeded)
         return
 
-      if (!uno)
-        uno = await getGenerator(options.configOrPath)
+      const uno = await getGenerator(options.configOrPath, unoContextFromVite)
 
       let preflightsSafelistCss = ''
       if (addPreflights || addSafelist) {
```

If the generator instantiation itself is heavy, [`klona`](https://www.npmjs.com/package/klona) might be of use.

What do you think ?

jacob-8 (approved):

jacob-8:
The fix looks good to me in my testing.

> I'm concerned may introduce performance issues, noticeably in larger apps because of the accumulated memory and added lookup times.

I'm inclined to say we merge the fix first and then optimize if it is indeed leading to issues. I have some larger apps that I use this in and can try things out. But I'm not familiar with the internals of how shortcuts work. I assume lookups would be done efficiently, as in a Map, when it's negligible if the Map gets large. @antfu do you foresee problems in large apps if each Svelte component adds its chunk of compiled styles (`_bg-red-500_vxas4i` = `bg-red-500`...) to `uno.config.shortcuts` and leaves them there? Should we optimize this with a separate uno generator for each Svelte component?

antfu:
It shouldn't be too heavy to spawn new generators. Meanwhile, I think it's ok to leave the shortcuts as long as they don't infer with each other. We did that in the compile mode:
https://github.com/unocss/unocss/blob/d44d745f97c0d4c55dd45a625992d54235aadcda/packages/transformer-compile-class/src/index.ts#L119-L122

scheatkode:
> I'm inclined to say we merge the fix first and then optimize if it is indeed leading to issues.

I'm in favor of this as well. The above is just me voicing some concerns as I have spent a bit of time debugging this issue, I think I have a basic grasp on how the generator works.

You're right in that the generator uses a `Map` under the hood for caching lookups, but that's of little use to `svelte-scoped` since shortcuts are short lived. The generator's source of truth is `.config.shortcuts`, a tuple that keeps growing as `svelte-scoped` generates *a lot* of classes.

Then again, this is in no way an actionable comment, just thinking out loud and voicing concerns that may be baseless. Time will tell.

Thank you both for your time !

PR #2925, opened 2023-07-30
fix(core): move magic-string to dependencies

vonagam:
`magic-string` is [imported](https://github.com/unocss/unocss/blob/ab78622a278d3d0432d95f927f1564019320d9b3/packages/core/src/utils/variantGroup.ts#L1) in `utils/variantGroup.ts` so should be listed as a dependency.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,8 +40,10 @@
     "build": "unbuild",
     "stub": "unbuild --stub"
   },
+  "dependencies": {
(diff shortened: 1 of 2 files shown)


antfu:
Oh this wasn't intended. We need to remove the deps on it instead.

antfu:
Fixed in ba77578318ba0c163990b8b948a545c3ac36d31b, thanks

PR #2932, opened 2023-08-01
fix(preset-attributify): add position type

### LINKED ISSUE #2931: AttributifyAttributes don't have `left` type
### UnoCSS version

0.54.1

### Describe the bug

image

### Reproduction

https://github.com/Debbl/unocss-type-test

### System Info

- MacOS
- Chrome
- VsCode

### Validations

zyyv:
close #2931


antfu:
JSX already has some performance issues as the combination gets bloated. I'd recommend using `pos="top-1"` instead of `top="1"`.

PR #2948, opened 2023-08-05
feat(cli)!: add `--write-transformed` option, disable source write back by default

brofrain:
Hello,
this humble PR adds ~~`--update-source`~~ `--write-transformed` flag for `@unocss/cli` used to toggle overwriting source files with transformed code. ~~It is enabled by default to preserve the current behavior, but can be disabled in case another tool takes care of transforming utility classes on the HTML side.~~

Background:
I'm trying to make UnoCSS work with Rust frameworks preferably without trade-offs like dropping Variant Groups. I've made a little [macro](https://github.com/brofrain/unocss-variant-group-transformer-rs) for that purpose and hopefully it will evolve someday to be executed on a WASM bundler side rather than polluting source code.
Regardless of the final approach to rendering valid HTML classes, keeping the source files not-transformed is desirable. Otherwise the Variant Groups feature isn't really much of use :)


packages/cli/src/cli-start.ts
```diff
@@ -21,6 +21,7 @@ export async function startCli(cwd = process.cwd(), argv = process.argv, options
     })
     .option('-c, --config [file]', 'Config file')
     .option('-w, --watch', 'Watch for file changes')
+    .option('--update-source', 'Update source files with transformed utilities', { default: true })
```
antfu, line 24: I feel this was a dangerous default we picked. I'd be happy to name it as `--write-transformed` and default to `false`.

brofrain: Thanks!
I've updated the flag and made sure that files transformed with directives transformer are also covered by tests.

antfu (approved):

lvjiaxuan:
Would it be better to output transformed files instead of overwriting source files?

PR #2957, opened 2023-08-07
fix: remove export types in eslint-plugin

userquin:
build.config file has declaration = false


antfu:
Oh thanks, I just removed it already

PR #2959, opened 2023-08-07
chore: fix build

userquin:
For some reason it seems tsc is using node10 resolution, this PR just excludes eslint-plugin from root typecheck script and includes a new typecheck script in eslint-plugin (that will run from root).


userquin:
It seems @cyberalien found the problem...

cyberalien:
> It seems @cyberalien has found the problem...

Unfortunately no. 

I was working with slightly outdated branch, updating `esno` fixed issue. Then noticed branch was outdated, pulled latest, which already has latest `esno` and it works without issues.

PR #2978, opened 2023-08-12
fix(astro): fix astro hmr warning

### LINKED ISSUE #2812: HMR weirdness with Astro + Uno
### UnoCSS version

0.53.4

### Describe the bug

There are a couple of issues I have noticed trying to use the latest versions of Astro and Uno. I'm not sure if all these problems are caused by Uno, downgrading to Astro ^2.6 (from 2.7.2) fixes it. But I'm not sure if Astro is at fault here.

1. Running dev mode, the first thing you see the web page is completely unstyled. Refreshing fixes it.

2. Uno logs "[unocss-hmr] failed to get unocss hash, hmr might not work"

3. Now, changing some classes, the site sometimes reverts to the unstyled version (I think this happens when a previously unused class is used). Using one that has been used before is fine.

Production build is fine.

### Reproduction

https://github.com/marvin-j97/astro-uno-hmr

### System Info

Ubuntu, Chrome

### Validations

GavinXue:
`return ''` will not generate __vite_css, and then caused hmr warnning.([unocss-hmr] failed to get unocss hash, hmr might not work)

This will fix #2812

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -93,7 +93,7 @@ function AstroVitePlugin(options: AstroVitePluginOptions): Plugin {
          */
         && !id.startsWith(nodeModulesPath)
       )
-        return ''
+        return null
     },
   }
 }


Dunqing:
Hello, `return null` will break https://github.com/unocss/unocss/issues/2513. We need a better way to solve this problem

GavinXue:
> Hello, `return null` will break #2513. We need a better way to solve this problem

I have tried some hours, but failed. Maybe it is better to let someone who fimiliar with this code base to handle it.
Here is some informations maybe help:
- `return ''` will empty `/__uno.css?uno-with-astro-key` css. This avoid injecting twice. But the global devplugin `unocss:global:post` depends on the generated css. The warning caused by the empty css.
- It seems that empty the content will not remove`/__uno.css?uno-with-astro-key` in <head>, Maybe it is better to remove one other than emtpy one (if we can get there.).
- I am not sure the hmr warning is truely affect hmr. if not, maybe we can just disable the astro hmr warnning.

Dunqing:
Thank you for all your hard work!

> * I am not sure the hmr warning truely affect hmr. if not, maybe we can just disable the astro hmr warning.

Yes, this is just a warning and has no real effect on hmr.

I'll take the time to fix this later.

Issue #2987, opened 2023-08-15
HMR with Astro + Uno


LiMao00:
### UnoCSS version

"@unocss/astro": "^0.55.0",

### Describe the bug

Whether following the steps on the official website (https://unocss.dev/integrations/astro) or the steps recommended by astro (https://www.elian.codes/blog/23-02-11-implementing-unocss-in-astro/), the integrated unocss has problems. During the initial load, the computer CPU will surge by about 20-30%, the tab page will continue to spin, and it will take about 10 seconds to finally load. In src\pages\index.astro, if you modify `<h1 class="bg-yellow">` to `<h1 class="bg-green">`, you will also encounter the same issues: CPU surge, tab page spinning, and waiting for about 10 seconds to load completely.

### Reproduction

https://gitee.com/limao-xs/del-astro

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

userquin:
/cc @Dunqing 

It seems the load hook returning empty string when `uno-with-astro-key` present, even when requesting `/__uno.css?uno-with-astro-key`, the HMR warning is there since the `__vite_css` is empty.

Excluding `__uno.css` here https://github.com/unocss/unocss/blob/main/packages/astro/src/index.ts#L88 seems to work.

![imagen](https://github.com/unocss/unocss/assets/6311119/428ce07f-95cb-4218-9647-4053b0b4d295)

![imagen](https://github.com/unocss/unocss/assets/6311119/463874fc-d9c7-4a84-b756-ea8f64fb871e)

Dunqing:
> It seems the load hook returning empty string when `uno-with-astro-key` present, even when requesting `/__uno.css?uno-with-astro-key`, the HMR warning is there since the `__vite__css` is empty.
> 
> Excluding `__uno.css` here https://github.com/unocss/unocss/blob/main/packages/astro/src/index.ts#L88 seems to work.

[...]

PR #2989, opened 2023-08-15
fix(astro): astro hmr warning and improve performance

Dunqing:
fix: #2987 
fix: #2812

re-implement: #2593 
We just align `data-astro-dev-id` with `data-vite-dev-id`

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -1,13 +1,13 @@
-import { resolve } from 'node:path'
+import { join, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import type { AstroIntegration } from 'astro'
 import type { VitePluginConfig } from '@unocss/vite'
 import VitePlugin from '@unocss/vite'
 import type { UserConfigDefaults } from '@unocss/core'
 import type { Plugin, ResolvedConfig } from 'vite'
+import { RESOLVED_ID_RE } from '../../shared-integration/src/layers'
 
 const UNO_INJECT_ID = 'uno-astro'
-const UNO_QUERY_KEY = 'uno-with-astro-key'
 const astroCSSKeyRE = /(\?|\&)lang\.css/
 
 interface AstroVitePluginOptions {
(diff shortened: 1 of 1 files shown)


packages/astro/src/index.ts
```diff
@@ -1,13 +1,13 @@
-import { resolve } from 'node:path'
+import { join, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import type { AstroIntegration } from 'astro'
 import type { VitePluginConfig } from '@unocss/vite'
 import VitePlugin from '@unocss/vite'
 import type { UserConfigDefaults } from '@unocss/core'
 import type { Plugin, ResolvedConfig } from 'vite'
+import { RESOLVED_ID_RE } from '../../shared-integration/src/layers'
```
userquin, line 8: Uhmm, shouldn't use shared package?

Dunqing: > Uhmm, shouldn't use shared package?

I'm not sure, the same usage is found in the vite plugin

packages/astro/src/index.ts
```diff
@@ -19,18 +19,24 @@ function AstroVitePlugin(options: AstroVitePluginOptions): Plugin {
...
+      root = config.root
       command = config.command
-      nodeModulesPath = `${config.root}/node_modules`
     },
     async resolveId(id, importer) {
+      if (RESOLVED_ID_RE.test(id)) {
+        // https://github.com/withastro/astro/blob/087270c61fd5c91ddd37db5c8fd93a8a0ef41f94/packages/astro/src/core/util.ts#L91-L93
+        // Align data-astro-dev-id with data-vite-dev-id to fix https://github.com/unocss/unocss/issues/2513
+        return join(root, id)
+      }
```
Dunqing, line 38: Can we move this logic to vite plugin?

packages/astro/src/index.ts
antfu: It's ok to have this, they file will be inline and bundled with this package

antfu (approved):
LGTM

userquin (approved):

PR #2992, opened 2023-08-16
fix(vscode): registerRoot calling

Thehrz:
In #2979, `registerRoot` was placed in the callback of the reload command, which required us to run the reload command in v0.55.1 in order to load the context. This may lead some users to think that the unocss vscode extension is not working.
This PR rolled back the behavior before v0.55.1.

### DIFF
--- a/packages/vscode/src/index.ts
+++ b/packages/vscode/src/index.ts
@@ -78,9 +78,10 @@ export async function activate(ext: ExtensionContext) {
     : projectPath
 
   try {
+    const contextLoader = await registerRoot(ext, status, cwd)
+
     ext.subscriptions.push(
       commands.registerCommand('unocss.reload', async () => {
-        const contextLoader = await registerRoot(ext, status, cwd)
         log.appendLine('🔁 Reloading...')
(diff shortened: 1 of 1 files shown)


KraXen72:
please also check, if this fixes files with `/* @unocss-include */` not loading. as you said, i have to run reload unocss command, to load the extension, but i also have to manually remove and add this for unocss to register the files. would be great if it was fixed in one pr!

Thehrz:
> please also check, if this fixes files with `/* @unocss-include */` not loading. as you said, i have to run reload unocss command, to load the extension, but i also have to manually remove and add this for unocss to register the files. would be great if it was fixed in one pr!

Hi, I checked the `@unocss-include` magic comment and it seems to work, can you please describe more? Or you can open a new issue.

KraXen72:
sorry, i tested it and it works now. i was using the svelte-scoped preset which does not have config hmr, so it wasn't working. but with normal unocss/vite, it works. thanks.

Thehrz:
fixed in #3005, close it.

PR #2994, opened 2023-08-17
fix: include typings for min-h min-w in preset-attributify

Hrishikesh-K:
Using types for attributes as explained here: https://unocss.dev/presets/attributify#attributify-with-prefix, threw errors when using `min-h` and `min-w`. Those attributes are not included in the list. This PR adds those.


antfu:
Thank you. However, it's already a known issue that this typing slows down the TypeScript checker significantly. I would avoid adding more attributes as it increases the combinations exponentially. I'd suggest you extend it on your local, or use `size="w-1 max-w-2"`

PR #3020, opened 2023-08-22
feat: added animated unocss logo as loader in interactive docs

therealokoro:
This pull request aims to replace the default nuxt spa loading indicator for with a custom one in the [interactive docs](https://unocss.dev/interactive/?s=animate)

Replaced with an animated unocss logo, makes it look more brandable if you know what I mean 😏😏


therealokoro:
looking at the preview I see that they delay is a bit too long

antfu:
Haha look cool! I was going to say we could just disable it, but since you already implemented it and it seems quite interesting - let's have it.

One thing tho, can we make it dark more aware, and also respect reduced-motion? Thanks

therealokoro:
Made a little tweak. This time I promise it works😅

PR #3021, opened 2023-08-22
fix(core): enable details for analyzer

enkot:
https://github.com/unocss/unocss/issues/2682

Currently Analyzer shows incorrect info, because `config.details` is disabled by default.


antfu:
I don't actually mind turning `details` on by default on dev mode. This PR's implementation feels a bit verbose and duplicated

enkot:
But in this case user will be able to break the Analyzer by setting "details: false" in config.

packages/core/src/config.ts
```diff
@@ -157,6 +158,7 @@ export function resolveConfig<Theme extends object = object>(
     extractors,
     safelist: getMerged('safelist'),
     separators,
+    details: config.details ?? process.env.NODE_ENV === 'development',
```
antfu, line 161: The core is env agnostic, we shouldn't rely on process. Might be better to rely on envMode and move this logics to the vite/webpack plugin

enkot:
@antfu 3rd attempt :)

antfu (approved):

PR #3028, opened 2023-08-24
feat(runtime): custom `inject` function

exside:
First of all, amazing work with UnoCSS, keep it up!

I'm liking runtime, but one thing I'm missing is the ability to control where the generated styles get injected into the DOM, right now they are basically even before the `head` tag. I'm using it in a context where i'd like to override existing classes from HTML markup I don't control via UnoCSS `shortcuts`, but I can't (without !important) because the styles from that platform are in the `head` (as usual) and single class selector's specificity is therefore higher than single class shortcuts from UnoCSS defined/injected earlier on.

image

So the change is relatively simple, it's just a user configurable function that receives the style element and the user can then inject that wherever they want. Default behaviour doesn't change. Layering logic is untouched, the function just controls where the first layer is injected.

One can now do the following:

```
	window.__unocss = {
		...,
		runtime: {
			inject: (styleElement) => document.head.append(styleElement),
		},
	};
```

which results in:

image

If you don't like the way it's implemented or where, let me know, glad to adapt to your liking. Just threw this together for myself, but thought it might be helpful for others too.


chu121su12:
The insertion before `<head>` was done so that unocss styles can be overriden by other library/styles that usually appear in the `<head>` or even `<body>`.

Per your example, 
image
The first `<style>` element can be overriden by unocss, especially if you are using reset/preflights.

You can get each layer's `<style>` element with the `update()` method and reposition them manually. See test example below.

https://github.com/unocss/unocss/blob/1073beac99d637c30604b42bf5aa843b0f576f2f/test/runtime-dom.test.ts#L32-L42

packages/runtime/src/index.ts
```diff
@@ -161,15 +162,15 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
...
 
       if (previousLayer == null) {
-        html().prepend(styleElement)
+        inject(styleElement)
       }
       else {
         const previousStyle = getStyleElement(previousLayer)
         const parentNode = previousStyle.parentNode
         if (parentNode)
           parentNode.insertBefore(styleElement, previousStyle.nextSibling)
```
chu121su12, line 171: This should be handled. Maybe use `previousStyle`/`previousLayer` for context

packages/runtime/src/index.ts
```diff
@@ -122,6 +122,7 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
 
   runtimeOptions.configResolved?.(userConfig, userConfigDefaults)
   const uno = createGenerator(userConfig, userConfigDefaults)
+  const inject = (styleElement) => runtimeOptions.inject ? runtimeOptions.inject(styleElement) : html().prepend(styleElement)
```
chu121su12, line 125: Please add the type and some description

packages/runtime/src/index.ts
```diff
@@ -161,15 +162,15 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
       styleElements.set(layer, styleElement)
 
       if (previousLayer == null) {
-        html().prepend(styleElement)
+        inject(styleElement)
```
chu121su12, line 165: You may want to send `layer` for context

exside:
> The insertion before <head> was done so that unocss styles can be overriden by other library/styles that usually appear in the <head> or even <body>.

absolutely agree! This is the normal use case, but in mine I want to use Uno CSS within a third party vendor platform that does a lot of things i can't control, including markup & styles etc., and here the issue is that I DON'T want the platform's styles to override my utils (or shortcuts) defined in Uno, of course I can use `!important` everywhere, but we all know that's bad practice and leads to lots of other issues, so simply being able to control where Uno's style elements are placed within the DOM can solve that problem.

packages/runtime/src/index.ts
exside: Not sure I undestand, I'm simply replacing the hardcoded `html()` with an optional custom method that prepends/appends the style element? What do you mean by "should be handled"? (I mean if a runtime user uses this config option he/she should know what it does IMHO 😉 )

exside: Ah maybe i get what you mean: You think we should give users the ability to inject each layer separately wherever they want it?

packages/runtime/src/index.ts
exside: Added to RuntimeOptions interfarce

chu121su12 (approved):

packages/runtime/src/index.ts
```diff
@@ -26,6 +26,10 @@ export interface RuntimeOptions {
    * Callback to modify config
    */
   configResolved?: (config: UserConfig, defaults: UserConfigDefaults) => void
+  /**
+   * Optional function to control UnoCSS style element(s) injection into DOM.
```
antfu, line 30:
```suggestion
   * Optional function to control UnoCSS style element(s) injection into DOM.
   * When provided, the default injection logic will be overridden.
```

Did I understand it right?

exside: yes! Well, maybe you can tell better if I missed something (see the comments about layer context etc. above)? But the initial DOM injection can be controlled like that and so far it's working great for me, I'm already using this in production and it does what I intended it to do. If there are more advanced use cases I didn't encounter yet, I'm glad to adapt to users needs.

antfu (approved):

PR #3029, opened 2023-08-24
feat(preset-mini): Support `max`  breakpoint

onlyfortesting:
TailwindCSS uses `max-` instead of `lt-` so it would be nice if we also support that. It helps users coming from TailwindCSS too.

https://tailwindcss.com/docs/responsive-design#targeting-a-breakpoint-range


zyyv:
@onlyfortesting  Can you add a test case?

onlyfortesting:
@zyyv good now?

antfu (approved):

PR #3030, opened 2023-08-24
Allow runtime mutation observer optimization

exside:
This is probably a more advanced usage config, but as it doesn't affect default behaviour I think it's a good addition with minimal bundle size impact.

The runtime's MutationObserver is basically watching the entire DOM for all mutations on all nodes and attributes, that's probably fine for most, but I would prefer if I could fine tune Uno CSS runtime so it does the least amount of work necessary, therefore I propose an addition the the runtime options, which allow granular control over:

a) what area of the DOM is monitored by returning a node to watch (including its child nodes) with `target`
b) which attributes to watch for mutations by providing an `attributeFilter` array

both config options are of course optional, default behaviour stays exactly as it is now and are summarized under `observer`

So now we could do this:

```
		...,
		runtime: {
			observer: {
				target: () => document.querySelector('.your-area-of-the-site-to-watch-for-mutations'),
				attributeFilter: ['class'], // only watch for changes in class attribute not all attributes
			},
		},
		...
```

### DIFF
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -251,13 +251,14 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
   function observe() {
     if (observing)
       return
-    const target = html() || defaultDocument.body
-    if (!target)
+    const target = () => runtimeOptions.observer?.target ? runtimeOptions.observer.target() : (html() || defaultDocument.body);
+    if (!target())
       return
-    mutationObserver.observe(target, {
+    mutationObserver.observe(target(), {
       childList: true,
       subtree: true,
       attributes: true,
+      ...(runtimeOptions.observer?.attributeFilter ? { attributeFilter: runtimeOptions.observer.attributeFilter } : {}),
     })
     observing = true
   }


packages/runtime/src/index.ts
```diff
@@ -251,13 +251,14 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
   function observe() {
     if (observing)
       return
-    const target = html() || defaultDocument.body
-    if (!target)
+    const target = () => runtimeOptions.observer?.target ? runtimeOptions.observer.target() : (html() || defaultDocument.body);
```
chu121su12, line 254: Maybe we don't need `target()` as function?

exside: removed the `target()` as a function, now a regular `const`

Tried to add the type defs, hope it's ok 😬

chu121su12 (approved):

packages/runtime/src/index.ts
```diff
@@ -251,13 +267,14 @@ export default function init(inlineConfig: RuntimeOptions = {}) {
...
       return
-    const target = html() || defaultDocument.body
+    const target = runtimeOptions.observer?.target ? runtimeOptions.observer.target() : (html() || defaultDocument.body)
     if (!target)
       return
     mutationObserver.observe(target, {
       childList: true,
       subtree: true,
       attributes: true,
+      ...(runtimeOptions.observer?.attributeFilter ? { attributeFilter: runtimeOptions.observer.attributeFilter } : {}),
```
ydcjeff, line 277: I think we can directly use `imeOptions.observer?.attributeFilter` since `attributeFilter` is optional by default ?

exside: It is optional, but `runtimeOptions.observer?.attributeFilter` will be undefined, and the MutationObserver options are an object which we are extending if required, afaik I can't just "spread" undefined, or maybe I'm not understanding correctly what you mean?

exside: @ydcjeff I think that's indeed what you meant, I tried and it works, if `attributeFilter: undefined` it monitors all attributes, simplified the code per your suggestion! Thanks for checking and giving great feedback, reduced bytes added to bundle!

ydcjeff (approved):
Generally LTGM.
Need to wait antfu for final review.

PR #3039, opened 2023-08-28
feat(autocomplete): add percentage shorthands

zyyv:
Hi, @userquin. We already have percentage autofill, but without %, it has more usages such as `op-50` , `op-<percent>`. I don't know if percentage is the right key name.


zyyv:
Feel free to close the pr if you don't like it.

test/countable-set.test.ts
```diff
@@ -13,13 +13,7 @@ describe('CountableSet', () => {
...
     `)
 
-    expect(s._map).toMatchInlineSnapshot(`
-      Map {
-        "bar1" => 1,
-        "bar2" => 2,
-        "bar3" => 1,
-      }
-    `)
+    expect(s._map).toMatchInlineSnapshot('Map {}')
```
antfu, line 16: Why would this get affected?

antfu (commented):
Yeah I think I'd prefer this approach more.

packages/preset-mini/src/_rules/align.ts
```diff
@@ -19,7 +19,13 @@ const verticalAlignAlias: Record<string, string> = {
 }
 
 export const verticalAligns: Rule[] = [
-  [/^(?:vertical|align|v)-([-\w]+%?)$/, ([, v]) => ({ 'vertical-align': verticalAlignAlias[v] ?? v }), { autocomplete: `(vertical|align|v)-(${Object.keys(verticalAlignAlias).join('|')}|<num>)` }],
+  [/^(?:vertical|align|v)-([-\w]+%?)$/, ([, v]) => ({ 'vertical-align': verticalAlignAlias[v] ?? v }), {
+    autocomplete: [
+      `(vertical|align|v)-(${Object.keys(verticalAlignAlias).join('|')})`,
+      '(vertical|align|v)-<num>',
```
userquin, line 25: The `<num>` in vertical-align without unit makes no sense

test/countable-set.test.ts
zyyv: Just run `test:update`.

zyyv: What should i do for reset the test case? 😂

test/__snapshots__/autocomplete.test.ts.snap
```diff
@@ -20,7 +20,7 @@ exports[`autocomplete > should accept variants 1`] = `
 
 exports[`autocomplete > should provide autocomplete 1`] = `
 {
-  "align-": "align-base align-baseline align-bottom align-btm align-end align-inherit align-initial align-mid align-middle align-revert",
```
userquin, line 23: Why removing this? These values are valid.

zyyv: Not removed.😂

userquin (approved):

test/autocomplete-parse.test.ts
```diff
@@ -64,21 +64,19 @@ describe('autocomplete-parse', () => {
   })
 
   it('shorthands', () => {
-    const parsed = parseAutocomplete('(m|p)<directions>-<num>')
+    const parsed = parseAutocomplete('(m|p)<directions>-<percentage>')
```
antfu, line 67: Can we add another entry instead of overriding the existing ones?

test/__snapshots__/autocomplete.test.ts.snap
```diff
@@ -20,7 +20,7 @@ exports[`autocomplete > should accept variants 1`] = `
...
+  "align-": "align-10% align-100% align-20% align-30% align-40% align-50% align-60% align-70% align-80% align-90%",
```
antfu, line 23: I am a bit unsure about this. Maybe we need a priority system, as I think the percentage usage should have a lower priority than the semantic ones.

zyyv: This is exactly what I am worried about, please allow me to deal with its priority in another PR.

test/countable-set.test.ts
zyyv: It's funny, the test output may be related to the OS.
Before: inter i5
After: M1 pro

Issue #3014, opened 2023-08-21
VSCode extension fails on ESM imports (Deno)


adamgreg:
### UnoCSS version

0.55.1

### Describe the bug

The VS Code extension does not work with Deno if there is an ESM import in the config file.

For example, if I import `presetUno` from esm.sh, in the Runtime Status I see a message like `Cannot find module 'https://esm.sh/@unocss/preset-uno@0.55.1' Require stack: - /workspace/uno.config.ts`

It does not seem to make a difference whether I import from an absolute URL or using the import map.

### Reproduction

Minimal Github repo for reproduction: https://github.com/adamgreg/unocss_vscode_extension

Alternatively, set up a Deno workspace, create a `uno.config.ts` file that includes an import (see below), and enable the extension.

```typescript
import presetUno from "https://esm.sh/@unocss/preset-uno@0.55.1";

export default {
  presets: [presetUno()]
}
```

### System Info

Deno 1.36.1, VS Code 1.81.1

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

adamgreg:
By the way, thank you very much for UnoCSS! Other than this VS Code issue, my experience so far of using it in my Deno Fresh project has been excellent. It's great to have the convenience of Tailwind utility classes, but with full control over the rules, and a great interface for SSR.

I've been helping with a PR for Deno Fresh which will hopefully lead to UnoCSS becoming the default solution for styling using that framework - https://github.com/denoland/fresh/pull/1303 .

zhmushan:
https://github.com/unocss/unocss/blob/580a30165a021365e6f426277486ec0b1161274c/packages/config/src/index.ts#L38-L57

UnoCSS uses [JITI](https://github.com/unjs/jiti) to compile configuration files. Unfortunately, it cannot handle modules imported from HTTP.

adamgreg:
[...]

PR #3040, opened 2023-08-28
feat: Support deno fresh framework

zhmushan:
https://github.com/unocss/unocss/issues/3014


adamgreg:
Firstly, thanks a lot for getting in to this so quickly! 

Am I right in thinking the issues here are Deno-related, rather than Fresh-related? I think it would be a better idea to add a `deno: true` flag instead of looking for a `fresh.config.ts` file.

* Allows the extension to work with non-Fresh Deno projects (other web frameworks or running UnoCSS CLI)
* Avoids non-obvious constraints on Fresh users to have a `fresh.config.ts` file. The existence of this config file is a fairly new suggestion, to help support an optional ahead-of-time build feature, and the filename has no special meaning to Fresh itself.
* A setting for the extension would be easier to explain in documentation
* If another way is introduced in future to automatically detect Deno-based Uno config files, you can deprecate / ignore an extension setting, without impact on the user's codebase.
* It does seem that UnoCSS works well in Deno, and the only stumbling block is the VS Code extension, which presumably runs Node internally.

I suspect that for 99% of cases the only aspect of the config file in a Deno project that causes a problem for the VS Code extension will be the presence of HTTPS imports. Node already supports these under an experimental flag: https://nodejs.org/api/esm.html#https-and-http-imports . Is that something we could try in the extension?

I'm keen to help, but I'm unfamiliar with a lot of this. I'm sorry if I've got the wrong idea about something!

zhmushan:
@adamgreg I mostly agree with your viewpoint, and it is more appropriate to use the `deno: true` flag.

However, I don't recommend using the HTTPS import provided by Nodejs. I have already tried it:
1. It has performance and stability issues compared to Deno (Node.js is much slower and prone to crashing programs).
2. Let's assume that users who set `deno: true` have Deno installed locally.
3. There shouldn't be any code related to the Deno Loader in the UnoCSS codebase. This would be disadvantageous for maintenance purposes. (99% is not accurate. Deno's configuration file can also have an impact on the specifier. Let's allow Deno to handle these tedious tasks itself)

miguelrk:
I'm with @adamgreg in that it's a better idea to add a `deno: true` flag instead of looking for a `fresh.config.ts` file, which is not really enforced for fresh projects. From your comment @zhmushan , it wasn't clear to me if we are indeed making this change? But in any case, let's wait for @antfu 's comments before proceeding 👌🏼 💯

adamgreg:
@zhmushan That all sounds good to me, thanks.

One concern - am I right in thinking the approach currently depends upon using Deno to serialise the config to stdout as JSON? Won't this omit any functions embedded in the config (dynamic rules, preflights etc.)? I tried using JSON to pass config to the browser runtime and it didn't work very well.

I just tried an alternative approach:

```js
import * as esbuild from "https://deno.land/x/esbuild@v0.18.11/mod.js";

import { denoPlugins } from "https://deno.land/x/esbuild_deno_loader@0.8.1/mod.ts";

const result = await esbuild.build({
  plugins: [...denoPlugins()],
  entryPoints: ["uno.config.ts"],
  bundle: true,
  outfile: "uno.config.js",
  platform: "node",
});
```

In my case, this outputs an (admttedly ugly!) config file in CJS, which the VS Code extension can use successfully. Could be a good alternative to JSON serialisation?

The only issue that I experienced was that it did not pick up the import map, which I had defined in a `deno.jsonc` file. esbuild did not pick it up automatically, and if I explicitly told it about `deno.jsonc` it couldn't handle the comments in there. I had to copy the relevant content to a separate `import_map.json` and pass that as `denoPlugins({importMapURL: new URL("./import_map.json", import.meta.url).toString()})`

zhmushan:
@antfu thanks for your reply!

Next I will create a function that retrieves config info using the specifier.

Behind the scenes, it get the bundled config file using `deno_emit`, which is then processed by JITI. When deno-related config files (deno.json or deno.jsonc) are detected, this function will be used to get information from uno.config or fresh.config. Do you agree with this approach?

adamgreg:
@zhmushan that sounds good to me!

I tried bundling my own config file with `deno_emit` but it didn't work (I don't remember why, sorry!). I had more luck with `esbuild`.

PR #3056, opened 2023-08-31
feat(cli): output transformed files

lvjiaxuan:
Related https://github.com/unocss/unocss/pull/2948#issuecomment-1698641346

How about this feature? I need it, so I created this PR. Anyway, it's up to who reviews it. Any suggestions are welcome. 😊


antfu:
I feel this is a bit too specific. I don't feel this is very common, plus you can always move the files later. Or if you want to full customization, it's better to write a script than use the CLI.

Issue #3034, opened 2023-08-26
VS Code IntelliSense break when use CustomIconLoader or FileSystemIconLoader


chz:
### UnoCSS version

0.55.3

### Describe the bug

1. Clone reproduction link to local computer. 
2. Run in vscode. 

**Custom CustomIconLoader will break VS code IntelliSense.**

Uncomment in **uno.config.ts**
```js
 // load: loadCustomIconSet(),
```
Uncomment in **pages/index.vue**
```html
<!-- <div>
      Test Icon Loaded: <i class="text-blue i-load:plus"></i>
    </div> -->
```

Also IntelliSense not work for FileSystemIconLoader loaded icons: 
```js
umico: FileSystemIconLoader(
  './assets/svg',
  svg => svg.replace(/#fff/, 'currentColor'),
),
```

### Reproduction

Repo: https://stackblitz.com/edit/unocss-unocss-xyzsnc?file=uno.config.ts

### System Info

------------------------------
- Operating System: Darwin
- Node Version:     v18.16.1
- Nuxt Version:     3.7.0
- CLI Version:      3.7.0
- Nitro Version:    2.6.1
- Package Manager:  yarn@1.22.19
- Builder:          -
- User Config:      devtools, modules, sourcemap, nitro, app, vite, routeRules, experimental, imports
- Runtime Modules:  @unocss/nuxt@0.55.3
- Build Modules:    -
------------------------------

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

chz:
Also only usage will break suggestion but intellisense will work for default uno presets:
```js
umico: FileSystemIconLoader(
  './assets/svg',
  svg => svg.replace(/#fff/, 'currentColor'),
),
```

userquin:
it seems to be ok: 

![imagen](https://github.com/unocss/unocss/assets/6311119/68cf6958-1ab8-4939-a50b-ad3ecbfa9572)

[...]

Issue #3009, opened 2023-08-20
VSCode IntelliSense does not work if the configuration file is outside of the root directory


jozefmery:
### UnoCSS version

v0.55.2

### Describe the bug

Placing the UnoCSS configuration file outside of the project root directory breaks the VSCode IntelliSense. Based on the extension's output, it seems that the extension correctly finds and loads the configuration. Also, I'm using the CLI for generation and it seems to work correctly.

### Reproduction

1. Move the UnoCSS configuration file from the project root `/unocss.config.ts` into a sub-directory, e.g., `/foo/bar/unocss.config.ts` 
2. Point the extension to the configuration file by adding the following line in your VSCode `settings.json`: `"unocss.root": "foo/bar"`
3. Reload the editor (`Close & Open VSCode`, or `F1 > Developer: Reload Window`, or `F1 > UnoCSS: Reload UnoCSS`)
4. The IntelliSense should be broken

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

harrytran998:
Same problem with me!

Do like the docs said (https://unocss.dev/integrations/vscode#config) - but no luck 😆. Hope team can fix this!

darkyelox:
This happens to me too, the settings.json configuration seems to be recognized but the intellisense is not working until I move my config file to root and removed the settings.json property :/ this is a problem for monorepos

Simon-He95:
I think this issue could be closed

Issue #3000, opened 2023-08-18
VSCode extension does not provide intellisense for shortcuts or rules added in custom preset.


KraXen72:
### UnoCSS version

0.55.1

### Describe the bug

Shortcuts and rules added in custom preset files (which are defined in `configDeps` vite plugin option and imported to `uno.config.ts`) work in the resulting app, but don't have autocomplete, unless VSCode extension is reloaded with the reload command.

### Reproduction

I tried to reproduce it in stackblitz, but that apparently doesen't use the VSCode extension (i got no autocomplete at all), and the playground does not allow creating arbitrary ts files to import in the config.
Anyway, i made the basic repro in stackblitz, you can download it and open in VSCode.

How to repro:
- https://stackblitz.com/edit/unocss-unocss-i31ibt?file=uno.config.ts
- run `UnoCSS: reload` command to load the unocss extension LSP (workaround for extension version 0.55.1)
- add a new shortcut / rule in the custom-preset file.
- try using it in the code; it will work, but you will not get intellisense for it
    - old shortcut:
    ![image](https://github.com/unocss/unocss/assets/21956756/bbe7438c-36f4-4b56-bf3c-1345e32f1c0c)
    - new shortcut:
    ![image](https://github.com/unocss/unocss/assets/21956756/5ac75c86-0891-4b43-8622-002ad267c3ec)

- current workaround: if you define your shortcuts / rules directly in `uno.config.ts`, you will get intellisense immediately
- run `UnoCSS: reload` command, you now have intellisense

### System Info

Windows 10, Vivaldi, svelte

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

sixdjango:
me too bro~

KraXen72:
[...]

Issue #2742, opened 2023-06-07
VS code extension: support multiple workspace roots and/or autodetection


PhilippMolitor:
### Clear and concise description of the problem

I am using UnoCSS with a Yarn 1 + Turborepo monorepo.
My uno config lies in `apps/frontend` - which works when I use uno only there.
The issue with monorepos is, that if I decide to create another web application (imagine `apps/docs` ), now there is no way to use unocss in that repo correctly.

### Suggested solution

**Automatic configuration detection**

1. Scan vsc workspace for files names `uno.config.ts`
2. Assume location of each `uno.config.ts` is the package root of that package (maybe consider looking up a `package.json` next to it to be sure)
3. add package root and config location to the list of detected unocss workspaces
4. Probably it is a good idea to allow a settings key for excluding dir globs like `**/node_modules/*` from scanning for performance.

**Manual configuration**
`.vscode/settings.json`:
```json
{
  "unocss.packageRoots": [
    {
        "packageDir": "apps/frontend"
    },
    {
        "packageDir": "apps/docs",
        // config in a custom subdirectory. maybe even make this path relative to the `packageDir`?
        "config": "apps/docs/config/uno.config.ts"
    }
  ]
}
```

**Editor -> UnoCSS associations**
1. Each open VS code editor is monitored for their full path
2. If the path is matching one of the detected or configured packageRoots, look up the config of that package.
3. Apply autocompletion with that configuration

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

GreatAuk:
https://github.com/unocss/unocss/pull/2101

mikestopcontinues:
The existing functionality doesn't support this.

Another possible improvement is using the relative path params VSCode uses for other settings. Being able to add `{workspaceRoot}`, `{projectRoot}`, etc to the "unocss.root" array would help avoid having to configure this for each project.

PhilippMolitor:
not stale

louiss0:
I was trying to ask for detection for different mono-repos.
I'm not a fan of this non-auto-detection thing.

louiss0:
Can we please have auto-detection I don't want to have to configure the extension every time I open a new repo.
[...]

PR #3069, opened 2023-09-03
feat(vscode): If uno.root is not set,  auto-detection

Simon-He95:
With the popularity of monorepo, more and more libraries may use this mode, but it is very troublesome to configure `settings.json` additionally. I hope to automate changes more intelligently based on the `unocss.config` of my current project. #3106 #3034 #3009 #2742 #3000


packages/vscode/src/index.ts
```diff
@@ -70,29 +69,41 @@ export async function activate(ext: ExtensionContext) {
...
+
+  const cacheMap = new Set()
+
+  const registerUnocss = async () => {
+    const url = window.activeTextEditor?.document.uri.fsPath
+    if (!url)
+      return
+    if (/node_modules/.test(url))
+      return
+    const target = await findUp(['package.json'], { cwd: url })
```
antfu, line 82: Maybe we should search for the uno.config directly instead.

packages/vscode/src/index.ts
```diff
@@ -70,29 +69,41 @@ export async function activate(ext: ExtensionContext) {
...
+    if (!target)
+      return
+    const cwd = target.slice(0, -13)
```
antfu, line 85: Not in favor of this magic number, can we use `basename`?

packages/vscode/src/index.ts
```diff
@@ -70,29 +69,41 @@ export async function activate(ext: ExtensionContext) {
...
-          registerAnnotations(cwd, contextLoader, status, ext)
-          registerSelectionStyle(cwd, contextLoader)
-        }
-      }
-      contextLoader.reload()
-      log.appendLine('✅ Reloaded.')
-    }),
-  )
+
+  const cacheMap = new Set()
```
antfu, line 74: We should also cache for file resolution, so we don't need to run `findUp` every time user navigate to the file.

antfu: In addition, we could also listen to the fs on adding/removing the config file and purge the cache.

packages/vscode/src/index.ts
```diff
@@ -70,29 +69,41 @@ export async function activate(ext: ExtensionContext) {
...
+
+  const registerUnocss = async () => {
+    const url = window.activeTextEditor?.document.uri.fsPath
+    if (!url)
+      return
+    if (/node_modules/.test(url))
```
antfu, line 80: I think this might be better to be configurable, as an `exclude` option maybe.

antfu (changes requested):

antfu:
Also I don't think we should completely remove the `unocss.root` support, we could have this as a smart default when root is not specified, but keep the original behaviour without breaking change.

Simon-He95:
> Also I don't think we should completely remove the `unocss.root` support, we could have this as a smart default when root is not specified, but keep the original behaviour without breaking change.

Thank you for your review, I made some adjustments, please help to check it out

test/assets/variant-group.vue
```diff
@@ -1,8 +1,8 @@
 <script setup lang="ts">
 const a = 1
 const b = 2
-// eslint-disable-next-line @stylistic/ts/space-infix-ops
-const _c = a-(b -a -b)
+ 
+const _c = a - (b - a - b)
```
antfu, line 5: Please revert this

packages/vscode/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "publisher": "antfu",
-  "name": "@unocss/vscode",
+  "name": "unocss",
```
antfu, line 3: This should only be changed in dev, should not be committed

packages/vscode/src/index.ts
```diff
@@ -1,6 +1,9 @@
 import path from 'path'
+import { watch } from 'fs'
```
antfu, line 2: Better use `createFileSystemWatcher` https://code.visualstudio.com/api/references/vscode-api#workspace

packages/vscode/src/index.ts
```diff
@@ -44,10 +46,11 @@ export async function activate(ext: ExtensionContext) {
   status.text = 'UnoCSS'
 
   const root = config.get<string | string[]>('root')
+  const exclude = config.get<string | string[]>('exclude')
+  const include = config.get<string | string[]>('include')
```
antfu, line 50: We need to add these two configs in package.json

antfu (changes requested):

packages/vscode/src/index.ts
Simon-He95: Learned

Simon-He95:
Do you think it is necessary for us to monitor the file name changes or new creations of uno.config?

antfu (approved):

antfu:
Nice work! Thanks

PR #3083, opened 2023-09-05
fix(vscode): remToPx support important

chizukicn:
This PR will fix the problem that the `px to rem preview` of the vscode extension does not support important

image

### DIFF
--- a/packages/vscode/src/utils.ts
+++ b/packages/vscode/src/utils.ts
@@ -51,8 +51,7 @@ export function addRemToPxComment(str?: string, remToPixel = 16) {
     index = end
   }
   output.push(str.slice(index))
-  const rs = output.join('')
-  return rs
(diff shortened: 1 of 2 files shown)


packages/vscode/src/utils.ts
```diff
@@ -40,17 +40,19 @@ export function addRemToPxComment(str?: string, remToPixel = 16) {
...
     const end = index + rem.index + rem[0].length
+
     output.push(str.slice(index, end))
     output.push(px)
     index = end
   }
   output.push(str.slice(index))
-  return output.join('')
+  const rs = output.join('')
+  return rs
```
zyyv, line 55:
```suggestion
  return output.join('')
```

zyyv (approved):

antfu (approved):

PR #3098, opened 2023-09-09
feat(core): support factory with `definePreset`

chizukicn:
This PR will add a new API to instead `definePreset`
Why is this new API needed?

When developing presets, `definePreset` does not conform to the normal mode, and even unocss's built-in presets do not use it.

Here I refer to the design of `unplugin`
eg
image


antfu:
I am confused, why this is needed, how it is differentiated from `definePreset`?

chizukicn:
> I am confused, why this is needed, how it is differentiated from `definePreset`?

`createPreset` is exported by `@unocss/core` and does not need to depend on the `unocss` package. 

It accepts a factory function and returns it
It is used outside the function, just like `createPlugin` of `unplugin` and `defineComponent` of `vue`
```ts
export const presetAwesome = createPreset((options)=>{
   return {
      name: "unocss-preset-awesome"
  }
});
```
And `definePreset` can only be used on return values
```ts
export const presetAwesome = (options)=>{
   return definePreset({
      name: "unocss-preset-awesome"
  })
};
```
According to my observation, almost no one uses the `definePreset` API, but directly annotates the return type.

```ts
export const presetAwesome = (options):Preset=>{
   return {
      name: "unocss-preset-awesome"
  }
};
```

antfu:
Why don't we just move definePreset to core and support function overloads? In core we could also support function as presets value and call them automatically.

chizukicn:
Look !
Now `definePreset` will return a `DefinePreset` type, which can be used as a function or directly as an object
```ts
    const preset = definePreset((option) => {
      return {
        name: 'custom-preset',
        rules: [
          ['text-option', {
            color: option.color,
          }],
        ],
      }
    }, {
      color: 'dark',
    })

    const uno = createGenerator({
      presets: [
        preset,
        // or functional
        preset({
          color: "dark"
       })
      ],
    })
```

chizukicn:
I refactored the type of `definePreset` so now we no longer need to explicitly specify generics, but this might be a break change @antfu 
image

packages/core/src/config.ts
```diff
@@ -212,3 +212,13 @@ function mergeAutocompleteShorthands(shorthands: Record<string, string | string[
...
 }
+
+export function definePreset<P extends Preset>(preset: PresetFactory<P>): DefinePreset<P>
+
+export function definePreset<P extends Preset>(preset: P): DefinePreset<P>
+
+export function definePreset<P extends Preset, Args extends any[]>(preset: PresetFactory<P, Args>, ...defaultOptions: Args): DefinePreset<P, Args>
+
+export function definePreset<P extends Preset, Args extends any[]>(preset: DefinePreset<P, Args>, ...defaultOptions: Args) {
+  return functional(preset, ...defaultOptions)
```
antfu, line 223: I don't see we need such proxy. Instead, we could just support evaluating functionally preset in core.

chizukicn:
Let’s implement a simple and type-friendly `definePreset` that returns a function first. @antfu

antfu (approved):

PR #3100, opened 2023-09-10
fix(astro): HMR not work in windows

Dunqing:
fix: https://github.com/unocss/unocss/issues/2987#issuecomment-1708376281

### DIFF
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -5,6 +5,7 @@ import type { VitePluginConfig } from '@unocss/vite'
 import VitePlugin from '@unocss/vite'
 import type { UserConfigDefaults } from '@unocss/core'
 import type { Plugin, ResolvedConfig } from 'vite'
+import { normalizePath } from 'vite'
 import { RESOLVED_ID_RE } from '../../shared-integration/src/layers'
 
 const UNO_INJECT_ID = 'uno-astro'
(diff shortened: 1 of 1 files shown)


userquin:
@Dunqing testing PR on my Windows laptop...

userquin:
@Dunqing still getting page reload:

imagen

Dunqing:
> @Dunqing still getting page reload:

Yes, I am not sure page reload In astro is expected. Can you use `vite` plugin to test it?

Dunqing:
> @Dunqing testing PR on my Windows laptop...
> 
> EDIT: since UnoCSS dependencies are linked we need to include `vite.server.fs.strict = false` in the example Astro config file: I'm getting 403 when resolving astro hmr JS module:
> 
> imagen

I tested it in this [example](https://github.com/unocss/unocss/tree/main/examples/astro), but I am not receiving any error. And looks like the error has nothing to do with the `astro` integration

userquin:
I have this on the Astro config file:

imagen

userquin:
@Dunqing adding the vite-plugin-inspect:

This is weird in SSR, there are a few `/__uno.css` virtual modules, one resolving to `/__uno.css` and another to the full path:

imagen

imagen

Then in the client, we have also 2 `__uno.css` (is there a missmatch? the HMR context is for `/__uno.css` and the virtual id for update, check also that id is resolved twice or 3 times):

imagen

imagen

Dunqing:
> @Dunqing adding the vite-plugin-inspect:
> 
> This is weird in SSR, there are a few `/__uno.css` virtual modules, one resolving to `/__uno.css` and another to the full path:
> 

This full path is to align `data-astro-dev-id` and `data-vite-dev-id`. 

You can look here
https://github.com/unocss/unocss/blob/92e2f8eece7b3be68c29ad0aa3365b6640a69820/packages/astro/src/index.ts#L34-L38

This is strange, and we may need to return the full `/__uno.css` path in the `vite` plugin to solve the problem of resolving it twice!

Dunqing:
If this PR does solve the problem of Windows `hmr` not working, I think it would be a good idea to merge this PR first to avoid Windows users not being able to use it for a long time. What do you think? @userquin

antfu (approved):

userquin (approved):

jamal7887:
Unocss doesn't work if it put it in the astro integration it only works when a put it as a vite plugin

Dunqing:
> Unocss doesn't work if it put it in the astro integration it only works when a put it as a vite plugin

Did you test it at this PR? The latest version does not include this PR.

PR #3115, opened 2023-09-13
chore(ci): remove 16 version

zyyv:


Dunqing:
Can we add 20?

antfu:
I think before we do this, we should update our engine field, and make a breaking change to say we no longer support Node 16

https://github.com/unocss/unocss/blob/4e3cdc7e5a5c6ec2b3e3065d7ff8b45e57020060/packages/config/package.json#L32-L34

PR #3116, opened 2023-09-13
feat(attributify-jsx): allow pass component as prop

zhiyuanzmj:
``` tsx
<div components={<div absolute bottom-5></div>}></div>
```
will be transform to
``` tsx
 <div components={<div absolute=\\"\\" bottom-5=\\"\\"></div>}></div>
```


antfu (approved):

zoeyzhao19:
seems to cause [3471](https://github.com/unocss/unocss/issues/3471)

Issue #3103, opened 2023-09-10
Abnormal Build Time When Used with Astro 3.0 ViewTransitions


Alecyrus:
### UnoCSS version

0.53.6

### Describe the bug

After using the new feature 'ViewTransitions' in Astro 3.0, when simultaneously using 'unocss', it takes more than 20 seconds during 'dev' or 'build', with a significant amount of time spent on building the client. However, without using 'ViewTransitions' or 'unocss', both 'dev' and 'build' processes are much faster.

### Reproduction

https://stackblitz.com/edit/withastro-astro-ptdkgl

1. run "npm run build", you will see it takes 19s to finish:
<img width="383" alt="CleanShot 2023-09-10 at 21 27 47@2x" src="https://github.com/unocss/unocss/assets/13865581/31ef8e32-5e2b-47fa-af1c-9506c1103582">

2. then comment `<ViewTransitions />` in `BaseHead.astro` or `Unocss` in `astro.config.mjs`, everything is fast as expected.

<img width="418" alt="CleanShot 2023-09-10 at 21 29 57@2x" src="https://github.com/unocss/unocss/assets/13865581/c1ea6120-71f2-4e3f-b462-3d9ae82edf56">

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Alecyrus:
https://github.com/withastro/astro/issues/8490

HummingMind:
+1. Same issue.

Dunqing:
Thanks for your feedback. I will take a look at this later.

HummingMind:
This issue is still present in v0.56.0
@antfu @Dunqing

Dunqing:
> This issue is still present in v0.56.0 @antfu @Dunqing

Sorry about that, I made a mistake 🥵, I sent a [PR](https://github.com/unocss/unocss/pull/3144) to fix this.

HummingMind:
@Dunqing 
[...]

PR #3123, opened 2023-09-14
fix(extractor-arbitrary-variants): remove sourcemap before matching attributes

Dunqing:
fix: #3103

After I did some digging, I found out that if the source code contains an inline source map, it makes the extractor very slow. So I think we should remove source map and then matching


antfu (approved):

Dunqing:
Can we move this change to `applyExtractor`. this will be useful for all extractors.

antfu:
I don't think we should do it in core, the core should be language agnostic.

Dunqing:
> I don't think we should do it in core, the core should be language agnostic.

Ok, I got it.

Issue #2874, opened 2023-07-15
Complex attributify usecase broken on >= 0.50.1


XiNiHa:
### UnoCSS version

`>= 0.50.1`

### Describe the bug

Bumping UnoCSS version to >= 0.50.1 broke my usecase, which is relatively complicated:
<img width="665" alt="스크린샷 2023-07-15 오후 4 03 03" src="https://github.com/unocss/unocss/assets/8155259/a882f4e5-a5ac-495e-bde4-7985bd9485ce">

You can spot that `un-h-full` is not generated.

Either removing `<Popover fixed={...} />` part or `class={clsx(...)}` part resolves the issue.

Maybe #2238 is the cause?

### Reproduction

https://stackblitz.com/edit/vitejs-vite-yqejct?file=src%2FApp.tsx

You can see that `p-10` and `shadow-xl` applies, but `un-bg-red` doesn't.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

antfu:
Would you like a send a PR with the failing test? Thanks

XiNiHa:
missed the notification 🤦

PR #3126, opened 2023-09-16
chore(attributify): add test case for #2874

XiNiHa:
Adds test case for #2874


zyyv:
Will be fixed in #3275

Issue #3071, opened 2023-09-03
theme() directive does not support color transparency


minht11:
### UnoCSS version

0.55.6

### Describe the bug

Using theme() directive with transparency does not work. This code...
```css
.item {
  background: theme('colors.red/10');
}
```
...fails. I tried using postcss plugin and transformer, neither worked. Using apply `bg-red/10` works, though it still have issues with resolving theme config.

Semi related:
https://github.com/unocss/unocss/issues/2946
https://github.com/unocss/unocss/issues/2851

### Reproduction

Try using
```css
.item {
  background: theme('colors.red/10');
}
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #3131, opened 2023-09-17
feat(transformer-directives): `theme()` support  color transparency

Dunqing:
close: #3071


packages/transformer-directives/src/theme.ts
```diff
@@ -1,4 +1,5 @@
 import type { Declaration } from 'css-tree'
+import { colorToString, parseCssColor } from '@unocss/preset-mini/utils'
```
hannoeru, line 2: Maybe we should move these utils to core?

packages/transformer-directives/package.json
```diff
@@ -37,6 +37,7 @@
   },
   "dependencies": {
     "@unocss/core": "workspace:*",
+    "@unocss/preset-mini": "workspace:*",
```
antfu, line 40: We shouldn't rely on the preset. I think we could move them to a new package `@unocss/utils`

antfu (approved):

PR #3134, opened 2023-09-18
feat: add `@unocss/utils` package

Dunqing:
Exporting generic functions from `@unocss/utils`.


packages/preset-mini/src/_utils/utilities.ts
```diff
@@ -1,7 +1,7 @@
 import type { CSSEntries, CSSObject, DynamicMatcher, ParsedColorValue, RuleContext, StaticRule, VariantContext } from '@unocss/core'
-import { isString, toArray } from '@unocss/core'
+import { toArray } from '@unocss/core'
 import type { Theme } from '../theme'
-import { colorOpacityToString, colorToString, parseCssColor } from './colors'
+import { colorOpacityToString, colorToString, getComponents, parseCssColor } from '../utils'
```
antfu, line 4: We could directly reference to `@unocss/utils` and avoid circular imports.

package.json
```diff
@@ -8,7 +8,7 @@
     "postinstall": "esno scripts/prepare.ts",
     "taze": "taze minor -wIr && pnpm -r --parallel run update-post",
     "bench": "npm -C bench run bench",
-    "build": "rimraf packages/*/dist && esno scripts/copy-files.ts && pnpm -r --filter=./packages/* run build && pnpm -r run build-post",
```
Dunqing, line 11: Because the `@unocss/inspector` is directly import `preset-mini`, we need to build the utils first.
https://github.com/unocss/unocss/blob/02b06c28d6748e6bb6c969597f25bba112238b0f/packages/inspector/src/analyzer.ts#L2

packages/utils/src/utilities.ts
```diff
@@ -0,0 +1,63 @@
+import { isString } from '@unocss/core'
```
zyyv, line 1: I think these basic utility functions can be moved into the `utils` package

packages/utils/package.json
```diff
@@ -0,0 +1,42 @@
...
+  ],
+  "engines": {
+    "node": ">=14"
+  },
+  "scripts": {
+    "build": "unbuild",
+    "stub": "unbuild --stub"
+  },
+  "dependencies": {
+    "@unocss/core": "workspace:^"
```
zyyv, line 40: As a utils package, we shouldn't have other dependencies. 
Instead, we should migrate the functions in `@unocss/core` to `@unocss/utils`, and import `@unocss/utils` dependencies in `@unocss/core`.

packages/utils/src/utilities.ts
Dunqing: Yes! We can

packages/utils/package.json
Dunqing: Agree

packages/core/src/utils/index.ts
```diff
@@ -1,10 +1,9 @@
...
 export * from './object'
-export * from './basic'
 export * from './helpers'
 export * from './map'
 export * from './countable-set'
 export * from './layer'
 export * from './variant-group'
 export * from './warn'
 export * from './handlers'
+export * from '@unocss/utils'
```
zyyv, line 9: We don't need to export it, I think. 🤔

packages/preset-mini/src/_utils/index.ts
```diff
@@ -1,5 +1,5 @@
-export * from './colors'
 export * from './mappings'
 export * from './handlers'
 export * from './variants'
 export * from './utilities'
+export * from '@unocss/utils'
```
zyyv, line 5: And it.

packages/core/src/utils/index.ts
zyyv: > It's a huge amount of work. 

I can help you do this.

> I don't think this is the time.

I dont't know what's your worried about? You mean, these changes will break the previous usage, right?
If true, 
Currently, most of the utility functions of `@unocss/core` and `@unocss/mini` are used by `preset` developers or others, but it do not affect the use of simple user. Now that `@unocss/utils` exists, we should require developers to import it correctly in new versions.

package.json
antfu: We can import the source file from the utils package, so build is not required.

packages/utils/package.json
antfu: Well, I am thinking the other way around. If some utils are used in core, then let them stay in core. `@unocss/utils` should be pictured as "utils for creating rules/presets", so it can have tools like color parsing etc. While the `core` is pretty much a CSS-agnostic engine that just caring about extracting and conacting strings.

I'd prefer `@unocss/utils` to deps on `@unocss/core` and keep `@unocss/core` deps-free.

If the name is a bit confusing, maybe we could have `@unocss/rule-utils`.

package.json
Dunqing: The `parseColor` is not in the utils. Since it relies on many `preset-mini` methods. So I think it should stay where it is

packages/utils/package.json
Dunqing: `@unocss/rule-utils` Seems good to me. What do you think? @zyyv

package.json
antfu: Then you will need to update the alias.ts and tsconfig.json to point the utils package to the ts entry

packages/core/src/utils/index.ts
Dunqing: Done

packages/utils/package.json
Dunqing: Now rename to `@unocss/rule-utils`

antfu (approved):

Dunqing:
These breaking changes look like they don't exist

antfu:
Oh yeah haha I forgot to push

PR #3137, opened 2023-09-18
Update postcss.md (add necessary in config nuxt)

lonagi:
Setuping project by this doc doesn't work, but when I add path to my css file (yes, assets/css) it works!

### DIFF
--- a/docs/integrations/postcss.md
+++ b/docs/integrations/postcss.md
@@ -28,6 +28,18 @@ This package is in an experimental state right now. It doesn't follow semver, an
 :::
 
 ```ts
+// nuxt.config.ts
+export default defineNuxtConfig({
+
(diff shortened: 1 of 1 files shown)


docs/integrations/postcss.md
```diff
@@ -27,6 +27,18 @@ This package is in an experimental state right now. It doesn't follow semver, an
...
+// nuxt.config.ts
+export default defineNuxtConfig({
+
+  css: ['~/assets/css/styles.css'], // add here path to your main css file
+  modules: [
+    '@unocss/nuxt',
+  ],
+})
+```
+:::
```
antfu, line 40: Emm, no, postcss is a framework-agnostic plugin. We should not add this here as it can be confusing to users of other frameworks. Also I believe `@unocss/nuxt` is not the correct entry to use postcsee

PR #3142, opened 2023-09-20
feat(eslint-plugin): Add calls support for order

Nirlah:
Extend the eslint’s order rule to support call expressions, and apply it to [clsx](https://www.npmjs.com/package/clsx).


antfu:
Can you add some tests? Thanks

Nirlah:
@antfu I looked into adding tests, but I can’t seem to find where the tests for the _eslint_ plugin reside. There is `test/sort-rules.test.ts`, which tests the core rather than the _eslint_ plugin.

I will be happy to look into adding a test suit for the plugin, but just making sure I’m not missing anything...

antfu:
We do have a simple fixture. But yeah it's missing the automation. Would be great if you can help improving that.

antfu:
I did this https://github.com/antfu/eslint-config/blob/main/test/fixtures.test.ts, for your reference

PR #3147, opened 2023-09-20
fix(transformer-attributify-jsx): transform error when html element is like a atomic class

evanryuu:
## Summary

This PR modifies [attributeRE](https://github.com/evanryuu/unocss/blob/main/packages/transformer-attributify-jsx/src/index.ts#L41) to solve transforming error when a class-like HTML element happens to have atomic attributes (not class names).

## Details

When the HTML element is like a class name, say `<h1>`, and the tag happens to have atomic attributes such as `text-red`, the transformer will accidentally transform the closing tag in an unexpected way.

Reproduction:
1. [Stackblitz](https://stackblitz.com/edit/vitejs-vite-k49hbr?file=uno.config.ts)
2. [My Github Repo](https://github.com/evanryuu/uno-jsx-repo)

**Code:**
image

**Error:**
image

This PR makes the reg will not match the attribute starts with `<` or `</`

### DIFF
--- a/packages/transformer-attributify-jsx/src/index.ts
+++ b/packages/transformer-attributify-jsx/src/index.ts
@@ -38,7 +38,7 @@ export interface TransformerAttributifyJsxOptions {
 }
 
 const elementRE = /(<\w[\w:\.$-]*\s)([\s\S]*?)(?=>[\s\S]?<\/[\s\w:\.$-]*>|\/>)/g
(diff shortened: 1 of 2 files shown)


evanryuu:
I added a new test case instead of modifying [originalCode](https://github.com/unocss/unocss/blob/main/test/transformer-attributify-jsx.test.ts#L9) because I'm not sure if it's suitable for me to change the original test case directly.

antfu (approved):

PR #3148, opened 2023-09-20
feat: improve interative docs search

evanryuu:
## Summary

This PR can make the search body more accurate when the input is class name (opionated😂)

### Before
Currently, in the [interactive docs](https://unocss.dev/interactive/), if I type `display: none`, the `.hidden` is not the first to show. From my personal point of view, it's a bit anti-instinct.

image

image

### After
If users input a CSS class name, I think maybe it's better to provide the exact atomic class that they need?

**P.S. It can also provide dynamic rules**

image
image

### DIFF
--- a/packages/shared-docs/src/search.ts
+++ b/packages/shared-docs/src/search.ts
@@ -34,7 +34,7 @@ export function createSearch(
         },
         {
           name: 'body',
-          weight: 0.4,
+          weight: 0.6,
         },
         {
           name: 'title',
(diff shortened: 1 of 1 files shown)


evanryuu:
I'm not sure if this is ideal but I personally think it will provide a better searching experience for the devs who just come across UnoCSS 😄 It is a reeeeeally powerful and fast atomic CSS library, so I hope the searching experience is also impeccable for users!

packages/shared-docs/src/search.ts
```diff
@@ -84,14 +84,24 @@ export function createSearch(
     if (input.match(/^rand(om)?:/))
       return sampleArray(fuseCollection, limit)
 
+    // css attr
+    if (input.match(/^[a-zA-Z-]+:\s?[a-zA-Z0-9\s.#]+;?$/)) {
```
antfu, line 88: I am not sure if this handling would be a bit too specific

evanryuu: > I am not sure if this handling would be a bit too specific

I think you are right 😂 Maybe we can identify the attr by `/.*:.*/` or just simply use `includes(':')` ?

PR #3087, opened 2023-09-06
feat(preset-mini): support fontSize with letterSpace

zyyv:
close #2906

PR #3149, opened 2023-09-21
feat(preset-mini): support fontSize with letter-space

### LINKED ISSUE #2906: Support additional config `letter-spacing` inside `fontSize`

Currently, UnoCSS supports specifying font size inside the fontSize theme config:
```js
theme: {
    fontSize: {
      sm: ['15px'],
    },
```
or additionally, also specify line-height:
```js
theme: {
    fontSize: {
      sm: ['15px', '1.5em'],
    },
```
Similar to Tailwind theme config, we should also allow an additional config for letter-spacing:
```js
theme: {
    fontSize: {
      sm: ['15px', '1.5em', '-0.004em'],
    },
```
This is helpful for typesetting for perfect looking text :)

### Suggested solution

Allow an optional additional config inside the array:

```js
theme: {
    fontSize: {
      sm: [size, line height, letter-spacing],
    },
```

```js
theme: {
    fontSize: {
      sm: ['15px', '1.5em', '-0.004em'],
    },
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

zyyv:
close #2906, duplicate of #3087 

Some build conifg changes refer to #711


packages/core/src/utils/basic.ts
```diff
@@ -1,7 +1,13 @@
+import type { Arrayify } from '../types'
+
 export function toArray<T>(value: T | T[] = []): T[] {
   return Array.isArray(value) ? value : [value]
 }
 
+export function toArrayReadonly<T>(value: T) {
+  return (Array.isArray(value) ? value : [value]) as Arrayify<T>
+}
```
antfu, line 9: I am not sure if to have a runtime utility that only does type assessment (increase bundle size but not helpful at runtime). Also I think the name `Arrayify` is a bit confusing as it does not indicate it's readonly part. Maybe a `Readonly<ToArray<T>>` would be more straightforward.

packages/preset-mini/build.config.ts
```diff
@@ -13,5 +13,9 @@ export default defineBuildConfig({
   declaration: true,
   rollup: {
     emitCJS: true,
+    dts: {
+      respectExternal: false,
+    },
   },
+  failOnWarn: false,
```
antfu, line 20: Any reason these are added? `failOnWarn: false` feels dangerous

packages/core/src/utils/basic.ts
antfu: Reverted them so I could merge this feature first. We could discuss about it somewhere else

packages/preset-mini/build.config.ts
zyyv: I also feel is not good, but ci error. see https://github.com/unocss/unocss/actions/runs/6608185990/job/17946548533?pr=3149#step:9:277

packages/core/src/utils/basic.ts
zyyv: tks ~

PR #3165, opened 2023-09-25
docs: improve "why" section

codejet:
Just a few small improvements/corrections to the "Why" section of the docs.

### DIFF
--- a/docs/guide/why.md
+++ b/docs/guide/why.md
@@ -14,9 +14,9 @@ We recommend you to read the blog post [Reimagine Atomic CSS](https://antfu.me/p
 
 UnoCSS was started by one of the [Windi CSS](https://windicss.org/)'s team members, with a lot of inspiration taken from the work we did in Windi CSS. While Windi CSS is no longer actively maintained (as of March 2023), you may consider UnoCSS as a *"spiritual successor"* of Windi CSS.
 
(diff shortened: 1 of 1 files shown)


docs/guide/why.md
```diff
@@ -14,17 +14,17 @@ We recommend you to read the blog post [Reimagine Atomic CSS](https://antfu.me/p
...
 
-UnoCSS inherits Windi's on-demand natural, [attributify mode](/presets/attributify), [shortcuts](/config/shortcuts), [variant groups](/transformers/variant-group), [compilation mode](/transformers/compile-class) and a lot more. On top of that, UnoCSS is built ground up with the maximum extensibility and performance in mind, making us able to introduce new features like [pure CSS icons](/presets/icons), [valueless attributify](/presets/attributify#valueless-attributify), [tagify](/presets/tagify), [web fonts](/presets/web-fonts), etc.
+UnoCSS inherits Windi's on-demand, natural [attributify mode](/presets/attributify), [shortcuts](/config/shortcuts), [variant groups](/transformers/variant-group), [compilation mode](/transformers/compile-class) and a lot more. On top of that, UnoCSS is built ground up with the maximum extensibility and performance in mind, making us able to introduce new features like [pure CSS icons](/presets/icons), [valueless attributify](/presets/attributify#valueless-attributify), [tagify](/presets/tagify), [web fonts](/presets/web-fonts), etc.
```
antfu, line 17: This was intentional. `on-demand` serves as an adj here. Any better way of expressing that?

codejet: But as an adjective of what exactly? "Natural" is an adjective itself, hence to me it looked like those two adjectives apply to Windi's attribute mode.

antfu: Oh, good point, then it should be `on-demand nature` 👍

codejet: Ah, got it ☺️ Will change it.

PR #3167, opened 2023-09-25
feat: CountableSet constructor with initial map

kovsu:
related to #2986


antfu:
Can you add a test to it? The main branch seems to be fine

kovsu:
@antfu I don't know why the ci's test all passed. but I get the test failed locally you can see here. :)
截屏2023-09-26 14 18 01

the [falied test file](https://github.com/unocss/unocss/blob/main/test/countable-set.test.ts) is here.

antfu:
Did you build first?

kovsu:
oh my bad T.T, sry.

Issue #3175, opened 2023-09-27
feat: ask for !(bg-black text-white) variant transformer group support


SnowingFox:
### Clear and concise description of the problem

```ts
<div className={'!bg-black !text-white'} />
```
we will write `!` twice to implement this style, and it's not intuitive

### Suggested solution

we can do it like this

```ts
<div className={'!(bg-black text-white)'} />
```

### Alternative

```ts
<div className={'!bg-black !text-white'} />
```

### Additional context

i will try to provide a pr to solve this problem

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
You could do `important:(bg-black text-white)` today tho 👀

PR #3177, opened 2023-09-27
feat: support "!(bg-black text-white)" variant group

SnowingFox:
close https://github.com/unocss/unocss/issues/3175

### DIFF
--- a/packages/transformer-variant-group/src/index.ts
+++ b/packages/transformer-variant-group/src/index.ts
@@ -1,5 +1,5 @@
 import type { SourceCodeTransformer } from '@unocss/core'
-import { parseVariantGroup } from '@unocss/core'
+import { defaultVariantGroupSeparators, parseVariantGroup } from '@unocss/core'
 
 export interface TransformerVariantGroupOptions {
   /**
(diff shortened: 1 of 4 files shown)


packages/transformer-variant-group/src/index.ts
```diff
@@ -26,7 +26,7 @@ export default function transformerVariantGroup(
     name: '@unocss/transformer-variant-group',
     enforce: 'pre',
     transform(s) {
-      const result = parseVariantGroup(s, options.separators)
+      const result = parseVariantGroup(s, options.separators || [...defaultVariantGroupSeparators, '!', ''])
```
antfu, line 29: I am not so sure if we should treat it this special. `!text-sm` is a syntax introduced by `preset-mini`, while `@unocss/transformer-variant-group` should be rule-agnostic as possible.

Simon-He95: I'm thinking if we could use a vscode plugin to provide some conversion functionality to reduce the work of unocss at runtime. Maybe I can implement `!(text-red bg-red)` into `!text-red !bg-red` in [`unot`](https://github.com/Simon-He95/unot)

Simon-He95: > I am not so sure if we should treat it this special. `!text-sm` is a syntax introduced by `preset-mini`, while `@unocss/transformer-variant-group` should be rule-agnostic as possible.

I'm thinking if we could use a vscode plugin to provide some conversion functionality to reduce the work of unocss at runtime. Maybe I can implement !(text-red bg-red) into !text-red !bg-red in [unot](https://github.com/Simon-He95/unot)

PR #3203, opened 2023-10-08
docs: introduce config opitons

zyyv:


antfu:
I appreciated the initiative, but I would honestly prefer to redirect ppl to read jsdocs instead, to avoid duplicating those documentation and creating misalignment and extra maintenance effort in the long run.

We could use apps like https://www.jsdocs.io/ (currently not working for some reason https://github.com/jsdocs-io/web/issues/503) or similar to display that information if users prefer to read them with better UI.

zyyv:
I see https://www.jsdocs.io/package/@unocss/core is already working fine.

All type analysis makes searching above somewhat difficult. 

> if users prefer to read them with better UI.

i think jsdocs does not have a `TOC` component that allows users to quickly and accurately locate, and `vitepress` has it and better UI.

We can supplement key configuration documents in the `unocss.dev` and link to `jsdocs` for global type descriptions. How do you think?

antfu:
The main thing I don't want is the duplication and maintenance effort. If you find a better UI than [jsdocs.io](https://www.jsdocs.io/) I am happy to adapt to it (or if you want to build one 😜), or somehow generate from source code and put it directly in our docs. I would prefer a external tool over a in house scripts is possible.

zyyv:
It's implemented like [`Vitest` option](https://vitest.dev/config/#options), I still expect it to be merged.

antfu (approved):

antfu:
I would still wish this could be automated eventually (maybe by creating a VitePress plugin?). But let's have it for now

PR #3204, opened 2023-10-09
feat(preset-mini): support for defining css property with the `theme()` function

Dunqing:
Reference: https://discord.com/channels/937808017016119440/937811334907723806/1160763565096062996
Tailwind has supported this feature https://play.tailwindcss.com/OE4yynQhku

### Main features

1. Support for defining css property with the `theme()` function

These usage `unocss` are support
```css
'[padding:theme(spacing.xl)]',
'[color:theme(colors.blue.300/40%)]',
'[--css-variable-color:theme(colors.red.500)]',
'[--css-variable-color:theme(colors.red.500/50%)]',
'[--css-variable-color:theme(colors.green.500),theme(colors.blue.500)]',
'[--css-spacing:theme(spacing.sm)]',
```

2. Using `theme()` no longer requires adding quotes（tailwind has been supports）

This means that these syntaxes are supported
```css
theme(colors.red.500)
theme('colors.red.500')
theme("colors.red.500")
```
### Bug fixes

If my code is  `[--css-variable-color:theme(colors.red.500)]'` it will not extractor. You can see [here](https://github.com/unocss/unocss/compare/main...Dunqing:feat/css-property/theme-function?expand=1#diff-b705f7c1fefaf1eb7db5ed6716afa4812a773cec63becf73dbe8801ea3cd87cdR13)

### Additional
To accomplish this and avoid having to re-write the processing logic for `theme()`, I've abstracted away the generic logic into `transformThemeFn` function and placed it into `rule-utils` package, which I'm not sure is the right place for it.


zyyv:
We could optimize this rule to support those features.
https://github.com/unocss/unocss/blob/f0bbe3dd631357f3682bfb02d092e2613c7b1737/packages/preset-mini/src/_rules/variables.ts#L28-L41

packages/rule-utils/src/themeFn.ts
```diff
@@ -4,7 +4,9 @@ import { colorToString, parseCssColor } from './colors'
 export const themeFnRE = /theme\(\s*['"]?(.*?)['"]?\s*\)/g
 
 export function hasThemeFn(str: string) {
-  return str.includes('theme(') && str.includes(')')
```
Dunqing: Personally, I think this is enough. We don't need completely accurate results here. This may also be a small performance optimization @zyyv

zyyv: I hold another opinion. I think strict judgment should be maintained here. Performance optimization here is ignorable.

packages/rule-utils/src/themeFn.ts
```diff
@@ -0,0 +1,47 @@
...
+  themeFnRE.lastIndex = 0
+  return result
+}
+
+export function transformThemeFn(code: string, theme: Record<string, any>, throwOnMissing = true) {
+  const s = new MagicString(code)
+  const matches = Array.from(code.toString().matchAll(themeFnRE))
+
+  if (!matches.length)
+    return s.toString()
```
antfu, line 17:
```suggestion
    return code
```

antfu: I think we could move `const s = new MagicString(code)` after this condition

packages/rule-utils/src/themeFn.ts
antfu: Yeah I agree with @Dunqing that `String.includes` is enough to serve as the early return guard. We could let `transformThemeFn` return `false` if not matched. This way we keep the common utils performant, while keeping it accurate.

packages/rule-utils/src/themeFn.ts
Dunqing: Good catch!

zyyv (approved):

packages/rule-utils/src/themeFn.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function hasThemeFn(str: string) {
+  return str.includes('theme(') && str.includes(')')
+}
+
+export function transformThemeFn(code: string, theme: Record<string, any>, throwOnMissing = true) {
+  const matches = Array.from(code.toString().matchAll(themeFnRE))
+
+  if (!matches.length)
+    return code
```
antfu, line 14: Here we could return `false`

packages/preset-mini/src/_rules/variables.ts
```diff
@@ -26,14 +26,21 @@ export const cssVariables: Rule[] = [
...
     const value = rest.join(':')
+
     if (!isURI(body) && prop.match(/^[a-z-]+$/) && isValidCSSBody(value)) {
-      const parsed = h.bracket(`[${value}]`)
+      let parsed
+
+      if (hasThemeFn(value))
+        parsed = transformThemeFn(value, theme)
+      else
+        parsed = h.bracket(`[${value}]`)
```
antfu, line 42:
```suggestion
      if (!parsed)
        parsed = h.bracket(`[${value}]`)
```

So here we could fallback if the theme fn ends up invalid by the regex.

packages/rule-utils/src/themeFn.ts
Dunqing: I think we should keep the return `code`. Usually, we don't need to handle the return. If it return `false`, we need to handle the `false` case in every place where we use `transformThemeFn`

packages/preset-mini/src/_rules/variables.ts
Dunqing: Can we use https://github.com/unocss/unocss/pull/3204/commits/5d53238ae5ef652ad7046c7279b0bc4e6d933eee instead?

PR #3207, opened 2023-10-09
feat(preset-mini): support for define css property

zyyv:
close #3204


Dunqing:
I don't see the obvious distinction between this PR and #3204. Are there any advantages to this PR implementation?

Issue #3206, opened 2023-10-09
transformer variant group changes regular expressions


schelmo:
### UnoCSS version

0.56.5

### Describe the bug

when using the variant group transformer and including js/ts files in your unocss pipeline like:
```js
UnoCSS({
  content: {
    pipeline: {
      include: [
        /\.(vue|svelte|[jt]sx?|mdx?|astro|elm|php|phtml|html)($|\?)/,
      ],
    },
  }
})
```
unocss is changing the regular expressions with brackets in it.
like the camelizeRE from vue
```js 
var camelizeRE = /-(\w)/g;
```
becomes:
```js
var camelizeRE = /-\w/g;
```
this happens also with user code:
checkout out `App.vue` in the reproduction and watch the Browser console
the brackets gets eliminated from the regular expression

for the vuejs transformations, watch the stackblitz console and then comment out the
```js 
transformers: [transformerVariantGroup()],
```
line

### Reproduction

https://stackblitz.com/edit/unocss-unocss-krecms?file=src%2FApp.vue,vite.config.ts,src%2Fmain.ts

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

LiZhequ:
I have the same problem

Simon-He95:
The transformer variant group needs to be redesigned. For non-css files, only the attribute area in the tag is processed, not globally. Or if you want a file to be used globally, add // @unocss-transformer-variant-group at the top. I don't know if this is reasonable. Do you have any good suggestions? @zyyv

Simon-He95:
related: #4388

PR #3211, opened 2023-10-10
fix(variant-group): transformer variant group changes regular express…

Simon-He95:
fix: #3206

I don't know if it is reasonable to filter out the matched regexClassGroup in the template in this way, or if there is a better implementation method, but this is the solution I currently think of without introducing parser.

### DIFF
--- a/packages/core/src/utils/variant-group.ts
+++ b/packages/core/src/utils/variant-group.ts
@@ -23,12 +23,23 @@ export function parseVariantGroup(str: string | MagicString, separators = ['-', 
   let content = str.toString()
   const prefixes = new Set<string>()
   const groupsByOffset = new Map<number, VariantGroup>()
-
+  const processedStr = (str: string) => str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+  const isAttributor = (matchText: string, offset: number) => {
+    const reg = new RegExp(`<[\\w_\\-]+[^>\\/]*${processedStr(matchText)}[^>\\/]*>`, 'gm')
+    for (const match of content.matchAll(reg)) {
+      const index = match.index!
+      if (index < offset && (index + match[0].length > offset))
+        return true
+    }
+    return false
+  }
   do {
     hasChanged = false
     content = content.replace(
       regexClassGroup,
       (from, pre: string, sep: string, body: string, groupOffset: number) => {
+        if (!isAttributor(from, groupOffset))
+          return from
         if (!separators.includes(sep))
           return from
 


antfu:
I am not sure about the complexity here. Is it possible to have a simpler solution? Tests are always preferred especially with RegExp.

Simon-He95:
> I am not sure about the complexity here. Is it possible to have a simpler solution? Tests are always preferred especially with RegExp.我不确定这里的复杂性。是否有可能有一个更简单的解决方案？测试始终是首选，尤其是对于正则表达式。



> I am not sure about the complexity here. Is it possible to have a simpler solution? Tests are always preferred especially with RegExp.我不确定这里的复杂性。是否有可能有一个更简单的解决方案？测试始终是首选，尤其是对于正则表达式。

Because after this adjustment, all test conditions should be wrapped on the element in order to be correctly parsed. I would like to ask first whether this method is the best way. What we should need to match is <div group-(x1 x2 )>exclude: group-(x1 x2) </div>, so this is the solution I am currently thinking of, so all the previous regular expressions need to be corrected and placed on the attributes of the element.

I feel like I can only think of using this method without using parser.

packages/core/src/utils/variant-group.ts
```diff
@@ -17,18 +17,29 @@ interface VariantGroup {
   items: HighlightAnnotation[]
 }
 
-export function parseVariantGroup(str: string | MagicString, separators = ['-', ':'], depth = 5) {
+export function parseVariantGroup(str: string | MagicString, separators = ['-', ':'], depth = 5, type?: string) {
```
antfu, line 20: Where the `type` used for, other than checking if it presents? Should be make a literal enum like `'shourtcut' | 'rule' | ...`?

Simon-He95: I changed it to a boolean type. It seems that I only need to determine whether I need to enable the conversion of attributes. I changed it to a boolean type. It seems that I only need to determine whether I need to enable the conversion of attributes.

packages/core/src/utils/variant-group.ts
```diff
@@ -17,18 +17,29 @@ interface VariantGroup {
...
+export function parseVariantGroup(str: string | MagicString, separators = ['-', ':'], depth = 5, onlyAttribute?: boolean) {
   const regexClassGroup = makeRegexClassGroup(separators)
   let hasChanged
   let content = str.toString()
   const prefixes = new Set<string>()
   const groupsByOffset = new Map<number, VariantGroup>()
-
+  const processedStr = (str: string) => str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+  const isAttributor = (matchText: string, offset: number) => {
+    const reg = new RegExp(`<[\\w_\\-]+[^>\\/]*${processedStr(matchText)}[^>\\/]*\/?>`, 'gm')
```
antfu, line 28: I am not sure about dynamic constructed RegExp, it would hurt the performance and also bring potential security concerns when we construct it from user input

Simon-He95: I tried to write a static regular expression for processing, and it seemed to pass all the tests. What do you think?

Bobakanoosh:
Any update on this MR? would love to be able to use variant groups with ts files

PR #3213, opened 2023-10-10
fix(cli): generate styles in text that should be skipped

Simon-He95:
I found that the content in @unocss-skip will still be styled. I think that before generating css, the input content in @unocss-skip can be replaced with an empty string, and then analyzed, it should be able to achieve the expectations.


zyyv (approved):
BTW, i don't like this test code, just like i don't like react.

Simon-He95:
> BTW, i don't like this test code, just like i don't like react. [BTW, i don't like this test code, just like i don't like react.]

I saw a problem with an issue, and then I thought I could use this method to avoid generating redundant styles. Then I found that the skip did not take effect, so I took his code to reproduce the demo. I don’t like react too. haha

johannschopplich (approved):

Issue #3214, opened 2023-10-10
Make `extendTheme` to accept a simple theme object rather than force a function


xsjcTony:
### Clear and concise description of the problem

Imagine this use case: I want to add a `color` called "test": [Playground](https://unocss.dev/play/?html=DwEwlgbgBA7gTgQwA5IKZwHwCgpVJKAF1QA9CBaYgZ0IwFUA7AewGEBlN4AenAm297YgA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAKq4IlAL5wsUCCDgByLpIDGdOsqRIMAD0ix02AIZcANvEw58REsFIAKZKjoALaDHU86ALkQqVGUAdyhTMDAMKGUA0IBaAEYABmSANxC4dyTUtPc5CwM4YBgMEDp49QxcUqg4ACsuOl4BSuranVRpITgDUtw0ABV3MownGBGQDABKV1Q4dWI6CEKAOgsIZwnR6aC4WhguKFxA%2BfnVi%2B2pnrPFjah-ObPUUua42jRO5%2Bk9n66ekQMGD%2BOAAbT2gPEkic0xu1FoQJYbE4PD4-BhPQAukhpLsgA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4IgLgTghgdgzgMwPYQLYgFwKgGzgUwF8g)

However, due to `colors` overwrites the original `theme.colors`, hence other default color like `blue` is not usable anymore.

[...]

PR #3215, opened 2023-10-10
chore: inherit baseTheme by default

Simon-He95:
related: #3214

I think this way will be more friendly and convenient to use that Merge initialized themes through custom attributes.

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -120,10 +120,9 @@ export function resolveConfig<Theme extends object = object>(
     .reverse() as ResolvedConfig<Theme>['rulesDynamic']
 
   let theme: Theme = mergeThemes(sources.map(p => p.theme))
-
   const extendThemes = getMerged('extendTheme')
   for (const extendTheme of extendThemes)
-    theme = extendTheme(theme) || theme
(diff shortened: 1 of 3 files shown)


zyyv:
This is not a good idea and it prevents me from clearing the uno theme.

```ts
extendTheme: () => ({}), // Not work.
```

Simon-He95:
> This is not a good idea and it prevents me from clearing the uno theme.
> 
> ```ts
> extendTheme: () => ({}), // Not work.
> ```

I want to know why you need to clear the default uno theme. Shouldn't it be used as the base theme and then extended?

zyyv:
We should keep `extendTheme` simple and have no side effects.

Simon-He95:
> We should keep `extendTheme` simple and have no side effects. And clear the theme just a example.

I personally think tailwind's merge is more reasonable, just like a component library. I hope to use its components, but for example, the color is not what I expected, so I overwrite it instead of writing a custom style. , causing the styles of its default component library to no longer take effect, and it is necessary to merge its default styles in customized places.

zyyv:
I think what you mean is more appropriate using the `theme` option, `extendTheme` means you have full control over the theme, different from tailwind.

Simon-He95:
Oh, I see, I thought they were both doing the same thing

Issue #3218, opened 2023-10-11
[🐞color]: `hsl` and `rgb` is opinionated to the legacy comma-separated syntax


xsjcTony:
### UnoCSS version

latest

## This issue is pretty critical.

[...]

PR #3221, opened 2023-10-11
fix(color)!: keep hsla and rgba consistent with tailwindcss

Simon-He95:
close: #3218

https://github.com/tailwindlabs/tailwindcss/blob/master/src/util/color.js#L83


chu121su12:
Theres more to update. Please see https://github.com/chu121su12/unocss/tree/space-color-syntax

Simon-He95:
> Theres more to update. Please see https://github.com/chu121su12/unocss/tree/space-color-syntax还有更多内容需要更新。请参阅 https://github.com/chu121su12/unocss/tree/space-color-syntax

Thanks for the reminder, I've made some updates

packages/rule-utils/src/colors.ts
```diff
@@ -58,9 +55,8 @@ export function colorToString(color: CSSColorValue | string, alphaOverride?: str
   alpha = alphaOverride ?? alpha
   type = type.toLowerCase()
 
-  // Comma separated functions
-  if (['hsla', 'hsl', 'rgba', 'rgb'].includes(type))
-    return `${type.replace('a', '')}a(${components.join(',')}${alpha == null ? '' : `,${alpha}`})`
+  if (['hsla', 'rgba'].includes(type))
+    return `${type}(${components.join(' ')}${alpha == null ? '' : ` / ${alpha}`})`
```
xsjcTony, line 59: Since we clearly agreed that we are going to align with Tailwind, I think this should use COMMA-SEPARATED syntax instead of SPACE-SEPARATED syntax.

Tailwind uses SPACE for `rgb` and `hsl`, but COMMA for `rgba` and `hsla`.

I can see that through out the MR we are changing everything to SPACE syntax, as well as `rgba` and `hsla`. This will work in the browser but that's not aligned with Tailwind.

xsjcTony: This will break for those who uses COMMA syntax with `rgba` and `hsla` (It makes sense to use legacy color function with legacy syntax? 🤣)

Simon-He95: I misunderstood it at first. I used tailwind play to check the output results and made adjustments.

packages/rule-utils/src/colors.ts
```diff
@@ -58,9 +55,11 @@ export function colorToString(color: CSSColorValue | string, alphaOverride?: str
   alpha = alphaOverride ?? alpha
   type = type.toLowerCase()
 
-  // Comma separated functions
-  if (['hsla', 'hsl', 'rgba', 'rgb'].includes(type))
-    return `${type.replace('a', '')}a(${components.join(',')}${alpha == null ? '' : `,${alpha}`})`
+  if (['hsla', 'rgba'].includes(type))
+    return `${type}(${components.join(', ')}${alpha == null ? '' : `, ${alpha}`})`
```
chu121su12, line 59: No need space after comma, just as the previous logic.

packages/rule-utils/src/colors.ts
```diff
@@ -58,9 +55,11 @@ export function colorToString(color: CSSColorValue | string, alphaOverride?: str
...
+
+  if (['rgb', 'hsl'].includes(type))
```
chu121su12, line 61: This condition and its branch is no longer necessary. The continuation after will format the rgb and hsl as usual

packages/preset-mini/src/_rules/ring.ts
```diff
@@ -8,8 +8,8 @@ export const ringBase = {
   '--un-ring-offset-width': '0px',
   '--un-ring-offset-color': '#fff',
   '--un-ring-width': '0px',
-  '--un-ring-color': 'rgba(147,197,253,0.5)',
-  '--un-shadow': '0 0 rgba(0,0,0,0)',
+  '--un-ring-color': 'rgba(147, 197, 253, 0.5)',
```
chu121su12, line 11: I think it's better to update all presets' colors to rgb + space too

packages/rule-utils/src/colors.ts
Simon-He95: In tailwind play, it will generate to `rgb`, if remove condition that `bg-red-500` will be generated to `rgba`
image

packages/rule-utils/src/colors.ts
Simon-He95: > No need space after comma, just as the previous logic.逗号后不需要空格，就像前面的逻辑一样。

Wouldn't it be more organized if there are spaces? I refer to tailwind's processing of color.

xsjcTony: I personally think this doesn't matter since the browser doesn't care about it at all. If less space means a smaller file size, then I agree with that, if not, I'd stand with adding the space to make it look better.

shellscape:
@antfu @Simon-He95 Gmail explicitly doesn't support space separated color function values, so when using uno tomgenerate cas for HTML used in email, we get undesirable rendering. What's the solution (if any) tomrevert to comma separated?

Issue #2572, opened 2023-05-05
Compatibility issue with border-[...]


relferreira:
### UnoCSS version

0.51.8

### Describe the bug

I'm migrating my project from Tailwind, and I saw a strange behavior with `border-[...]`.  Uno is picking `border-color,` but it should be `border-width` based on Tailwind's doc.

Reproduction: https://unocss.dev/interactive/?s=border-[6px]
Tailwind Docs: https://tailwindcss.com/docs/border-width#arbitrary-values

Is this the expected behavior?

Thanks in advance.

### Reproduction

https://unocss.dev/interactive/?s=border-[6px]

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

cesswhite:
Hi, I have a issue similar with border-


example normal class:
_//Dont work_
```
<div class="border-2 border-green-500">
 </div>
```
example custom color class:
_//Dont work_
```
<div class="border-2 border-brand-green">
 </div>
```

`_unocss.config.ts_`
```
import { defineConfig, presetUno, presetAttributify } from 'unocss'

export default defineConfig({
    theme: {
        colors: {
            'brand': {
                'green': '#036631',
                'yellow': '#D9E021',
                'gray': '#4D4D4D'
            }
        }
    },
    presets: [
        presetAttributify(),
        presetUno()
    ]
})
```

_`main.ts`_
`import 'vritual:uno.css'`

henrikvilhelmberglund:
This is still broken when using `calc` for example `border-[calc(var(--border-width)*1px)]`.

zyyv:
> This is still broken when using `calc` for example `border-[calc(var(--border-width)*1px)]`.

`border-[xxx]` gives priority to colors. As a workaround, you can use `border-width-[xxx]`   
https://unocss.dev/interactive/?s=border-width-[calc(1px%2B2px)]

PR #3230, opened 2023-10-14
fix(preset-mini): border-[calc(xx)] is generated to border-color

Simon-He95:
related: #2572


packages/preset-mini/src/_rules/border.ts
```diff
@@ -80,6 +80,12 @@ function borderColorResolver(direction: string) {
       }
     }
     else if (color) {
+      if (color.startsWith('calc')) {
+        return {
+          'border-width': color,
+        }
+      }
```
antfu, line 87: This does not sound like a long term solution. Do you have reference in TW's codebase?

henrikvilhelmberglund: There are other utilities with the same issue like `outline` so you would have to go utility by utility this way.

I looked at the TW code and it seems like they have a regex checking for `min`, `max`, `clamp` or `calc`. Not sure if it could work here too
reference: https://github.com/tailwindlabs/tailwindcss/blob/master/src/util/dataTypes.js#L9

Simon-He95: > There are other utilities with the same issue like `outline` so you would have to go utility by utility this way. [There are other utilities with the same issue like outline so you would have to go utility by utility this way.]
> 
> I looked at the TW code and it seems like they have a regex checking for `min`, `max`, `clamp` or `calc`. [I looked at the TW code and it seems like they have a regex checking for min, max, clamp or calc.] Not sure if it could work here too [Not sure if it could work here too] reference: [https://github.com/tailwindlabs/tailwindcss/blob/master/src/util/dataTypes.js#L9 [reference: https://github.com/tailwindlabs/tailwindcss/blob/master/src/util/dataTypes.js#L9] ](https://github.com/tailwindlabs/tailwindcss/blob/master/src/util/dataTypes.js#L9)

thx, I add `isCSSFunction` to replace `color.startsWith('calc')`

packages/preset-mini/src/_utils/utilities.ts
```diff
@@ -243,3 +243,7 @@ export function resolveVerticalBreakpoints({ theme, generator }: Readonly<Varian
 export function makeGlobalStaticRules(prefix: string, property?: string): StaticRule[] {
   return globalKeywords.map(keyword => [`${prefix}-${keyword}`, { [property ?? prefix]: keyword }])
 }
+
+export function isCSSFunction(value: string) {
+  return cssFunctions.some(fn => new RegExp(`^${fn}\\(.*\\)`).test(value))
```
antfu, line 248: We should construct a simple regex, and cache it as a singleton. Instead of creating dynamic regex in every loop.

antfu (changes requested):

packages/preset-mini/src/_utils/utilities.ts
Simon-He95: wow，great insight

packages/preset-mini/src/_utils/mappings.ts
```diff
@@ -113,3 +113,5 @@ export const globalKeywords = [
   'revert-layer',
   'unset',
 ]
+
+export const cssFunctions = ['min', 'max', 'clamp', 'calc'].map(fn => new RegExp(`^${fn}\\(.*\\)`))
```
zyyv, line 117:
```suggestion
export const cssMathFnRE = new RegExp(`^(${['min', 'max', 'clamp', 'calc'].join('|')})\\(.*\\)`)
```

packages/preset-mini/src/_utils/utilities.ts
```diff
@@ -243,3 +243,7 @@ export function resolveVerticalBreakpoints({ theme, generator }: Readonly<Varian
 export function makeGlobalStaticRules(prefix: string, property?: string): StaticRule[] {
   return globalKeywords.map(keyword => [`${prefix}-${keyword}`, { [property ?? prefix]: keyword }])
 }
+
+export function isCSSFunction(value: string) {
```
zyyv, line 247:
```suggestion
export function isCSSMathFn(value: string) {
```

zyyv (commented):
We maintain rigor in variable naming.

packages/preset-mini/src/_utils/mappings.ts
Simon-He95: good refactoring

test/transformer-compile-class.test.ts
```diff
@@ -177,4 +177,23 @@ describe('transformer-compile-class', () => {
 
     expect(invalidateFn).toHaveBeenCalledTimes(2)
   })
+
+  it('test border', async () => {
+    const result = await transform(`
+    <div class="border-[calc(var(--border-width)*1px)]"/>
```
zyyv, line 183: The location file of the test case is incorrect, it should be in `test/assets/preset-mini-targets.ts`

zyyv (changes requested):

test/transformer-compile-class.test.ts
Simon-He95: thx~

zyyv (approved):

packages/preset-mini/src/_utils/mappings.ts
```diff
@@ -113,3 +113,5 @@ export const globalKeywords = [
   'revert-layer',
   'unset',
 ]
+
+export const cssMathFnRE = new RegExp(`^(${['min', 'max', 'clamp', 'calc'].join('|')})\\(.*\\)`)
```
antfu, line 117: It's not dynamic, let's just write it as regex literal

antfu (approved):

Issue #3076, opened 2023-09-04
[✨Feature Request] Support `@hover` in `group-hover`


xsjcTony:
### Clear and concise description of the problem

First of all, thanks for the `@hover` experimental feature, as this is one of the main reasons I'm migrating.
However, I found that seems `group-hover:` does not support that.
None of `group-@hover:`, `group:@hover:`, `@group-hover:`, `@group:hover:` works for me.

💚

### Suggested solution

`group-@hover:` or `group:@hover:`

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

chu121su12:
Please see example in usage & test in #3235

PR #3233, opened 2023-10-14
feat(preset-mini): add default mouse media as alternative to @hover

chu121su12:
Due to how tagged pseudo works (`group-x`, etc), it is difficult to weave `@hover` support.

This PR copy one of `media-x` variant default theme (`mouse`) from preset-wind so we can use `media-mouse` variant in preset-mini to add the `@hover` media query without adding the `:hover` pseudo. Note that `media-x` is separate from `group-x`, so to combine you may need to, well, combine the syntax (see test) or use shortcut. I personally think that this combination is better served via shortcut to keep the atomicity of the variant.

Reference: https://github.com/unocss/unocss/blob/5dcc8ceddb7b6acbb06be914b5f1bd3a444c8976/packages/preset-wind/src/theme.ts#L193

This should help #3076 with only preset-mini.

### DIFF
--- a/packages/preset-mini/src/_theme/default.ts
+++ b/packages/preset-mini/src/_theme/default.ts
@@ -1,6 +1,6 @@
 import { colors } from './colors'
 import { fontFamily, fontSize, fontWeight, letterSpacing, lineHeight, textIndent, textShadow, textStrokeWidth, wordSpacing } from './font'
-import { borderRadius, boxShadow, breakpoints, duration, easing, lineWidth, ringWidth, spacing, verticalBreakpoints, zIndex } from './misc'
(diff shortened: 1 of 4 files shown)


antfu:
Would that make sense to be `media-mouse-hover:` or shortcut as `hover-mouse:`, to be more explicit?

/cc @LittleSound I remember you were working on a similar things, wonder if you have any feedback on this?

LittleSound:
> Would that make sense to be `media-mouse-hover:` or shortcut as `hover-mouse:`, to be more explicit?
> 
> /cc @LittleSound I remember you were working on a similar things, wonder if you have any feedback on this?

Responding to hover events on mobile webpages can be quite tricky. To overcome this, I usually write `hover:bg-primary` as `sm:hover:bg-primary` to make it only work on larger screens. Of course, this doesn't truly detect whether it's a mouse or not, it's just generally touchscreen devices have smaller screens. If this PR can provide a better solution, I'm looking forward to it.

PR #3235, opened 2023-10-15
feat(preset-uno): add shortcut for media query-based hover

chu121su12:
Simple shortcuts in preset-uno as alternative to #3233


antfu:
Close in favor of https://github.com/unocss/unocss/pull/3233, I think we'd better to avoid using shortcuts in presets as possible, to leave that to userland. Thanks

PR #3236, opened 2023-10-15
feat(preset-mini)!: rename variant `uno-layer-x` to `sheet-x`

chu121su12:
Breaking changes with the renaming, but only the variant. The terminology, comment label, and core api are still using `layer`.

Also, if the naming is not acceptable, please close the PR.


antfu:
What's the reason of doing so?

PR #3248, opened 2023-10-19

chu121su12:
Reopened #2341, #2344, #2345

> Export all built-in presets as individual preset key/function pair. The main purpose of this is to be able to use presetIcons online without needing to use bundler. With separate presets, each of them can be used with their respective options.

> As for the existing cdns, we can consider deprecating all but core and full. core is the recommended one to be used for integration, while full is for showcase shortcut.


test/preset-icons.test.ts
```diff
@@ -62,4 +62,8 @@ describe('preset-icons', () => {
     expect(css).toContain('data:image/svg+xml;utf8,%3Csvg')
     await expect(css).toMatchFileSnapshot('./assets/output/preset-icons-unit-svg-prologue.css')
   })
+
+  it('exports default collection list', () => {
+    expect(icons).toMatchSnapshot()
```
antfu, line 67: We probably don't need to test it. The list is kinda external, whenever the list updates, it introduce uncessary snapshot changes.

antfu (approved):

PR #3250, opened 2023-10-19
docs: update `svelte-coped` hooks.server.js location

TGlide:
I've had to refer to this page a few times, and since `hooks.server.js` is not a common file, I always forget where to put it, and have to refer to SvelteKit's docs.

My PR adds a simple change that adds the `src` folder to the code snippet. Alternatively, I suggest linking to SvelteKit docs.


zyyv:
> I suggest linking to SvelteKit docs.
Agree

PR #3256, opened 2023-10-21
fix(webpack): escape backslash in css

### LINKED ISSUE #2794: (@unocss/webpack): class util with variant prefixes in webpack preset does not work
### UnoCSS version

0.53.3

### Describe the bug

Class util like `sm:bg-red-500` does not work in  webpack project with `@unocss/webpack`
```html
<template>
  <div class="flex justify-center">
    <p class="sm:bg-red-500">The background should be red but it does not</p>
  </div>
</template>
```
The style append to body is
```css
@media (min-width: 320px) {.sm:bg-red-500{--un-bg-opacity:1;background-color:rgba(239,68,68,var(--un-bg-opacity))} }
```

The expected css result is 
```css
@media (min-width: 320px) { .sm\:bg-red-500{--un-bg-opacity:1;background-color:rgba(239,68,68,var(--un-bg-opacity))} }
```
Maybe something got wrong when write to webpack assets without further escape for the generated css 

### Reproduction

- git clone [repo](https://github.com/zoeyzhao19/unocss-webpack-issue) or view online [sandbox](https://codesandbox.io/p/github/zoeyzhao19/unocss-webpack-issue/master?workspaceId=8a7a8b86-fa11-43ae-9a2d-fc45b02e677c)
-  run `pnpm dev` 

### System Info

- OS: Windows 10 10.0.22621

### Validations

chu121su12:
Close #2794

### DIFF
--- a/packages/webpack/src/index.ts
+++ b/packages/webpack/src/index.ts
@@ -118,7 +118,7 @@ export default function WebpackPlugin<Theme extends object>(
                   : (result.getLayer(layer) || '')
 
                 if (!quote)
-                  return css
+                  return css.replaceAll(/\\/g, '\\\\\\\\')
 
                 // the css is in a js file, escaping
                 let escaped = JSON.stringify(css).slice(1, -1)


packages/webpack/src/index.ts
```diff
@@ -118,7 +118,7 @@ export default function WebpackPlugin<Theme extends object>(
                   : (result.getLayer(layer) || '')
 
                 if (!quote)
-                  return css
+                  return css.replaceAll(/\\/g, '\\\\\\\\')
```
antfu, line 121: Why do we need so many slashes?

chu121su12: I'm not sure but I'm not good with webpack. I found this with trial and error. So if somebody have better webpack-fu, please advice. Having said that, dumping this `RawSource`:

https://github.com/unocss/unocss/blob/5d0a295860edc6c905af831b29ef9050f73086ff/packages/webpack/src/index.ts#L131

We can look for the output which is something like this:
```js
eval(
  /*v this is one single string to be eval-ed,  
       v with inlined backtick */
    "<... webpack content ...>
     ___CSS_LOADER_EXPORT___.push([
       module.id,
       `<... unocss' css in backtick ...>`,
       \"\"]);
    <... more webpack content ...>")
```
I hope you can see where this is going.

The css in the backtick (after this PR) already be outputted as 4x `\`:
image

And my guess is within the 2 quotes, the outer string ("") and the inner literal (\`\`), all the backslashes were reduced down to single `\` for output.

I think more escaping needs to be done with the quoted branch, but let's wait for issue or reproduction.
https://github.com/unocss/unocss/blob/5d0a295860edc6c905af831b29ef9050f73086ff/packages/webpack/src/index.ts#L120-L128

chu121su12:
@high1 please provide issue & reproduction. thank you

PR #3260, opened 2023-10-21
fix(preset-mini): use array cache sorted breakpoints

### LINKED ISSUE #2960: Broken container styles when breakpoints are set in preset and contain a number, like 2xl
### UnoCSS version

v0.54.1

### Describe the bug

**Use Case:**
I have a div. I want this div to get a max-width at 2xl breakpoint.
I use windicss `container` class for this. But prefix it with 2xl breakpoint, like:

`<div class="2xl:container">My container</div>`

Within my preset i defined new breakpoints, as i wanted to add breakpoint xxs (320) and xs (480). I read in the unocss documentation, as soon as providing a breakpoints object it will fully overwrite the default breakpoints, so i have to define them all like on my preset:

```
{
      name: 'my-preset',
      theme: {
        breakpoints: {
          'xxs': '320px',
          'xs': '480px',
          'sm': '640px',
          'md': '768px',
          'lg': '1024px',
          'xl': '1280px',
          '2xl': '1536px',
        },
      }
}
```
**Expectation:**
I would expect the div to be 100% width until breakpoint 2xl is reached and from that on get a max-width equal to 2xl breakpoint. Reference: https://windicss.org/utilities/layout/container.html

**Current behaviour:**
The container class styles are somehow broken when new breakpoints are defined and contain a number within the key, like 2xl, 3xl,..
Instead of applying max-width style when 2xl breakpoint is reached, the max widths for xxs ans xs get set:
Bildschirmfoto 2023-08-08 um 00 38 33

from xs and up the container always has the xs breakpoint max width.

Bildschirmfoto 2023-08-08 um 00 48 16

You can see the behaviour in the unocss playground here: https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIBGBzAtAgW0wFMAnKAF2IA8LcB3ACzCqgOvwFcKB7KAJmpwAXDB4A7CgjDiy6AHwAoKFADCEqTLJQkjHpzggo2YhTYIO9MCAqMoEuAE8oCM5lLEEAawAOPGWbognDoisAA9OAQ8kA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owCqWEcAvnOlBCHAOQCujAYypVuAKFGoAHpFgoMAQ14AbeGkw58hYEQAUCUXHKUaMKgC44AbQOGj1Wgwg6AlCRuH9t21nkhUF7hAATwBaCnsYbjcvQxgAC1Q-C08YwwAjSnkAa0hsU2T3VJ5JSREAgGYAJgAGMEkowtTuUu4AgBYADlr66KLDbioQVp4ANjbuhr7%2BkGRh7gB2EY66yanuJSI5gEZqyraV3r7mpW3KroPGmO5KyROArYBWcpGLvuZD22ZCr8MAXTdmM4gA&css=Q&options=N4IgzgLgTglgxhEAuaBXApgXyA

It does NOT happen when breakpoint keys are without numbers, like here: https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIBGBzAtAgW0wFMAnKAF2IA8LcB3ACzCqgOvwFcKB7Ka6nABcMHgDsKCMGLLoAfACgoUAMLjJ0slCSMenOCCjZiFNgg70wICoyji4ATygJTmUsQQBrAA49pp9AAmQXQFYAB6cAg5IA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owCqWEcAvnOlBCHAOQCujAYypVuAKFGoAHpFgoMAQ14AbeGkw58hYEQAUCUXHKUaMKgC44AbQOGj1Wgwg6AlCRuH9t21nkhUF7hAATwBaCnsYbjcvQxgAC1Q-C08YwwAjSnkAa0hsU2T3VJ5JSREAgGYAJgAGMEkowtTuUu4AgBYADlr66KLDbioQVp4ANjbuhr7%2BkGRh7gB2EY66yanuJSI5gEZqyraV3r7mpW3KroPGmObJE4CtgFZykYu%2B5kPbZkLPwwBdN2ZnEA&css=Q&options=N4IgzgLgTglgxhEAuaBXApgXyA

It also does not happen when not defined as preset, like here: https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIBGBzAtAgW0wFMAnKAF2IA8LcB3ACzCqgOvwFcKB7KAJmpwAXDB4A7CgjDiy6AHwAoKFADCEqTLJQkjHpzggo2YhTYIO9MCAqMoEuAE8oCM5lLEEAawAOPGWbognDoisAA9OAQ8kA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owCqWEcAvnOlBCHAOQCujAYypVuAKFGoAHpFgoMAQ14AbeGkw58hYEQAUCUXDgwAFqhCoAXIgOHbhgEaV5Aa0jYYVK-rs%2BekySJW3ADMAEwADGCS3CQ2vobcAdxBACwAHJHRsfG23FQgyTwAbCmZMXHx3CDIhdwA7EVpUeU5CUpEtQCM4aEpzdmtiUpdoRn9Fb7coZLDQZ0ArMFF4znMA3bMoswAlEA&css=Q&options=N4IgzgLgTglgxhEAuaBXApgXyA

### Reproduction

https://uno.antfu.me/play/?html=DwEwlgbgBAxgNgQwM5ILwCIBGBzAtAgW0wFMAnKAF2IA8LcB3ACzCqgOvwFcKB7KAJmpwAXDB4A7CgjDiy6AHwAoKFADCEqTLJQkjHpzggo2YhTYIO9MCAqMoEuAE8oCM5lLEEAawAOPGWbognDoisAA9OAQ8kA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owCqWEcAvnOlBCHAOQCujAYypVuAKFGoAHpFgoMAQ14AbeGkw58hYEQAUCUXHKUaMKgC44AbQOGj1Wgwg6AlCRuH9t21nkhUF7hAATwBaCnsYbjcvQxgAC1Q-C08YwwAjSnkAa0hsU2T3VJ5JSREAgGYAJgAGMEkowtTuUu4AgBYADlr66KLDbioQVp4ANjbuhr7%2BkGRh7gB2EY66yanuJSI5gEZqyraV3r7mpW3KroPGmO5KyROArYBWcpGLvuZD22ZCr8MAXTdmM4gA&css=Q&options=N4IgzgLgTglgxhEAuaBXApgXyA

### System Info

Macbook Pro, Apple M2 Max, MacOS Ventura Version 13.4.1
Google Chrome Version 115.0.5790.114 (Offizieller Build) (arm64)

### Validations

zyyv:
close #2960

In #2960 case, it use breakpoints with a number key as screen point. We use `Object.keys` get all points without correct size sorting.

I can only compare the boundary numbers of points in advance and use an array to maintain regular ordering.

As for points in different units, we can explain them in the documentation and let users align and sort the units themselves.

### DIFF
--- a/packages/preset-mini/src/_rules/size.ts
+++ b/packages/preset-mini/src/_rules/size.ts
@@ -43,8 +43,8 @@ export const sizes: Rule<Theme>[] = [
       '(max|min)-(w|h)-full',
     ],
   }],
-  [/^(?:size-)?(min-|max-)?(h)-screen-(.+)$/, ([, m, w, s], context) => ({ [getPropName(m, w)]: handleBreakponit(context, s, 'verticalBreakpoints') })],
-  [/^(?:size-)?(min-|max-)?(w)-screen-(.+)$/, ([, m, w, s], context) => ({ [getPropName(m, w)]: handleBreakponit(context, s) }), {
+  [/^(?:size-)?(min-|max-)?(h)-screen-(.+)$/, ([, m, w, s], context) => ({ [getPropName(m, w)]: handleBreakpoint(context, s, 'verticalBreakpoints') })],
+  [/^(?:size-)?(min-|max-)?(w)-screen-(.+)$/, ([, m, w, s], context) => ({ [getPropName(m, w)]: handleBreakpoint(context, s) }), {
     autocomplete: [
       '(w|h)-screen',
       '(min|max)-(w|h)-screen',
@@ -56,7 +56,7 @@ export const sizes: Rule<Theme>[] = [
   }],
 ]
 
-function handleBreakponit(context: Readonly<RuleContext<Theme>>, screen: string, key: 'breakpoints' | 'verticalBreakpoints' = 'breakpoints') {
+function handleBreakpoint(context: Readonly<RuleContext<Theme>>, screen: string, key: 'breakpoints' | 'verticalBreakpoints' = 'breakpoints') {
   const bp = resolveBreakpoints(context, key)
   if (bp)
     return bp.find(i => i.point === screen)?.size
--- a/packages/preset-mini/src/_rules/size.ts
+++ b/packages/preset-mini/src/_rules/size.ts
@@ -43,8 +43,8 @@ export const sizes: Rule<Theme>[] = [
       '(max|min)-(w|h)-full',
     ],
   }],
-  [/^(?:size-)?(min-|max-)?(h)-screen-(.+)$/, ([, m, w, s], context) => ({ [getPropName(m, w)]: handleBreakpoint(context, s, 'verticalBreakpoints') })],
-  [/^(?:size-)?(min-|max-)?(w)-screen-(.+)$/, ([, m, w, s], context) => ({ [getPropName(m, w)]: handleBreakpoint(context, s) }), {
+  [/^(?:size-)?(min-|max-)?(h)-screen-(.+)$/, ([, m, h, p], context) => ({ [getPropName(m, h)]: handleBreakpoint(context, p, 'verticalBreakpoints') })],
+  [/^(?:size-)?(min-|max-)?(w)-screen-(.+)$/, ([, m, w, p], context) => ({ [getPropName(m, w)]: handleBreakpoint(context, p) }), {
     autocomplete: [
       '(w|h)-screen',
       '(min|max)-(w|h)-screen',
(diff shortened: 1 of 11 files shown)


packages/preset-mini/src/_rules/size.ts
```diff
@@ -56,6 +56,12 @@ export const sizes: Rule<Theme>[] = [
   }],
 ]
 
+function handleBreakponit(context: Readonly<RuleContext<Theme>>, screen: string, key: 'breakpoints' | 'verticalBreakpoints' = 'breakpoints') {
```
userquin, line 59: typo

Issue #3196, opened 2023-10-03
ring-offset-0 missing unit


DSeeLP:
### UnoCSS version

0.55.7

### Describe the bug

`ring-offest-0` is missing unit in generated css and declaring a unit explicitly doesn't work either `ring-offset-0px`.

CSS output from playground from reproduction:
```css
.ring-offset-0,
.ring-offset-0px,
.ring-offset-0rem{--un-ring-offset-width:0;}
```

### Reproduction

[...]

PR #2153, opened 2023-02-05
fix(preset-mini): remove units about `0` and `none` (#2153)close https://github.com/unocss/unocss/issues/2152

zyyv:
close #2152

PR #3271, opened 2023-10-24
fix(preset-mini): remove num === 0 in px handler to fix offset utility issues

henrikvilhelmberglund:
This PR removes a `num === 0` check in the px handler which returned `0` which broke a bunch of offset-utilities (`ring-offset-0`, `underline-offset-0`). This means it returns `0px` now and these utilities work again.

The `num === 0` check is still there for the rem handler so `inset-0` still gives a `0` result. (I'm not sure why this is better than 0px though)

It's possible there's a better way to fix this, in that case feel free to close this PR.

fixes https://github.com/unocss/unocss/issues/3196

### DIFF
--- a/packages/preset-mini/src/_utils/handlers/handlers.ts
+++ b/packages/preset-mini/src/_utils/handlers/handlers.ts
@@ -106,11 +106,8 @@ export function px(str: string) {
     return
   const [, n, unit] = match
   const num = Number.parseFloat(n)
-  if (!Number.isNaN(num)) {
-    if (num === 0)
-      return '0'
+  if (!Number.isNaN(num))
     return unit ? `${round(num)}${unit}` : `${round(num)}px`
-  }
 }
 
 export function number(str: string) {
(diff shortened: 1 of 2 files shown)


zyyv:
> The num === 0 check is still there for the rem handler so inset-0 still gives a 0 result. (I'm not sure why this is better than 0px though)

Cause #2153 changes to produce this issue that i'm not consided it. And thanks for your feedback, I'll consider if there's a better way to fix it or revert it.

henrikvilhelmberglund:
A fun(?) thing I noticed is that `decoration-offset-0` in UnoCSS seems broken in Brave, I get something like this. Very strange...
image

antfu (approved):

henrikvilhelmberglund:
> > The num === 0 check is still there for the rem handler so inset-0 still gives a 0 result. (I'm not sure why this is better than 0px though)
> 
> Cause #2153 changes to produce this issue that i'm not consided it. And thanks for your feedback, I'll consider if there's a better way to fix it or revert it.

Do you want to merge it like this or did you have some other solution in mind?

zyyv (approved):

PR #3286, opened 2023-10-29
feat(svelte-scoped): transform class directives wrapped in quotes

fehnomenal:
I need this because I have to use a preprocessor that wraps attribute values from `foo={bar}` to `foo="{bar}"` and breaks the class directive transformation.


antfu:
Could you add some tests to prevent future regressions? Thanks

fehnomenal:
@antfu Done!

jacob-8:
Looks good to me. Your use case makes sense though I hope people don't write it this way as the quotes are unnecessary and not in the Svelte documentation.

I cleaned up a few test titles while we were in there.

Issue #3293, opened 2023-10-31
Confusion over the order of the same rules


yveeven:
### UnoCSS version

0.57.1

### Describe the bug

As with the minimal implementation, it's not retrieved from the beginning nor the end, instead it uses red in the middle, but that's not to say that it's necessarily the middle, my feeling seems to be that there's a set of default rules in order for a reason, but that doesn't cover the same rules

### Reproduction

https://stackblitz.com/edit/nuxt-starter-a4phwp?file=app.vue

### System Info

  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

yveeven:
I noticed your solution and I am now asking how the solution is, thank you

chu121su12:
By default, output of those kind of rules (as in example, `bg-<color>`) are sorted alphabetically

chu121su12:
> I noticed your solution and I am now asking how the solution is, thank you

Try #3261

yveeven:
> By default, output of those kind of rules (as in example, `bg-<color>`) are sorted alphabetically

First of all I'm not defining my own rules, I'm just using the official presets, so I don't think it has anything to do with layers!

Then I would like to point out that the current unocss implementation does not match the order explicitly stated in the documentation

Also there is an unclear order relationship in the case of non-colour values as in the minimal implementation of two/two-1 positions, if it is alphabetical then why is `rounded-sm` a lower priority than `xl`.

Secondly alphabetical ordering is not as intuitive as the documented priority of the latter, and the main point I'm trying to make is that priority is in effect when there are multiple occurrences of the same rule, whereas alphabetical ordering of a collection of rules doesn't seem to be a problem to me.


[...]

PR #3294, opened 2023-10-31
fix(generator): order in which css is generated

Simon-He95:
close: #3293


chu121su12:
I think it's better to have the output sorted and use the layering utility and other variants to modify the priority/specificity.

Otherwise, looking at the updated css output, this should be tagged as breaking change.

Simon-He95:
> I think it's better to have the output sorted and use the layering utility and other variants to modify the priority/specificity. [I think it's better to have the output sorted and use the layering utility and other variants to modify the priority/specificity.]
> 
> Otherwise, looking at the updated css output, this should be tagged as breaking change. [Otherwise, looking at the updated css output, this should be tagged as breaking change.]

The output results of sorting in alphabetical order are indeed relatively good-looking, but they ignore the issue of class order coverage, which may cause some hidden dangers of styles being overwritten and not taking effect. Is it possible to consider that some shortcuts are not layered separately, but retain their order and put comments on the generated css to indicate that it comes from shortcuts?

chu121su12:
> The output results of sorting in alphabetical order are indeed relatively good-looking, but they ignore the issue of class order coverage, which may cause some hidden dangers of styles being overwritten and not taking effect.

It is the consequence of css using classes. Consider this example with 2 divs:

```html
<div class="bg-red bg-blue">main article</div>
<!-- ... later in html -->
<div class="bg-blue bg-red">alert with red background</div>
```

How should the css got generated? With deterministic alphabetical sorting, red would always be shown and the dev can use other workaround

> Is it possible to consider that some shortcuts are not layered separately, but retain their order and put comments on the generated css to indicate that it comes from shortcuts?

this would be a welcome addition, especially in dev mode

Simon-He95:
Indeed, it seems that we have to leave it to the developers to solve this problem themselves.

PR #3284, opened 2023-10-27
feat(preset-mini): add subgrid rules in preset mini

henrikvilhelmberglund:
This PR adds subgrid rules in preset mini: 
`grid-cols-subgrid`
`grid-rows-subgrid`

> [!NOTE]
> Subgrid explanation: https://web.dev/articles/css-subgrid

> [!WARNING]
> Note that subgrid isn't available in all browsers yet. https://caniuse.com/css-subgrid

PR #3301, opened 2023-11-02
fix(preset-mini): change grid-template-cols to grid-template-columns

henrikvilhelmberglund:
I noticed I made a mistake in https://github.com/unocss/unocss/pull/3284 by writing `grid-template-cols` in the rule instead of `grid-template-columns` 😱

This PR fixes that so the resulting CSS is valid. Would probably be helpful to have some tests for these too but I have no idea where to add them.

### DIFF
--- a/packages/preset-mini/src/_rules/grid.ts
+++ b/packages/preset-mini/src/_rules/grid.ts
@@ -74,5 +74,5 @@ export const grids: Rule<Theme>[] = [
 
   // template subgrid
   ['grid-rows-subgrid', { 'grid-template-rows': 'subgrid' }],
-  ['grid-cols-subgrid', { 'grid-template-cols': 'subgrid' }],
+  ['grid-cols-subgrid', { 'grid-template-columns': 'subgrid' }],
 ]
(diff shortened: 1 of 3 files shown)


henrikvilhelmberglund:
I tried to add a test(?) but got this
```
 FAIL  |unit| test/countable-set.test.ts > countableSet > getCount
AssertionError: expected +0 to be 1 // Object.is equality

- Expected
+ Received

- 1
+ 0

 ❯ test/countable-set.test.ts:44:32
     42|     const s = new CountableSet(['bar1', 'bar2', 'bar2'])
     43|
     44|     expect(s.getCount('bar1')).toBe(1)
       |                                ^
     45|     expect(s.getCount('bar2')).toBe(2)
     46|   })
```
so reverted that and am pretending nothing happened. Anyway this PR fixes the issue with the `grid-cols-subgrid` utility.

userquin (approved):

henrikvilhelmberglund:
@userquin Do you know why CI failed in the commit I tried? Your one basically looks the same but with the order swapped in one of them and seems to work. 🤯

userquin:
You need to update also the css file or run `nr test:update`

henrikvilhelmberglund:
> You need to update also the css file or run `nr test:update`.
> 
> No idea about the order, I just run test:update script.

I did press `u` in watch mode but I guess it didn't work. Will try `test:update` the next time, thanks for fixing it!

PR #3337, opened 2023-11-15
fix(vscode): rootRegisterAuto broken

Simon-He95:


packages/vscode/src/index.ts
```diff
@@ -33,7 +33,7 @@ export async function activate(ext: ExtensionContext) {
 
   const ctx = (Array.isArray(root) && root.length)
     ? await rootRegisterManual(ext, root, projectPath, status)
-    : await rootRegisterAuto(ext, typeof root === 'string' ? root : projectPath, config, status)
+    : await rootRegisterAuto(ext, typeof root === 'string' ? root : projectPath, config, status, projectPath)
```
antfu, line 36: Should we resolve root in the second argument directly?

Simon-He95: sure

Simon-He95:
I think rootRegisterManual and rootRegisterAuto seem to be merged into one. I will open a new pr to reconstruct them later.

Issue #3339, opened 2023-11-15
preset里添加边框复合写法的支持


cunzaizhuyi:
### Clear and concise description of the problem

目前只支持边框单个属性设置，建议支持复合写法；

比如，对于border-bottom: 1px solid #ccc;
如果使用现有的class，只能写成class="b-b-1 b-b-solid b-b-#ccc"，比较麻烦（事实上，我经常因为这个原因而把边框的设置放到css里去写，而非直接在div上写class）。
建议可以简写成诸如：class="b-b-1px-solid-#ccc"的形式；

同理，对于border: 1px solid #ccc;
希望能简写成：class="b-1px-solid-#ccc"的形式；

### Suggested solution

因为不熟悉unocss项目的结构，可能还要涉及改文档，代码规范之类的问题，所以如果维护者觉得本提议可以支持，可以照以下代码帮忙实现一下。

```
// 抄自unocss
export const directionMap = {
  l: ['-left'],
  r: ['-right'],
  t: ['-top'],
  b: ['-bottom'],
  s: ['-inline-start'],
  e: ['-inline-end'],
  x: ['-left', '-right'],
  y: ['-top', '-bottom'],
  '': [''],
  bs: ['-block-start'],
  be: ['-block-end'],
  is: ['-inline-start'],
  ie: ['-inline-end'],
  block: ['-block-start', '-block-end'],
  inline: ['-inline-start', '-inline-end'],
};


rules: [
        [/^b-([rltbxy])-(\w+)-(\w+)-([#\w]+)$/, (match) => {
          const resultMap = {};
          directionMap[match[1]].forEach((direction) => {
            resultMap[`border${direction}`] = `${match[2]} ${match[3]} ${match[4]}`;
          });
          return resultMap;
        }],
        [/^b-(\w+)-(\w+)-([#\w]+)$/, match => ({
          border: `${match[1]} ${match[2]} ${match[3]}`,
        })],
      ],
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [ ] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Simon-He95:
First option:
[...]

PR #3342, opened 2023-11-16
fix(preset-mini): border(-[trlb])?-[a_b_c]

Simon-He95:
related: #3339

### DIFF
--- a/packages/preset-mini/src/_rules/border.ts
+++ b/packages/preset-mini/src/_rules/border.ts
@@ -1,7 +1,7 @@
 import type { CSSEntries, CSSObject, Rule, RuleContext } from '@unocss/core'
 import { colorOpacityToString, colorToString } from '@unocss/rule-utils'
 import type { Theme } from '../theme'
-import { cornerMap, directionMap, globalKeywords, h, hasParseableColor, isCSSMathFn, parseColor } from '../utils'
+import { cornerMap, directionMap, globalKeywords, h, hasParseableColor, isCSSMathFn, isSize, parseColor } from '../utils'
 
 export const borderStyles = ['solid', 'dashed', 'dotted', 'double', 'hidden', 'none', 'groove', 'ridge', 'inset', 'outset', ...globalKeywords]
 
@@ -56,7 +56,7 @@ function borderColorResolver(direction: string) {
     if (!data)
       return
 
-    const { alpha, color, cssColor } = data
+    const { alpha, color, cssColor, name } = data
 
     if (cssColor) {
       if (alpha != null) {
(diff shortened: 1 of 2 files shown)


packages/preset-mini/src/_rules/border.ts
```diff
@@ -80,6 +80,23 @@ function borderColorResolver(direction: string) {
       }
     }
     else if (color) {
+      if (/\[[^\]]*\]/.test(name)) {
+        const multipleValue = name.slice(1, -1).split('_')
+        if (multipleValue.length) {
+          return multipleValue.reduce((result, v) => {
```
chu121su12, line 86: I'd rather let syntax with spaces not get processed further.

test/preset-mini.test.ts
```diff
@@ -353,4 +353,30 @@ describe('preset-mini', () => {
         .text-sm{font-size:0.875rem;line-height:1.25rem;}"
       `)
   })
+
+  it('border(-[trlb])?-[a_b_c]', async () => {
+    const uno = createGenerator({
```
chu121su12, line 358: Also you can add the test token to the [preset mini target file](https://github.com/unocss/unocss/blob/main/test/assets/preset-mini-targets.ts).

packages/preset-mini/src/_rules/border.ts
Simon-He95: What are you referring to? Can you give me a specific example?

test/preset-mini.test.ts
Simon-He95: good idea

packages/preset-mini/src/_rules/border.ts
chu121su12: like when using `b-[a_b_c]` it should just resolve to `border: a b c` since `a`, `b`, and `c` can appear in any order. See
https://developer.mozilla.org/en-US/docs/Web/CSS/border#formal_syntax
https://developer.mozilla.org/en-US/docs/Web/CSS/Value_definition_syntax#double_bar

Simon-He95: Do you want to `say b-[a_b_c]`, is there something wrong with the order of generation, or do you think the corresponding `border-width`、`border-style`、`border-color` should not be generated but merged into attributes in the `border`?

antfu (changes requested):

test/assets/preset-mini-targets.ts
```diff
@@ -867,6 +867,9 @@ export const presetMiniTargets: string[] = [
   'color-scheme-dark',
 
   // variables
+  'border-[1_red_dashed]',
+  'border-[calc(10px+1px)_red_dashed]',
+  'border-b-[1px_red_solid]',
```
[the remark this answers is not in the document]
Simon-He95: I found that the same is true for tailwind, so just keep it consistent with tailwind and don’t do too much design.

PR #3370, opened 2023-11-22
fix(eslint-plugin): `blocklist` rule cant match variants 

chizukicn:
This PR will add a new function `isBlockedWithStrict` to uno generator and fix the problem that eslint-plugin blocklist cannot match variants

before: It is actually blocked, but eslint-plugin will not prompt
e686d6b8bedf72912f04e5638f26f23d

after:
19a9a217b2426ebf31c57921797c9e25

### DIFF
--- a/packages/eslint-plugin/src/worker.ts
+++ b/packages/eslint-plugin/src/worker.ts
@@ -25,20 +25,43 @@ export async function getGenerator() {
   return await promise
 }
 
+export function setGenerator(generator: Awaited<UnoGenerator<any>>) {
+  promise = Promise.resolve(generator)
+}
+
 async function actionSort(classes: string) {
   return await sortRules(classes, await getGenerator())
 }
 
 async function actionBlocklist(classes: string, id?: string) {
   const uno = await getGenerator()
+  const blocked = new Set<string>()
+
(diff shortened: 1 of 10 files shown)


packages/eslint-plugin/src/worker.ts
```diff
@@ -32,7 +32,9 @@ async function actionSort(classes: string) {
 async function actionBlocklist(classes: string, id?: string) {
   const uno = await getGenerator()
   const extracted = await uno.applyExtractors(classes, id)
-  return [...extracted.values()].filter(i => uno.isBlocked(i))
+  const values = [...extracted.values()]
+  const blocked = await Promise.all(values.map(i => uno.isBlockedWithStrict(i)))
+  return values.filter((_, i) => blocked[i])
```
antfu, line 37: If it's an eslint only feature, how about moving it out of core? Ideally we could have an option for it (I am ok with default strict)

chizukicn: I prefer to support strict mode directly in `isBlocked`. I need to spend some time researching how to implement a simple variants of the sync API

chizukicn:
It seems that it is unrealistic to handle variants synchronously, I moved this function into `eslint-plugin`

Issue #3358, opened 2023-11-20
vite 5.0 Unknown word


Leovenn:
### UnoCSS version

0.57.6

### Describe the bug

Just now, my project upgraded vite 5.0. After I saved the project when using vue-scoped, the vite hot update prompt "Unknown word" was displayed.
![image](https://github.com/unocss/unocss/assets/57058087/5ae9a0b6-a202-4bf7-85a4-71d3347d21e1)
Sorry, I don't know what the problem is, the following example may not report an error when starting the project, but when you add a new class such as "flex" save, it will report an error later
![image](https://github.com/unocss/unocss/assets/57058087/88d82f3c-976d-4e27-a35a-e2156e1e7fd8)
Sometimes this is also suggested

### Reproduction

https://stackblitz.com/edit/vitejs-vite-8bqxui?file=src%2Fcomponents%2FHelloWorld.vue

### System Info

win10 chrome 119.0.6045.160

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Leovenn:
I don't know if it's vite or unocss, but when I turn vue-scoped off, he doesn't report an error

Simon-He95:
> I don't know if it's vite or unocss, but when I turn vue-scoped off, he doesn't report an error

I don't get this error
https://stackblitz.com/edit/vitejs-vite-t3jjwe?file=vite.config.ts,src%2FApp.vue,src%2Fmain.ts,src%2Fcomponents%2FHelloWorld.vue

Leovenn:
you need set "mode: 'vue-scoped'"

Leovenn:
![test](https://github.com/unocss/unocss/assets/57058087/ee34f93b-6dfa-4d08-832d-ce78afa86d02)

Leovenn:
@Simon-He95

coderwyd:
Can this problem be fixed?

HuangZumao:
[...]

PR #3387, opened 2023-11-28
fix: style comparison error when hotUpdate

brenner8023:
FIXED: #3358

I think it need not transform code when hotUpdate. File content comparison when hotUpdate should compare raw code, not transformed code. The transformed code include `unocss:vue-scoped`'s generated code and raw code, generated code is not chaged.

There is a comment from `@vitejs/plugin-vue`
`packages/plugin-vue/src/main.ts`
image

### DIFF
--- a/packages/vite/src/modes/vue-scoped.ts
+++ b/packages/vite/src/modes/vue-scoped.ts
@@ -30,14 +30,5 @@ export function VueScopedPlugin({ uno, ready }: UnocssPluginContext): Plugin {
         return
       return transformSFC(code)
     },
-    handleHotUpdate(ctx) {
-      const read = ctx.read
-      if (filter(ctx.file)) {
-        ctx.read = async () => {
-          const code = await read()
-          return await transformSFC(code) || code
-        }
-      }
-    },
   }
 }


auvred (commented):
If we don't inject styles on `handleHotUpdate`, all applied styles disappear after HMR (you can try to run it locally to see what happens). Most likely it's the bug of `@vitejs/plugin-vue`
Ref: https://github.com/vitejs/vite-plugin-vue/issues/301

brenner8023:
> If we don't inject styles on `handleHotUpdate`, all applied styles disappear after HMR (you can try to run it locally to see what happens). Most likely it's the bug of `@vitejs/plugin-vue` Ref: [vitejs/vite-plugin-vue#301](https://github.sheincorp.cn/vitejs/vite-plugin-vue/issues/301)

I run it locally and everything seems to be ok. Because `unocss:vue-scoped`'s transform hook is called after every HMR. I think applied styles will not disappear.

```js
return {
    name: "unocss:vue-scoped",
    transform(code, id) {
      if (!filter(id))
        return;
      return transformSFC(code); // will run after every HMR
    },
    handleHotUpdate(ctx) {
      // const read = ctx.read;
      // if (filter(ctx.file)) {
      //   ctx.read = async () => {
      //     const code = await read();
      //     return await transformSFC(code) || code;
      //   };
      // }
    }
  };
```

auvred:
Yes, you are right, transform hook is called. But updated styles are not pushed

Reproduction: https://stackblitz.com/edit/vitejs-vite-mnvzcb?file=src%2FApp.vue
Wait for dev server to start and change class `text-red` to `text-blue`. The text should change its color, but it doesn't do so

I just commented out `handleHotUpdate` code (`patch-unocss.mjs`).

But maybe we need to push this change as temporary workaround? I guess he runtime error is worse than missed hot update

coderwyd:
Is there a temporary solution now?

PR #3423, opened 2023-12-05
feat(preset-mini): add `gap-(col|row)-` shorthands

### LINKED ISSUE #3416: column-gap and row-gap shorthand naming
### UnoCSS version

0.58.0

### Describe the bug

Currently, `column-gap` and `row-gap` are output from `gap-x-*` and `gap-y-*` respectively, but if I understand correctly, it looks like an inconsistent behavior.

`column-gap` and `row-gap` are used in multi-column, flex, or grid layouts, thus always tied with logical directions. Other rules as far as I found in the repo with `[xy]` all compile to `-x`/`-y` or `-left|right`/`-top|bottom`, leaving this one as exception.

Since CSSWG is not going to add new physical direction properties regarding text flow, we probably don't need to delete `[xy]` shorthand, but better add synonyms that clearly indicate logical directions.

### Reproduction

See how `px-*` and `gap-x-*` behave differently.

https://unocss.dev/play/?html=DwEwlgbgBA7gTmALgUwLQWXRYDGBDAG1QQHMALRKAMwOQA9ra7UlkBbAZ1R2QDsU4UEngAOqZgCYAfACgoUYCKgjmAFigAjEqjxsNmKQEFgAehGz5i5Ws3bd%2BuFIBCp83IVKRAT1TqtOvQMAYVdZU3AIKSA&config=PTAEFMGcBsEsDsAuBaAJrSBDARtcz5wAPFOQ0eAe2QCcpEbYBjRcVZWAWwAdKbFIAKC69%2BoAN6DQoVOABmCcAGFK8BQHMANFNDc6kcIgCCiBrGwBXRLDkBPbdL1RDASSarID3fsMBVKtoAvqByNJScoADkFlRMkJCRgoLEoogy8pgW0GmyCoQqarDqABSS0jRZUABcoADaOtK1kUwWkIjhtJWRmhKg7tB8NZF0qJGggQC6XlM6kAAWfIgtAjVl0lEtbR3zi8uRQ6wkyNDqoIcofJjw6uCgCwBu4DRV58ismNDdOoFeTgYrdQa3mciH8lGKAEovI4fMZTIxLNY7JDocD-m4PKUgdJIEwPuAagBGAB0ACZUdImKh4EM5qZuJAqiAoJxifNgF91uMoToZoEIYIgA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4IgLgTghgdgzgMwPYQLYAkyoDYgFwJTZwCmAvkA

### System Info

_No response_

### Validations

yheuhtozr:
Closes #3416.


packages/preset-mini/src/_rules/gap.ts
```diff
@@ -20,4 +22,5 @@ function handleGap([, d = '', s]: string[], { theme }: RuleContext<Theme>) {
 export const gaps: Rule[] = [
   [/^(?:flex-|grid-)?gap-?()(.+)$/, handleGap, { autocomplete: ['gap-$spacing', 'gap-<num>'] }],
   [/^(?:flex-|grid-)?gap-([xy])-?(.+)$/, handleGap, { autocomplete: ['gap-(x|y)-$spacing', 'gap-(x|y)-<num>'] }],
```
chu121su12, line 24: I think the use of x/y should be marked as deprecated

yheuhtozr: I don't have any opinion. There has been no such discussion yet, and either way is fine, provided that you can tell me how to handle deprecation...

chu121su12: I am thinking using `warnOnce` with adjusted message

https://github.com/unocss/unocss/blob/0d463d0cc7e78eaf28c1121c94bb8adaa8b72800/packages/preset-mini/src/_rules/container.ts#L5-L6

antfu: If it's still used in Tailwind, I guess we can't just deprecate it. For ppl that understand the behaviour of `gap-x|y`, I guess it's fine to keep using them (I don't think ppl would have enough incentive to migrate existing projects as well)

yheuhtozr: @chu121su12 So, is it okay to go as it is?

chu121su12: Yes. LGTM

antfu (approved):

PR #3425, opened 2023-12-06
feat(preset-mini): support `size-*` shorthand

Mini-ghost:
This PR adds the `size-*` shorthand to support setting both `width` and `height` in a class.

Example:

```css
/* <div class="size-[calc(1000px-4rem)]></div>  */
.size-\[calc\(1000px-4rem\)\] {
  width: calc(1000px - 4rem);
  height: calc(1000px - 4rem);
}

/* <div class="size-[var(--something)]></div> */
.size-\[var\(--something\)\] {
  width: var(--something);
  height: var(--something);
}

/* <div class="size-10></div>  */
.size-10 {
  width: 2.5rem;
  height: 2.5rem;
}

/* <div class="size-max-10></div>  */
.size-max-10 {
  max-width: 2.5rem;
  max-height: 2.5rem;
}

/* <div class="size-min-10></div>  */
.size-min-10 {
  min-width: 2.5rem;
  min-height: 2.5rem;
}
```


chu121su12 (commented):
How about moving the rule before the `size-w/h` so `size-w-10` can override `size-10`?

Mini-ghost:
Thank you for the reminder, this is indeed very important.

antfu (approved):

zyyv (approved):

antfu (approved):

cywhale:
Sorry, I'm confused about this shorthand rule. After upgrade to UnoCSS >0.58.1 version, my old code with DOM:
```
<input type="text" size="8" />
```
would have a new CSS class:
```
[size~="\38 "] {
    width: 2rem;
    height: 2rem;
}
```
and make my old app looks quite different. Is there anyway to prevent/disable this rule, maybe in unocss.config.ts?

PR #3432, opened 2023-12-11
fix(autocomplete): fixed suggestions to be sorted

### LINKED ISSUE #3412: VSCode Extension - Sort IntelliSense Variants

Correct sorting of autocomplete options, taking into account numeric and string components in substrings.

At the moment, apparently, the options are sorted by the built-in `sort` method without taking into account various factors (see the attached image).

This spoils the development experience and generally looks like a bug.

image

### Suggested solution

I have implemented a base class to properly sort the autocompletion options, taking into account the string and numeric components in the substrings. This class guarantees a UX-correct list of options (see the output list below).

- - - - - - -

<details>
<summary><strong>Class</strong></summary>

```ts
// Methods:
class SmartSort {
    // Constants:
    private readonly _SUBSTRING_EXTRACTOR = /[\d\.]+|[^\d\.]+/g;

    // Variables:
    private _OPTIONS            : string[]                         = [];
    private _OPTIONS_CACHE      : Record<string, RegExpMatchArray> = {};
    private _LOAD_TO_TYPE_CACHE : Record<string, string | number>  = {};

    /**
     * Sorting based on numeric and string components in substrings.
     * @param options - Array of options.
     */
    constructor (options?: string[]) {
        if (options) this._OPTIONS = options;
    }

    /**
     * Convert a string to a number if possible. The method uses a cache.
     * @param value - The original string.
     */
    private _leadToType (value: string): string | number {
        if (this._LOAD_TO_TYPE_CACHE[value]) {
            return this._LOAD_TO_TYPE_CACHE[value];
        } else {
            this._LOAD_TO_TYPE_CACHE[value] = isNaN(+value) ? value : +value;
            return this._LOAD_TO_TYPE_CACHE[value];
        }
    }

    /**
     * Extract string and numeric substrings from a string. The method uses a cache.
     * @param option - The original string.
     */
    private _extractSubstring (option: string): RegExpMatchArray {
        if (this._OPTIONS_CACHE[option]) {
            return this._OPTIONS_CACHE[option]!;
        } else {
            this._OPTIONS_CACHE[option] = option.match(this._SUBSTRING_EXTRACTOR)!;
            return this._OPTIONS_CACHE[option];
        }
    }

    /**
     * Sorting algorithm based on numeric and string components in substrings.
     */
    private _sortAlgorithm (a: RegExpMatchArray, b: RegExpMatchArray): number {
        const totalLengthA = a.reduce((acc, val) => acc + val.length, 0);
        const totalLengthB = b.reduce((acc, val) => acc + val.length, 0);
    
        for (let i = 0; i < a.length; i++) {
            const firstEl  = this._leadToType(a[i]);
            const secondEl = this._leadToType(b[i]);
    
            if (firstEl === secondEl) {
                continue;
            }
            if (typeof firstEl !== typeof secondEl) {
                if (totalLengthA !== totalLengthB) {
                    return totalLengthA - totalLengthB;
                } else {
                    return typeof firstEl === 'number' ? 1 : -1;
                }
            }
            if (typeof firstEl === 'number' && typeof secondEl === 'number') {
                return firstEl - secondEl;
            }
            if (typeof firstEl === 'string' && typeof secondEl === 'string') {
                return firstEl.localeCompare(secondEl);
            }
        }
        return 0;
    }

    /**
     * Set new parameters.
     */
    public set (options: string[]): void {
        this._OPTIONS = options;
    }

    /**
     * Sort the installed options.
     */
    public sort (): string[] {
        return this._OPTIONS
            .map((el) => this._extractSubstring(el))
            .sort((a, b) => this._sortAlgorithm(a!, b!))
            .map((el) => el!.join(''));
    }
}
```
</details>

<details>
<summary><strong>Output</strong></summary>

```ts
// Test:
const STEP    = 4;
const QUOTA   = 8;
const COUNT   = 8;
const NUMBERS = Array(COUNT).fill(null).map((_el, id) => (id < QUOTA) ? (id + 1) : QUOTA + STEP * ((id + 1) - QUOTA));
const OPTIONS = [
   ...NUMBERS.map((el) => `size-${el}`),
   ...NUMBERS.map((el) => `size-1-5-x-${el / 2}`),
   ...NUMBERS.map((el) => `size-1-3-a-${el / 2}`),
   ...NUMBERS.map((el) => `size-1-${el / 2}`),
];

// Variables:
const sortedOptions = new SmartSort(OPTIONS).sort();

// Output with sort method:
console.log(sortedOptions)
```
```shell
# READ ME FROM LEFT TO RIGHT, FROM TOP TO BOTTOM
[ 
   'size-1',         'size-1-0.5',
   'size-1-1',       'size-1-1.5',
   'size-1-2',       'size-1-2.5',
   'size-1-3-a-0.5', 'size-1-3-a-1',
   'size-1-3-a-1.5', 'size-1-3-a-2',
   'size-1-3-a-2.5', 'size-1-3-a-3',
   'size-1-3-a-3.5', 'size-1-3-a-4',
   'size-1-3',       'size-1-3.5',
   'size-1-4',       'size-1-5-x-0.5',
   'size-1-5-x-1',   'size-1-5-x-1.5',
   'size-1-5-x-2',   'size-1-5-x-2.5',
   'size-1-5-x-3',   'size-1-5-x-3.5',
   'size-1-5-x-4',   'size-2',
   'size-3',         'size-4',
   'size-5',         'size-6',
   'size-7',         'size-8'
]
```
</details>

### Alternative

_No response_

### Additional context

I don't have enough time to figure out the code on my own and provide a ready-made solution, and I have no experience in building extensions of this scale for VSCode. This code also needs to be tested more thoroughly. I tried to make this code optimized and fast, however, the use of caches may be redundant in the final implementation.

### Validations

s4nju:
This PR fixes #3412


antfu:
We already doing localeCompare here:
https://github.com/unocss/unocss/blob/147044776be55ce1cb3105fd37dd0197315246d1/packages/autocomplete/src/create.ts#L191-L204

I guess this evolved with a bit more complex sorting algorithm.

s4nju:
I figured that for values like size-1, size-2 etc... both `\d` and `\D` were returning true, which should not be the case.. Hence added `^`.

Assumption: We are trying to give numerical values precedence over the string values. @antfu let me know if any further changes are required or if am missing any edge case. will be happy to fix it. 😄

antfu:
Nice found! I just updated the snap, I don't think the new results are preferable, can you check again and see how it could get improved?

antfu:
Supposely #3620 should improve the situation, if not feel free to bring it back.

PR #3480, opened 2024-01-01
feat: introduce legacy-compat preset

antfu:
Trying to resolve #3478, resolve #3221

TODO:

- [ ] Do we have more options to add?
- [x] Docs page

### DIFF
--- a/packages/preset-legacy-compat/src/index.ts
+++ b/packages/preset-legacy-compat/src/index.ts
@@ -1,0 +1,37 @@
+import { definePreset } from '@unocss/core'
+import { toCommaStyleColorFunction } from './comma-color'
+
+export interface LegacyCompatOptions {
+  /**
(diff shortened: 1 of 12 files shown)


packages/preset-legacy-compat/src/index.ts
```diff
@@ -0,0 +1,37 @@
...
+        if (commaStyleColorFunction)
+          value = toCommaStyleColorFunction(value)
+        if (value !== i[1])
+          i[1] = value
+      })
+    },
+  }
+})
+
+export default presetRemToPx
```
Simon-He95, line 37: This line of code doesn't seem like it should be here

Simon-He95:
image
Should we also add it here?

Simon-He95:
At present, I can only think of this option. If there is any in the future, we can do PR to supplement it.

Issue #3472, opened 2023-12-29
Nuxt 3.9.0 Internal server error: Soft-invalidated module "entry.js" should not have existing transform result


serkodev:
### UnoCSS version

0.58.2

### Describe the bug

After running nuxt dev server, go to browser and refresh the page once, the error will show up.
```
Internal server error: Soft-invalidated module "/node_modules/nuxt/dist/app/entry.js" should not have existing transform result
```

### Reproduction

1. create a new nuxt project
2. add `@unocss/nuxt` module
3. run `nuxt dev`
4. go to browser, no error on first load
5. (Require this step) refresh the page once and the error shows up

https://stackblitz.com/edit/github-xtxzaj?file=package.json

### Additional info
Already try to add the following config regarding to https://github.com/unocss/unocss/issues/3468#issuecomment-1871049463 but still having error
```ts
export default defineConfig({
  features: {
    inlineStyles: false,
  },
})
```

### System Info

```
"nuxt": "^3.9.0",
"vue": "^3.4.0",
"vue-router": "^4.2.5"
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

serkodev:
update: after testing with the latest unocss (0.58.3), this issue is still not solved

kissu:
I do indeed have the same issue but it kinda disappears after a few refreshes? Not really consistent but not a huge blocker either. 🤔

cloydlau:
+1, I have temporarily downgraded nuxt to version 3.8.2.

gigantino:
+1, having the same issue.

jandobrx:
Also having this issue 

> I do indeed have the same issue but it kinda disappears after a few refreshes? Not really consistent but not a huge blocker either. 🤔

But this does seem to do the trick as-well no issues with my build and deployments! But big red internal server error sends gives me chills 🙃

kissu:
[...]

Issue #3468, opened 2023-12-28
Nuxt 3.9.0 Build Error: [unocss] does not found CSS placeholder in the generated chunks


silverwolf42:
### UnoCSS version

0.58.2

### Describe the bug

Integration with the Nuxt(nuxt 3.9.0) module for UnoCSS(0.58.2).

Run `nuxt build` or `nuxt generate` got the error:

```
ERROR  [unocss:global:build:bundle] [unocss] does not found CSS placeholder in the generated chunks
```

### Reproduction

https://stackblitz.com/edit/github-onkxmp

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

designbyfaizi:
Same issue. I'm not able to build since upgrading to Unocss v0.58.2

antfu:
This is caused by #3448 + Nuxt 3.9 changes. 

An workaround for now is to add this to your nuxt config:

```ts
export default defineConfig({
  features: {
    inlineStyles: false,
  },
})
```

Need to investigate deeper to find a better fix.

/cc @zyyv

Nuzzlet:
I'm experiencing this error on Nuxt 3.8.2 when I update @unocss/nuxt beyond version 0.58.0

QaahirStewart:
Same issue here on unocss 0.58.2

PR #3482, opened 2024-01-01
fix(nuxt): temporary revert #3448

zyyv:
Reverts unocss/unocss#3448

Sorry, cause my lack of some tests, this error appeared.

I will consider another way to implement `injectposition`.

Related #3472 #3468

### DIFF
--- a/packages/nuxt/src/types.ts
+++ b/packages/nuxt/src/types.ts
@@ -31,6 +31,14 @@ export interface UnocssNuxtOptions extends UserConfig {
   preflight?: boolean
 
   /**
+   * Adjust the position of the `uno.css` injection. (Depends on `mode`)
+   *
+   * @default 'first'
+   * @deprecated Temporarily removed, will be added back in the future.
+   */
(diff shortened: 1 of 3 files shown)


packages/nuxt/src/types.ts
```diff
@@ -30,13 +30,6 @@ export interface UnocssNuxtOptions extends UserConfig {
    */
   preflight?: boolean
 
-  /**
-   * Adjust the position of the `uno.css` injection. (Depends on `mode`)
-   *
-   * @default 'first'
-   */
-  injectPosition?: 'first' | 'last' | number | { after?: string }
```
antfu, line 38: Removing this will be a breaking change, we should keep the type and do a console warning to say "We temporarily reverted the support for `injectPosition` while seeking for better solution. It's not effective at this moment."

antfu (approved):

NamesMT:
Would love to see an update on this, this have been an occasional blocker headache.

Issue #3474, opened 2023-12-30
Add `important` option to `preset-wind` as it's supported in tailwind


xsjcTony:
### Clear and concise description of the problem

Support tailwind's `important` config.

It's going to help resolve my issue when using with ant design: https://github.com/ant-design/ant-design/issues/46707

### Suggested solution

https://tailwindcss.com/docs/configuration#important

More specifically, this is what I needed: https://tailwindcss.com/docs/configuration#selector-strategy

I want to be able to transform those atomic classes with an selector prefix, e.g. `.p-4` => `#root .p-4`

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

zyyv:
I implemented this in [unocss-preset-useful](https://github.com/unpreset/unocss-preset-useful/blob/60df1b3f0c59475637dc6b4b10bbaf69c995edbd/packages/core/src/core/postprocess.ts#L23-L30).

Check the test case. https://github.com/unpreset/unocss-preset-useful/blob/60df1b3f0c59475637dc6b4b10bbaf69c995edbd/test/postprocess.test.ts#L39-L61

zyyv:
We provide `postprocess` hook, I guess you can handle it yourself?

xsjcTony:
I want to add `#root ` selector before the atomic classname. Let me try whether I can figure it out using variant.

antfu:
I am ok having it to align with tailwind, PR welcome

xsjcTony:
@antfu Wondering whether you want it in `preset-wind` or `preset-mini`?

Also, beyond the tailwind behavior (which I will introduce an option called `important: boolean | string`) which will apply globally, I also want to introduce something like `ancestor-xxx:yyy` variant to partially opt-in.

E.g. `ancestor-#root:m-4` will be `#root .m-4 { ... }`, `ancestor-[#root_>_#container_>]:p-2` will be `#root > #container > .p-2`

Any thoughts on this? If you approve, would you like me to do it in the same PR or separately?

PR #3484, opened 2024-01-02
feat(preset-wind): add `important` option

xsjcTony:
This is the implementation for https://github.com/unocss/unocss/issues/3474

Regarding the usage of `:is()` selector, please read https://github.com/tailwindlabs/tailwindcss/pull/10835

### DIFF
--- a/packages/preset-wind/src/index.ts
+++ b/packages/preset-wind/src/index.ts
@@ -6,15 +6,29 @@ import { rules } from './rules'
 import { shortcuts } from './shortcuts'
 import { theme } from './theme'
 import { variants } from './variants'
+import { postprocessors } from './postprocessors/default'
 
 export { colors, preflights } from '@unocss/preset-mini'
 export type { Theme } from '@unocss/preset-mini'
 
 export { rules, shortcuts, theme, variants }
 
-export interface PresetWindOptions extends PresetMiniOptions { }
+export interface PresetWindOptions extends PresetMiniOptions {
+  /**
+   * The important option lets you control whether UnoCSS’s utilities should be marked with `!important`.
+   *
+   * This can be really useful when using UnoCSS with existing CSS that has high specificity selectors.
+   *
+   * You can also set `important` to a selector like `#app` instead, which will generate `#app .m-1 { ... }`
+   *
+   * @default false
+   */
+  important?: boolean | string
+}
 
 export const presetWind = definePreset((options: PresetWindOptions = {}) => {
+  options.important = options.important ?? false
+
   return {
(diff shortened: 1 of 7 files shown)


packages/preset-wind/src/index.ts
```diff
@@ -6,22 +6,37 @@ import { rules } from './rules'
...
+  options.important = options.important ?? false
+
   return {
     ...presetMini(options),
     name: '@unocss/preset-wind',
     theme,
     rules,
     shortcuts,
     variants: variants(options),
+    postprocess: postprocessors(options),
```
antfu, line 39: I feel it would be better to update the variant directly instead of relying on postprocess, which will affect all `!important` even if it's the user's explicit rule that does not come from the important variant.

xsjcTony: I don't get your point. if you mean it's better to change `important:xxx` variant, then it's another story. The purpose of this PR is as I described in https://github.com/unocss/unocss/issues/3474, which essentially applies to all rules (and that's what Tailwind's `important` global option does too). It will help users fight against those UI library's styles with higher specificity, globally.

Overall, I don't think changing the `important:xxx` variant is a good idea since essentially user will sometimes need to write the real `!important`. Changing its behavior makes no sense.

For your partially opt-in idea, I have a better (at least from my perspective) solution described in https://github.com/unocss/unocss/issues/3474#issuecomment-1873334354, please welcome to have a look and leave your comment.

antfu: Ok, I see. I thought it was the configuration for the important variant, but actually not. Making sense then, I think we should mention that this would affect all rules.

xsjcTony: It's tailwind's functionality so I think it makes sense to include it in preset-wind. But it's up to you anyway.

If you want to proceed with that, I'll add a warning block to the doc. Otherwise you can close this and I don't mind to put it into my own preset for my own usages.

By the way what do you think about my `ancestor-xxx:` variant idea https://github.com/unocss/unocss/issues/3474#issuecomment-1873334354? it could be a potential solution for this.

xsjcTony: I've updated the doc

packages/preset-wind/src/index.ts
```diff
@@ -6,22 +6,37 @@ import { rules } from './rules'
...
```
Jungzl, line 39:
```suggestion
    postprocess: [
      ...toArray(presetMini(options).postprocess),
      ...postprocessors(options),
    ],,
```

PR #3488, opened 2024-01-03
docs: clarify requirements for svelte-scoped hook

mass8326:


jacob-8:
@mass8326 thanks for this addition! I added a bit about the reason why the transform must remain in the server hooks file and removed your sequence example code as SvelteKit docs make it pretty clear how to use sequence. The important thing from our end is just for users to know to keep it the hooks file.

@antfu this is good to merge.

jacob-8 (approved):

PR #3500, opened 2024-01-09
fix: certain inputs causing regex matching to hang

lewisl9029:
Hi there! I recently discovered [an input file](https://github.com/lewisl9029/unocss/blob/d37c170120ae3f1f95254416ee36c19f1874e562/test/assets/regex-dos.ts) that causes UnoCSS to completely hang with 100% CPU usage, and after some investigation it seems to be due to [this regex](https://github.com/unocss/unocss/blob/901db3459b3eff16e230a820549038f28518a722/packages/extractor-arbitrary-variants/src/index.ts#L5C14-L5C37).

You can reproduce the issue by pasting the contents of that input file into https://unocss.dev/play/. The entire tab will freeze.

I went and updated the regexes here to replace all the unbounded `+`s and `*`s with bounded `{1,64}`s and `{0,64}`s respectively, and added a test with the problematic input that now completes successfully in under 1s on my machine (previously the test will hang at 100% CPU as well).

I chose 64 as the upper bound for all of these expressions for now, since it seems like a safe enough upper limit for the kinds of tokens UnoCSS looks for. Hoping tests will verify whether we need to increase it. Though we could probably drastically improve the worst-case performance of the regex by restricting specific pieces to lower limits. Will leave that to someone who understands the regex better than I do. 🙏


antfu:
Thank you!

PR #3506, opened 2024-01-11
feat(eslint-plugin): add new rule: `enforce-class-compile` (#3505)

### LINKED ISSUE #3505: Add rule: `enforce-class-compile` to eslint-plugin

It would be great to have some automated solution when developers want to use a [compile class transformer](https://unocss.dev/transformers/compile-class) across the whole project

### Suggested solution

The possible solution is to add a new ESLint rule, which can handle it

Example:

```diff
<div
-  class="flex p-2"
+  class=":uno: flex p-2"
-  :class="'flex p-2'"
+  :class="':uno: flex p-2'"
-  :class="{flex}"
+  :class="{':uno: flex': flex}"
></div>
```

### Alternative

Alternative 1: First I tried to omit an `:uno:` prefix by using custom regexp ```/(?<quot>["'`])([^$\n"'`]+?)(?:\k<quot>)/g``` but it can brake the app when attribute value is the same as relevant uno class. The value will be replaced with `uno-<hash>`, which is not desired

Example:
```html
<input type="hidden">
```
will be replaced with
```html
<input type="uno-<hash>">
```

---

Alternative 2: create a Vite plugin that will statically analyze the code and add an `:uno:` prefix to all class attributes/directives.

Disadvantages:
- can take a lot of time to implement
- it enforces class compilation without the ability to partially apply compilation and thence developers have no leverage over it

### Additional context

_No response_

### Validations

levchak0910:
This ESLint rule can help with the code base migration to [transformer-compile-class](https://unocss.dev/transformers/compile-class), closes #3505 

Currently only Vue files supported, since I didn't work with Svelte and JSX hence I may not know all usage scenarios. If this PR accepted, I will file two new issues with ask to help implement support for those files too.


antfu:
Can we have some tests to it? Thanks

levchak0910:
Yes, I can add.
I didn't add them initially because I didn't find a prepared lib/code for it. I will add mocha and tests for this rule later.
Thanks )

packages/eslint-plugin/package.json
```diff
@@ -36,7 +36,8 @@
   "scripts": {
     "build": "unbuild",
     "typecheck": "tsc --noEmit",
-    "lint": "nr build && cd ./fixtures && eslint ./src"
+    "lint": "nr build && cd ./fixtures && eslint ./src",
+    "test": "mocha --loader=tsx \"tests/**/*.ts\" --reporter dot --timeout 60000"
```
antfu, line 40: Why mocha instead of Vitest? 👀

levchak0910: I have never seen vitest as a test framework for eslint packages yet
I will explore how to configure it

antfu: Here is an example: https://github.com/eslint-stylistic/eslint-stylistic/blob/main/vitest.config.unit.ts#L8

Basically, you only need to turn globals on.

levchak0910: ready

vitest.config.ts
```diff
@@ -13,5 +13,6 @@ export default defineConfig({
     name: 'unit',
     setupFiles: ['./test/setup.ts'],
     exclude: [...defaultExclude, '**/svelte-scoped/**'],
+    globals: true,
```
levchak0910, line 16: @antfu is it fine? Or do you know any better option how to run eslint tests?
Without this option no tests are found in the eslint-plugin package

packages/eslint-plugin/package.json
levchak0910: thanks for advise )

jacob-8:
> Currently only Vue files supported, since I didn't work with Svelte and JSX hence I may not know all usage scenarios. If this PR accepted, I will file two new issues with ask to help implement support for those files too.

I would recommend not opening an issue for help in doing this with Svelte unless a Svelte user specifically does it themselves. The reason is that if someone is looking to save on size with compilation mode then they're probably already using the `svelte-scoped` mode which automatically compiles all UnoCSS.

antfu (approved):

Issue #3533, opened 2024-01-26
[Feature Proposal] Allow 'grid-' part to be optional when specifying grid columns


AndrewBogdanovTSS:
### Clear and concise description of the problem

As a developer using UnoCSS I want to be able to use shorter utilities when spcifying grid column sizes

### Suggested solution

https://github.com/search?type=code&q=repo:unocss/unocss%20%22%2F%5Egrid-(rows%7Ccols)-(%5Cd%2B)%24%2F%22
```
[/^?:grid-(rows|cols)-(\d+)$/, ([, c, d]) => ({ [`grid-template-${rowCol(c)}`]: `repeat(${d},minmax(0,1fr))` }), { autocomplete: ['grid-(rows|cols)-<num>', 'grid-(rows|cols)-none'] }],
```

`grid-` in the above regular expression should be optional pretty much the same as it's already done for grid auto flow utility
https://github.com/unocss/unocss/blob/8505f48d4ccba055c0613bd22f120312447682ca/packages/preset-mini/src/_rules/grid.ts#L54

### Alternative

Right now I just have to manually keep those rules in the project
```
import type { Rule } from 'unocss'
import { h } from '@unocss/preset-mini/utils'

function rowCol(s: string) {
  return s.replace('col', 'column')
}
function rowColTheme(s: string) {
  return s[0] === 'r' ? 'Row' : 'Column'
}

export const grids: Rule[] = [
  [/^(rows|cols)-(.+)$/, ([, c, v], { theme }) => ({
    [`grid-template-${rowCol(c)}`]: theme[`gridTemplate${rowColTheme(c)}`]?.[v] ?? h.bracket.cssvar(v)
  })],
  [/^(rows|cols)-minmax-([\w.-]+)$/, ([, c, d]) => ({
    [`grid-template-${rowCol(c)}`]: `repeat(auto-fill,minmax(${d},1fr))`
  })],
  [/^(rows|cols)-(\d+)$/, ([, c, d]) => ({
    [`grid-template-${rowCol(c)}`]: `repeat(${d},minmax(0,1fr))`
  }), { autocomplete: ['(rows|cols)-<num>', '(rows|cols)-none'] }]
]
```
but it would be great if that would come out of the box, given that the change is just about adding `?:` to the regexp

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
[...]

PR #3542, opened 2024-01-29
feat(preset-mini): allow shorter syntax for grid-cols utilities (#3533)

AndrewBogdanovTSS:
Closes https://github.com/unocss/unocss/issues/3533
Makes grid- prefix optional when specifying grid-cols-[val] utility.
Before: grid-cols-2
After: cols-2


antfu:
Could you help adding a few cases to the tests to guard that the usage won't be changed accidentally in the future? Thanks

antfu (approved):

PR #3559, opened 2024-02-08
docs(cli): add info for cli with pnpm (#3545)

### LINKED ISSUE #3545: 'unocss' is not recognized as an internal or external command, operable program or batch file
### UnoCSS version

latest

### Describe the bug

Following the guide in https://unocss.dev/integrations/cli, but turns out it's not working.
Snipaste_2024-02-01_22-34-09

### Reproduction

This is a minimal repo that's created by `create-next-app` with `uno.config.ts` setup. Currently at the stage to generate `uno.css` by CLI but failed to do so.
https://github.com/xsjcTony/PokeTool

### System Info

Windows 11, Chrome 121, Next.js 14

Snipaste_2024-02-01_22-40-50

### Validations

xsjcTony:
docs(cli): add info for cli with pnpm, closes https://github.com/unocss/unocss/issues/3545

### DIFF
--- a/docs/integrations/cli.md
+++ b/docs/integrations/cli.md
@@ -41,6 +41,10 @@ You can also install the standalone package:
   ```
 :::
 
+::: info
+If you are not able to find the binary (e.g. with `pnpm` and only `unocss` is installed), you'll need to explicit install `@unocss/cli` standalone package.
+:::
+
 ## Usage
 
 You can also pass multiple glob patterns to `@unocss/cli`:


docs/integrations/cli.md
```diff
@@ -41,6 +41,10 @@ You can also install the standalone package:
   ```
 :::
 
+::: info
+If you are not able to find the binary (e.g. with `pnpm` and only `unocss` is installed), you'll need to explicit install `@unocss/cli` standalone package.
+:::
```
antfu, line 46: We could probably also remove L16-28

xsjcTony: If I remove L16-28, then the tip I added makes no sense, since we are letting user explicitly install it anyway. Which approach do you prefer?

- Remove L16-28, remove what I've added, and Remove `you can also install the standalone package:`
Or
- Keep it as is.

antfu: Let's remove L16-28 then. It won't work anyway

xsjcTony: No, it works with `npm`, or `yarn`, and I believe (guess) with `pnpm`'s `--shamefully-hoist` being enabled.

PR #3571, opened 2024-02-22
Fix playground link in presets/wind.md

yooneskh:
It seems like the plaayground link was changed but not updated here.


antfu:
No, it's a different thing! https://unocss.dev/interactive/

yooneskh:
It is strange because it was giving me 404 error. Sorry and thank you.

Issue #3574, opened 2024-02-22
@apply Directives Not Working With Custom Variant


mr-thomas-z:
### UnoCSS version

0.58.5

### Describe the bug

I followed the docs to wrap custom variant with quotes (e.g., `"hover:bg-white"`, but the output CSS was not converted. Instead, here are the before and after code:
|Original|Converted|
|:-|:-|
|`@apply flex flex-row p-4 text-white "hover:color-red-400";`|`display:flex;flex-direction:row;padding:1rem;--un-text-opacity:1;color:rgb(255 255 255 / var(--un-text-opacity));`<br><br>(custom variant is gone?)|
|`@apply "hover:bg-white" "hover:bg-opacity-5";`|`@apply "hover:bg-white" "hover:bg-opacity-5";`<br><br>(custom variant is not converted?)|

The only way I would get it working is to use a `--at-apply:` for each custom variant class. I'm not sure if this is the intended behavior or a bug.
|Not working|Worked|
|:-|:-|
|`@apply "hover:bg-white" "hover:bg-opacity-5";`|`--at-apply: 'hover:bg-white';`<br>`--at-apply: 'hover:bg-opacity-5';`|

### Reproduction

[...]

PR #3578, opened 2024-02-25
docs(directives): avoid requiring quotes in `--at-apply` causes misun…

zhangzheheng12345:
In issue #3574 a user has used `@apply 'hover:bg-red'`  (for exmaple) in which the quotes are actually not required. ONLY `--at-appy` needs them. The doc should be fixed to avoid this misunderstanding.

### DIFF
--- a/docs/transformers/directives.md
+++ b/docs/transformers/directives.md
@@ -41,7 +41,7 @@ export default defineConfig({
 
 ```css
 .custom-div {
-  @apply text-center my-0 font-medium;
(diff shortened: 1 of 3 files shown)


Simon-He95 (approved):

zhangzheheng12345:
**This change should be reconsidered.**

Further feedback from issue #3574 shows that formatter of the editor will automically format `@apply hover:bg-red` to `@apply hover: bg-red` (tested on VS Code, builtin formatter). So I think we should SUPPORT quotes when using `@apply`.

PR #3632, opened 2024-03-18
fix(vscode): completion color

Simon-He95:
At present, vscode does not seem to support the use of rgb (r g b / a) form. I think it can be temporarily converted to rgba form and give color tips. If vscode supports it later, it will be removed.

image

image

image

### DIFF
--- a/packages/vscode/src/utils.ts
+++ b/packages/vscode/src/utils.ts
@@ -145,7 +145,10 @@ export function getColorString(str: string) {
   // if (!(new TinyColor(colorString).isValid))
   //   return
 
-  return colorString
+  if (/\/\)/.test(colorString))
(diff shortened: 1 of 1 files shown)


zyyv:
Hi, is this a new feature?
Please test this, if it was a shortcut that contained multiple colors, what would it show?

Simon-He95:
> Hi, is this a new feature?你好，这是新功能吗？ Please test this, if it was a shortcut that contained multiple colors, what would it show? 请测试一下，如果是包含多种颜色的快捷方式，会显示什么？

Because the form that was previously synchronized to tailwind was changed to rgb(), but we have transparency, vscode does not seem to recognize the form of rgb (r g b / a). For the sake of compatibility, it is processed into rgba when prompted by the vscode plug-in.

antfu (approved):

Issue #3663, opened 2024-03-28
fontSize theme configuration, css not compiled correctly.


RayGuo-ergou:
### UnoCSS version

0.58.7

### Describe the bug

```ts
export default defineConfig({
  theme: {
    fontSize: {
      // '2xl': ['1.5rem', {
      //   'line-height': '2rem',
      //   'letter-spacing': '-0.01em',
      //   'font-weight': '500',
      // }],
      '2xl': ['1.5rem', {
        lineHeight: '2rem',
        letterSpacing: '-0.01em',
        fontWeight: '500',
      }],
    },
  },
});
```

Pass in this config, the css will build like this:

```css
.text-2xl{font-size:1.5rem;lineHeight:2rem;letterSpacing:-.01em;fontWeight:500}
```

which does not recognize by any browser.
![image](https://github.com/unocss/unocss/assets/33137074/63529743-3d4d-4d13-9c56-2801d180a7df)
 

### Reproduction

https://stackblitz.com/edit/vitejs-vite-2mouvf

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Simon-He95:
https://unocss.dev/config/rules#static-rules

RayGuo-ergou:
Why this is closed? They are not even the same thing. ( theme vs rules )

Or does this mean it will never been supported?

Simon-He95:
The document gives the method of use - splicing. You should follow the document when using it.

RayGuo-ergou:
I understand that there's a work around of this issue ( I even have that in the issue report ).

The main purpose of this issue is because camel case `lineHeight` etc if from tailwind's configuration ( see: [here](https://tailwindcss.com/docs/font-size#providing-a-default-line-height)).

As also in unocss documentation:
> UnoCSS also supports the theming system that you might be familiar with in Tailwind CSS / Windi CSS. At the user level, you can specify the theme property in your config, and it will be deep-merged to the default theme.

[...]

PR #3666, opened 2024-03-28
fix(theme): Compatible with tailwind’s theme configuration

Simon-He95:
close: #3663


antfu:
Emm, no, I made that decision on purpose. I don't think we should allow ppl to use camelCase especially because it's not that hard to write kebab-case that aligns with how CSS would work. TBH it's more like a "bad habit" the community has inherited from CSS-in-JS. Uno has no plan to support that for both performance as well as convention-free.

We already note it here: https://unocss.dev/config/rules#static-rules

Thanks for the PR.

PR #3670, opened 2024-03-31
feat(core): optimize core code base

eryue0220:
This pr is just to upgrade `unocss/core` codebase.


antfu:
Thank you, but no, I don't consider this as an "optimize".

Issue #3675, opened 2024-04-01
wrong types/subpackage exports in 0.59.0-beta.1


userquin:
### UnoCSS version

0.59.0-beta.1

### Describe the bug

Issue for tracking packages that should be fixed in #3380:
- ✔️ `@unocss/eslint-config`: missing node10 types in `@unocss/eslint-config/flat` subpackage export (missing types versions (?) ): (https://github.com/unocss/unocss/pull/3677)
- 👍  `@unocss/nuxt`: node16-cjs being masquerading (only review) ( **this package is fine** )
- ✔️ `@unocss/preset-icons`: missing node10 types in `@unocss/preset-icons/browser` and `@unocss/preset-icons/core` subpackage exports (missing types versions (?) ) (https://github.com/unocss/unocss/pull/3677)
- ✔️ `@unocss/preset-uno`: missing node10 types in `@unocss/preset-uno/theme`, `@unocss/preset-uno/colors` and `@unocss/preset-uno/utils` subpackage exports (missing types versions (?) ) (https://github.com/unocss/unocss/pull/3677)
- ✔️ `@unocss/scope`: no types for  for node16-esm and bundler (only review) (https://github.com/unocss/unocss/pull/3676)
- ✔️ `@unocss/vite`: missing node10 types in `@unocss/vite/client` subpackage export (missing types versions (?) ) (https://github.com/unocss/unocss/pull/3677)
- `@unocss/vscode`: **NOT PUBLISHED IN NPM REGISTRY** (release failed) => we don't need to check this package here
- `@unocss/webpack`: ESM Only? Shouldn't be dual ESM/CJS? or `unocss/webpack` shouldn't be ESM Only?
- ✔️ `@unocss/postcss`: should be ESM first + dual ESM/CJS, missing `@unocss/postcss/esm` subpackage export? https://github.com/unocss/unocss/pull/3380#issuecomment-1870490289 ( https://github.com/unocss/unocss/pull/3678 )

/cc @antfu @kwaa 

### Reproduction

NA

### System Info

NA

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #3676, opened 2024-04-01
fix(scope): fix export types & main entry

kwaa:
Part of https://github.com/unocss/unocss/issues/3675

### DIFF
--- a/packages/scope/package.json
+++ b/packages/scope/package.json
@@ -12,11 +12,17 @@
   },
   "exports": {
     ".": {
-      "import": "./index.mjs",
-      "require": "./index.cjs"
+      "import": {
+        "types": "./index.d.ts",
+        "default": "./index.mjs"
+      },
+      "require": {
+        "types": "./index.d.cts",
+        "default": "./index.cjs"
+      }
     }
   },
-  "main": "./index.mjs",
+  "main": "./index.cjs",
   "module": "./index.mjs",
   "types": "./index.d.ts"
 }
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -24,12 +24,17 @@
   "sideEffects": false,
   "exports": {
     ".": {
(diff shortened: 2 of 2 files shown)


kwaa:
It looks like `@unocss/nuxt` doesn't output `index.d.cts`.

userquin:
the cjs file is on root

kwaa:
> the cjs file is on root

I know, I just don't quite understand if an `./dist/index.d.ts` declaration file should be used in this case.

userquin:
the cjs file is on root: may we should move it to ESM first with dual ESM/CJS (VueUse  using that approach: you will need to patch generated cjs file)

kwaa:
> the cjs file is on root: may we should move it to ESM first with dual ESM/CJS (VueUse using that approach: you will need to patch generated cjs file)

Hmm... Maybe this should be done in a separate PR.

packages/scope/package.json
```diff
@@ -1,22 +1,28 @@
...
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/unocss/unocss",
+    "directory": "packages/scope"
+  },
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./index.d.ts",
```
userquin, line 16: shoud be d.mts

userquin: upps, there is no d.mts file...

kwaa: > from my review this should be ESM first, still missing node16-esm and bundler in your local? can you upload an screenshot?

These two PRs I created via github.com...

userquin: ok, I'll try to review it later on my local

userquin (approved):

packages/scope/package.json
userquin: imagen

Issue #3604, opened 2024-03-08
Custom option: discardComments


green-mike:
### Clear and concise description of the problem

I wanna remove css comment, but cssnano config has been hardcode as 
```
{ mergeRules: false, normalizeWhitespace: false, discardComments: false }
```
https://github.com/unocss/unocss/blob/09514dbb19178476cc1be2c7eae5b5a60a2cecc9/packages/nuxt/src/index.ts#L89-L94
### Suggested solution

Add a option to override config.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #3679, opened 2024-04-02
fix(nuxt): Custom option priority in cssnano

Simon-He95:
close: #3604

### DIFF
--- a/packages/nuxt/src/index.ts
+++ b/packages/nuxt/src/index.ts
@@ -88,8 +88,8 @@ export default defineNuxtModule<UnocssNuxtOptions>({
       const preset = nuxt.options.postcss.plugins.cssnano.preset
       nuxt.options.postcss.plugins.cssnano = {
         preset: [preset?.[0] || 'default', Object.assign(
-          preset?.[1] || {},
           { mergeRules: false, normalizeWhitespace: false, discardComments: false },
+          preset?.[1],
         )],
       }
     }


sibbng (approved):
It's nice to have, but I'm sceptical about whether it solves the issue. These options are only enforced when using vite builder, and esbuild already removes comments unless explicitly disabled. 

cssnano's `discardComments` optimization [was causing issues with empty CSS variables](https://github.com/unocss/unocss/pull/2286) for older browsers (Chrome <= 98), that is why it's disabled. 

There is a chance that the underlying issue in https://github.com/unocss/unocss/issues/3604 is something else.

Simon-He95:
> It's nice to have, but I'm sceptical about whether it solves the issue. These options are only enforced when using vite builder, and esbuild already removes comments unless explicitly disabled.很高兴拥有，但我怀疑它是否能解决问题。这些选项仅在使用 vite builder 时强制执行，并且 esbuild 已删除注释，除非明确禁用。
> 
> cssnano's `discardComments` optimization [was causing issues with empty CSS variables](https://github.com/unocss/unocss/pull/2286) for older browsers (Chrome <= 98), that is why it's disabled.cssnano 的 `discardComments` 优化导致旧版浏览器 (Chrome <= 98) 出现空 CSS 变量问题，这就是它被禁用的原因。
> 
> There is a chance that the underlying issue in #3604 is something else.#3604 中的根本问题可能是其他问题。

Maybe you are right, but according to what he mentioned in this issue, the customized attribute settings are overwritten by the default ones. I think this point can be solved.

Issue #3691, opened 2024-04-05
Add support for peer and group variants on aria-* variants


Johnkat-Mj:
### Clear and concise description of the problem

Currently, Unocss does not support applying utility variants for ARIA attributes (such as aria-expanded) in conjunction with peer or group selectors. This means we cannot style elements based on the state of a related element using ara-* variants within peer or group selectors

### Suggested solution

Extending Unocss to allow using peer and group selectors with ara-* variants. This would enable more granular and dynamic styling based on ARIA attributes. Here's an example:
```
<button data-accordion-trigger aria-label="toggle button"
    class="... group" ...>
    Can it be animated?
    <span
        class=".... group-aria-expanded-rotate-180"></span>
</button>
```

Currently, to accomplish the same effect, I have to write code like this:

```
<button data-accordion-trigger aria-label="toggle button"
    class="... [&[aria-expanded=true]>span]-rotate-180" ...>
    Can it be animated?
    <span data-trigger-icon
        class="ease-linear duration-300 ..."></span>
</button>
```

This would enable flexible and accessible styling for components based on changes in aria attribute states. It would enhance the maintainability and reusability of components by separating styling logic from component logic like in the given example.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #3692, opened 2024-04-06
feat(preset-mini): Add support for peer and group variants on aria-* variants to Preset-Mini

Johnkat-Mj:
close: https://github.com/unocss/unocss/issues/3691

This pull request enable the use of peer and group selectors in conjunction with ara-* variants ( #3691 )
This will provide the ability to achieve styles like:

- peer-aria-expanded:bg-red: Applies a red background to an element when its peer element has the aria-expanded attribute set to true.

- group-aria-expanded-p-4: Adds a padding of 4 units to elements within a group when a group member has the aria-expanded attribute set to true.

Example
```
<button data-accordion-trigger aria-label="toggle button"
    aria-expanded="true"
    class="... group" ...>
    Can it be animated?
    <span
        class=".... group-aria-expanded-rotate-180"></span>
</button>
```
In this example, the `rotate-180`  will be applied to the span element when the group member (button) has the aria-expanded attribute set to true.


zyyv:
@Johnkat-Mj Can you add some test cases?

Johnkat-Mj:
@zyyv Can you guide me through this?
Should I just add a usage example in EXAMPLES or should I write a test for this?

docs/presets/community.md
```diff
@@ -27,6 +27,7 @@
 - [unocss-preset-animations](https://unocss-preset-animations.aelita.me/) - Preset for flexible and composable animations by [@Aelita](https://github.com/xsjcTony).
 - [unocss-preset-shadcn](https://github.com/hyoban/unocss-preset-shadcn#readme) - Use [shadcn/ui](https://ui.shadcn.com) with UnoCSS by [@zouhang](https://github.com/zouhangwithsweet) and [@hyoban](https://github.com/hyoban).
 - [@yeungkc/unocss-preset-safe-area](https://github.com/YeungKC/unocss-preset-safe-area) - Use safe area with Unocss by [YeungKC](https://github.com/YeungKC)
+- [@unifyui/unify-preset](https://github.com/unify-ui-dev/unify-preset/blob/main/README.md) - Atomic Theming UI Libray Powered By UnoCSS, like DaisyUI but customizable. By [@johnkat-mj](https://github.com/Johnkat-Mj)
```
antfu, line 30: Can you do this with a separate PR? One PR should only do one thing if possible

zyyv:
> @zyyv Can you guide me through this? Should I just add a usage example in EXAMPLES or should I write a test for this?

Like this https://github.com/unocss/unocss/blob/5ae78dfb3e4a9c6d5de326de6a9ac6de9b5e87c6/test/assets/preset-mini-targets.ts#L1176-L1177
Add your test cases, update the snapshot and check it right?

Johnkat-Mj:
Hey @zyyv 
The test fails when I add `group-aria-[expanded=true]:bg-green-400` in `test/assets/preset-mini-targets.ts`, but I can't figure out the cause. After running `nr build`, I added an example in `examples/nuxt3`, and it worked fine as expected without any issues. Even the 'output' after running `pnpm build` showed no problems. However, I'm still unable to understand why the test is failing, the logic is the same as that of tagged data attributes

Simon-He95:
close: #3691

antfu (approved):

Issue #3684, opened 2024-04-03
[VSCode extension]: Underline decoration not show for attribute if class use multiple selector before it


JianJroh:
### UnoCSS version

0.58.9

### Describe the bug

For ```mt-1```

**Not show**
![image](https://github.com/unocss/unocss/assets/45714701/ba853478-fcc0-49e5-afb3-72347d0faa11)

<hr />

**Show**

![image](https://github.com/unocss/unocss/assets/45714701/bc8b60ae-422f-452f-8375-e5a4ee15c3d4)

![image](https://github.com/unocss/unocss/assets/45714701/40336df1-9d86-4805-a069-7ca85bb3391c)

### Reproduction

In vscode

### System Info

vscode: 1.87.2

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Simon-He95:
<img width="394" alt="image" src="https://github.com/unocss/unocss/assets/57086651/06edcf25-1614-43e0-bae3-ea7457b13654">

You can add a space to avoid this problem, I will try to fix it later

Simon-He95:
@JianJroh I found that this is not a bug, it is because the `presetAttributify` plug-in is not introduced, and it supports the mode of using attributes.
[...]

PR #3696, opened 2024-04-08
fix(shared-integration): defaultIdeMatchInclude

Simon-He95:
close: #3684


antfu:
Could we have some simple tests to avoid future regressions?

PR #3708, opened 2024-04-11
fix: move `@unocss/webpack` to ESM First

userquin:
From this comment https://github.com/unocss/unocss/issues/3675#issuecomment-2030417133

imagen


antfu:
Wouldn't that break existing usages?

userquin:
that's why it is breaking

antfu (approved):

Issue #3646, opened 2024-03-21
Download functionality to presetWebFonts for offline/local network website


HassanDomeDenea:
### Clear and concise description of the problem

When you add fonts using presetWebFonts, for example:

```
//uno.config.ts
export default defineConfig({
  presets: [
    //   ...
    presetWebFonts({
      fonts:{
        lobster: 'Lobster',
        vibes: 'Vibes',
      }
    })
  ]
})
```

this will create `@font-face` definition for each font, containing the URL from google for example:

```

@font-face {
  font-family: Vibes;
  font-style: normal;
  font-weight: 400;
  font-display: swap;
  src: url(https://fonts.gstatic.com/s/vibes/v14/QdVYSTsmIB6tmYd1FJfcuA.woff2) format("woff2");
  unicode-range: U+0600-06FF, U+0750-077F /* .... */
}
```

For some web application that works offline or inside (local network), we cannot use this as we need the files to those fonts in the assets of our applications to work properly.

### Suggested solution

My suggested solution is inspired from a vite plugin [vite-plugin-webfont-dl](https://github.com/feat-agency/vite-plugin-webfont-dl)
An option (global or per font) inside presetWebFonts config, to allow downloading fonts locally, thus the generated styles use the relative path to font file from our project.

Example for suggested config:
```

presetWebFonts({
  downloadLocally: true, //Global setting, unless overridden 
  fonts: {  
       vibes: 'Vibes',
       lobster: [
           {
                 name: 'Lobster,
                 provider: 'none',
                 downloadLocally: false, // Overridden here
           }
        ],
    ],
  },
})

```

Generated code will be for example:
```
@font-face {
  font-family: Vibes; 
  src: url(assets/font-vibes-QdVYSTsmIB6tmYd1FJfcuA.woff2) format("woff2");
}
```

### Additional context

This vite plugin [https://github.com/feat-agency/vite-plugin-webfont-dl](url) unfortunately does not work with font-faces generated by unocss/presetWebFonts (whether inline import is true or false)
Or did I miss a working configuration or workaround? 

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

antfu:
[...]

PR #3710, opened 2024-04-11
feat(web-fonts): add local font support to web font preset

onmax:
🏗️ WIP

Solves #3646

### DIFF
--- a/packages/preset-web-fonts/src/local-font.ts
+++ b/packages/preset-web-fonts/src/local-font.ts
@@ -1,0 +1,63 @@
+/**
+ * Inspired by:
+ * https://github.com/feat-agency/vite-plugin-webfont-dl/blob/master/src/downloader.ts
+ */
+
+import { mkdir, stat, writeFile } from 'node:fs/promises'
+import { Buffer } from 'node:buffer'
+import { dirname, resolve } from 'node:path'
+import { fileURLToPath } from 'node:url'
+import type { WebFontsOptions } from './types'
+
+const fontUrlRegex = /[-a-z0-9@:%_+.~#?&/=]+\.(?:woff2?|eot|ttf|otf|svg)/gi
+
+const dir = dirname(fileURLToPath(import.meta.url))
+
+const cachedUrls = new Map<string, string>()
+
+export async function useLocalFont(css: string, fetch: WebFontsOptions['customFetch']) {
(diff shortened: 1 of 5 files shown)


packages/preset-web-fonts/src/local-font.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  async function saveFont(url: string, path: string) {
+    const response = await fetch!(url, { headers: { responseType: 'arraybuffer' } }) as ArrayBuffer
+    const content = new Uint8Array(response)
+    await writeFile(path, Buffer.from(content))
+    return { url, path }
+  }
+
+  // FIXME Currently hardcoded to the playground folder
+  const fontsFolder = resolve(dir, '../../../playground/assets/fonts')
```
onmax, line 39: I need to write the new fonts into the assets folder but I am not sure how to get that **relative** path from the 'virtual:uno.css'

Any ideas?

playground/uno.config.ts
onmax, line 1: When this PR is done I will revert this changes. Just for testing!

onmax:
I have no idea how to get the path from the root project in order to write the fonts to the assets folder. 

I don't even know if this is the right solution...

Or should I use a virtual Vite file instead? I am blocked now 😓


Any ideas are much welcomed! Thanks!

userquin:
I guess you will need to add an option to the preset  (where to download the fonts?, by default `${process.cwd()}/public`, sveltekit should be `${process.cwd()}/static`) and also a plugin to serve fonts: you also need to expose that option in the context, and so the plugin can resolve it. Check preset icons, we also have some option in the preset to resolve local collections.

I don't see any protection when running in the browser, you should protect the preset: check [preset-icons](https://github.com/unocss/unocss/blob/main/packages/preset-icons/src/core.ts#L177)

userquin:
I'll try to add initial logic for vite (with small preset changes), later we can add webpack and CLI support (CLI will not require any integration since it is not a bundler, we only need to download logic): we need to add support for dev server to download the fonts

userquin:
@onmax shouldn't the css also be saved locally?

onmax:
> @onmax shouldn't the css also be saved locally?

Why? Don't we want to let UnoCSS handle the CSS for us?

Once you have the fonts downloaded in your machine, the preset will no longer make requests to the font provider but we will always update the CSS `url()` imports to link to the local font file

onmax:
#3723 3723

Issue #3711, opened 2024-04-11
High CPU usage in VS Code extension


AnotiaWang:
Type: <b>Bug</b>

VS Code stopped showing code suggestions, and plugins became unresponsive. A `Code Helper (Plugin)` process consumes 100% CPU, and can't be killed when reloading the window, resulting CPU being drained after several reloads. I used VS Code's Extension Bisect feature and located UnoCSS plugin.

Maybe related: https://github.com/vuejs/language-tools/issues/4239

Extension version: 0.59.1
VS Code version: Code 1.88.0 (5c3e652f63e798a5ac2f31ffd0d863669328dc4c, 2024-04-03T13:28:09.495Z)
OS version: Darwin arm64 23.4.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M2 Pro (10 x 24)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|2, 3, 4|
|Memory (System)|32.00GB (5.00GB free)|
|Process Argv|--crash-reporter-id 33f2db18-d6d2-4571-bcf7-5ce1184f51fc|
|Screen Reader|no|
|VM|0%|
</details>

Saeid-Za:
I have the same issue, vscode is completely frozen and unresponsive for me.
The current workaround is to downgrade to `v0.59.0`.

Simon-He95:
I'll check it later.

Wangijun:
I also encountered VsCode freeze in version 1.88.1 and unocss 0.59.1. The current solution is to downgrade it to unocss 0.59.0,


PS: Operating System Win10

lainbo:
I'm facing the same issue where my VSCode is unable to load any functionality other than typing. Using Start Extension Bisect, I found out the issue is with this extension, so I've now downgraded to v0.59.0 for use

Simon-He95:
I have found the reason and will submit a PR to fix it later.

Saganic:
[...]

PR #3712, opened 2024-04-12
fix(shared-integration): vscode plug-in stuck

Simon-He95:
close: #3711

The original regular expression would cause an error when matching code. Now it has been updated and the ending tag is filtered.

## origin:

image

image

## new:

image


zcf0508:
image

Simon-He95:
> image
You can take a look to see if there are any problems with this regular rule.
`/<[^\/](?:[^<>\[\]"]+)(?:>|"[^"]*"(?:(?:[^<>\[\]{]+)(?:>|{[^}]*}|(?:\[[^\]]*\])))*|(?:\[[^\]]*\](?:(?:[^<>\[\]{"]+)(?:>|"[^"]*"|{[^}]*}|(?:\[[^\]]*\])))*))/g`

packages/shared-integration/src/defaults.ts
```diff
@@ -15,8 +15,8 @@ export const defaultFilesystemGlobs = [
 export const defaultIdeMatchInclude: RegExp[] = [
   // String literals
   /(['"`])[^\1]*?\1/g,
-  // HTML tags ignore [>]
-  /<(?:[^<>\[\]]+|\[[^\]]*\])*>/g,
+  // HTML tags
+  /<[^\/](?:[^<>\[\]"]+)(?:>|"[^"]*"(?:(?:[^<>\[\]{]+)(?:>|{[^}]*}|(?:\[[^\]]*\])))*|(?:\[[^\]]*\](?:(?:[^<>\[\]{"]+)(?:>|"[^"]*"|{[^}]*}|(?:\[[^\]]*\])))*))/g,
```
antfu, line 19: This seems to be over complicated to me

Maybe:

```suggestion
  /<(?:"[^"]*"|'[^"]*'|[^>])+>/g,
```

Simon-He95: image
He will miss some needed matches

Simon-He95:
> image
> 
> image

This is an irregular way of writing. You should not write similar code directly in attributes. In addition, you should not use words such as relative when naming.

PR #3722, opened 2024-04-13
feat: replace fast-glob with fdir

43081j:
This replaces fast-glob with fdir for file-system traversal.

`fdir` has zero dependencies and is often faster than `fast-glob`.

Note that the `postcss` package is still using `fast-glob` as it has a requirement for accessing the `Stat` of each found file (which `fdir` does not expose yet).

If/when fdir exposes the file stats, we can drop `fast-glob` there too.

maintainers: let me know if this makes sense to you. i'm just roaming around the ecosystem doing this in various places as part of the ecosystem cleanup, so if there's some reason you want to keep `fast-glob`, just let me know.


antfu:
Is there any significant problem with `fast-glob`? While I am not against squeezing out more performance is possible, I think the first thing to to compare the trade-offs the ROI of the migration.

1. If `fdir` could have a compatible API as `fast-glob` to serve as a drop-in replacement, that would be a huge plus (but if that's the case, I guess it make sense to help improving `fast-glob`?
2. `fast-glob` is spread out the entire ecosystem, estimate `fast-glob` from UnoCSS does not save space in a realworld project (but actually could add new `fdir` dependencies to overall packages). In an incomplete ecosystem migration, it actually increases the install size. 

I saw you list a few packages with this in https://github.com/43081j/ecosystem-cleanup/issues - so the problem becomes that we can get only the benefits of it when **all packages agree to migrate**. So I main concern is that because the API design and feature scope are quite different between `fast-glob` and `fdir`, I don't think this would be a no-brainer migration. I think there would be a high chance that some packages can't or not willing to migrate to `fdir`. Which might make this end up not getting the value that was originally aiming for.

antfu:
Thanks for the PR! At this moment I think it's not very valuable to do the switch especially the APIs are different. I suggest we wait for a while and use the native `fs.glob` later.

PR #3723, opened 2024-04-14
feat(web-fonts): add downloadLocally options

onmax:
🏗️ WIP

Solves #3646 

# How does it work?

## Using remote font

It works as before. We just moved the logic of fetching the remote css to its own function: [`getRemoteCSS`](https://github.com/unocss/unocss/pull/3723/files#diff-76141617ba437b245325874fd62f4d50847a8f341697a55de101d185e360d4caR5)

## Using local font

1. It will try to read the file if it has been already generated using [`readFontCSS`](https://github.com/unocss/unocss/pull/3723/files#diff-207312299160f8555e9495fafbbf6e11cbf29e4c602befc9571a495178eadd67R56). If it has not been generated a placeholder will be return. The content of that function is the content that it is returned to the `getCSS` of the preflight of the preset.
2. In the [Vite Plugin](https://github.com/unocss/unocss/pull/3723/files#diff-134ebd7597fd226904db2644435204eca57069b8d4d11c8e662260fa8afcdbb5R12-R30) and then mainly use the [`useLocalFont` function](https://github.com/unocss/unocss/pull/3723/files#diff-207312299160f8555e9495fafbbf6e11cbf29e4c602befc9571a495178eadd67R18-R37):
  2.1 Download the remote CSS using [`getRemoteCSS`](https://github.com/unocss/unocss/pull/3723/files#diff-76141617ba437b245325874fd62f4d50847a8f341697a55de101d185e360d4caR5)
  2.2 Extract the urls
  2.3 Download the urls and save them locally
  2.4 Update the CSS file with the local paths
  2.5 Write the CSS file

----

## TODO

- [ ] Verify that this process makes sense
- [ ] Make tests


onmax:
@userquin would you mind checking that the code makes sense?

Is there anything missing other than support for svelte for example? Thank you!

packages/vite/src/web-fonts.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// eslint-disable-next-line node/prefer-global/process
+const isNode = typeof process !== 'undefined' && process.stdout && !process.versions.deno
+
+export function createWebFontPlugins(ctx: UnocssPluginContext): Plugin[] {
+  return [
+    {
+      name: `unocss:web-fonts-local:dev`,
+      enforce: 'pre',
+      apply: 'serve',
```
userquin, line 15: Remove apply, we also need generateBundle if output folder not publicDir not assetsDir: vite won't bundle css and fonts if not using any of those folders

userquin: Let Anthony check this PR, maybe he can provide some hint

packages/preset-web-fonts/src/remote-font.ts
```diff
@@ -0,0 +1,44 @@
...
+  for (const provider of enabledProviders) {
+    const fontsForProvider = fonts.filter(i => i.provider.name === provider.name)
+
+    if (provider.getImportUrl) {
+      const url = provider.getImportUrl(fontsForProvider)
+      if (url)
+        preflights.push(await importUrl(url))
+    }
+
+    preflights.push(provider.getPreflight?.(fontsForProvider))
```
onmax, line 40: This is missing for `downloadLocally`

userquin:
My commit should fix lint and build, we need to review logic for local font folder:
- check  nuxt package to override css folder in `resolveConfig`
- check astro package
- check sveltekit package

The preset will not initialize the css folder, the vite plugin will check if configured, if missing then it will use `viteConfig.publicDir`. Nuxt will require some logic to include the default value, and SvelteKit will require to change it to use `static` folder (IIRC kit will configure publicDir properly, I need to confirm it).

We also need to simplify the logic, we don't want to deal with public, assets and external folders: my suggestion is to use always `<public>/unocss-fonts`:
- Vite/Astro/Nuxt: `${viteConf.publicDir}/unocss-fonts`
- Kit: `static/unocss-fonts`

Of course, we also need to check Next, Webpack, Remix.

userquin:
We cannot use webfont preset in playground, it is client side only, we should add a new example and some tests.

userquin:
We also need to include preconnect and font in the entry point (index.html), rn not adding those head entries.

packages/nuxt/src/web-fonts.ts
```diff
@@ -0,0 +1,23 @@
...
+export function configureWebFontPreset(nuxt: Nuxt, options: UnocssNuxtOptions) {
+  const webFontPreset = lookupPreset(options, '@unocss/preset-web-fonts')
+  if (webFontPreset && !!webFontPreset.options?.downloadLocally) {
+    const downloadDir = `${nuxt.options.dir.public}/unocss-fonts`
+    webFontPreset.options.downloadLocally = {
+      downloadDir,
+      downloadBasePath: nuxt.options.app.baseURL,
+    }
+    nuxt.options.css ??= []
+    nuxt.options.css.push(`~/${relative(dirname(nuxt.options.dir.public), downloadDir)}/${defaultFontCssFilename}`.replaceAll('\\', '/'))
```
userquin, line 16: should use `unshift

userquin:
we'll need to include css and font folders: css should go to assets folders in Vite and styles folder in Astro and Nuxt (this way we can use some CSS processing optimizations, Astro for example excluding css styles in public folder => https://docs.astro.build/en/guides/styling/#load-a-static-stylesheet-via-link-tags)

userquin:
@onmax can you merge main and resolve the conflicts in the kit example? I will not touch the PR until tmr.

onmax:
Sorry, I was doing other stuff. Anything missing in the PR?

userquin:
> Sorry, I was doing other stuff. Anything missing in the PR?

We need to add webpack and CLI.

We should review the logic: check [nuxt google fonts](https://github.com/nuxt-modules/google-fonts/) and  [google-fonts-helper](https://github.com/datalogix/google-fonts-helper)

antfu:
Thank you for working on it! https://github.com/unocss/unocss/pull/3962 should cover this already.

Issue #3518, opened 2024-01-20
vue2/3 keep-alive + transformerVariantGroup hmr error


webaddkevin:
### UnoCSS version

0.58.3

### Describe the bug

In a Vue + Unocss project, using the keep-alive component, the code hot update during inactive pages will cause the components containing the "unocss transformerVariantGroup" code to reload as well

### Reproduction

https://github.com/unocss/unocss/assets/20873249/a3925961-8ac3-4bd9-89fc-a610e905fcfc

### System Info
demo：https://github.com/webaddkevin/uno-test

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

webaddkevin:
demo：https://github.com/webaddkevin/uno-test

MrFoxPro:
Probably I have similiar issue with SolidJS + UnoCSS as well. Enabling `transformerVariantGroup` make Solid's HMR go crazy on first run. But then, after caching. it works fine.

MrFoxPro:
Well, it doesn't fix issue, but whole page refresh is not so annoying
Had to disable HMR at all due UnoCSS continues to infinity refresh iteself

webaddkevin:
> Well, it doesn't fix issue, but whole page refresh is not so annoying Had to disable HMR at all due UnoCSS continues to infinity refresh itself



> Probably I have similiar issue with SolidJS + UnoCSS as well. Enabling `transformerVariantGroup` make Solid's HMR go crazy on first run. But then, after caching. it works fine.
> 
> UPD: @webaddkevin setting `mode: "per-module"` in config fixed issue for me.



> Probably I have similiar issue with SolidJS + UnoCSS as well. Enabling `transformerVariantGroup` make Solid's HMR go crazy on first run. But then, after caching. it works fine.
> 
> UPD: @webaddkevin setting `mode: "per-module"` in config fixed issue for me.

Thanks, setting 'mode: "per-module" is useful, problem solved

[...]

PR #3741, opened 2024-04-18
fix(variant-group): hmr in vite

Simon-He95:
close: #3518

### DIFF
--- a/packages/shared-integration/src/transformers.ts
+++ b/packages/shared-integration/src/transformers.ts
@@ -11,6 +11,7 @@ export async function applyTransformers(
   id: string,
   enforce: SourceCodeTransformerEnforce = 'default',
 ) {
+  ctx.uno.currentFile = id
   if (original.includes(IGNORE_COMMENT))
     return
 
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -14,6 +14,7 @@ export class UnoGenerator<Theme extends object = object> {
   public events = createNanoEvents<{
     config: (config: ResolvedConfig<Theme>) => void
   }>()
+  public currentFile?: string
 
   constructor(
     public userConfig: UserConfig<Theme> = {},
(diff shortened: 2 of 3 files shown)


antfu:
I am not sure about it - can you explain a bit more about the problem and why this would be helpful?

Simon-He95:
> I am not sure about it - can you explain a bit more about the problem and why this would be helpful?
The `affectedModules` themselves collect all the ids that need to be updated, but our HMR needs to filter these ids based on the `clientImportedModules` that depend on them, which should enable on-demand updates

Simon-He95:
Morning, @antfu 
I have a question, should we only hot update the current file, or the current file and its sub-files? If it is the current file, there is no need to add `ctx.affectedModules.add(id)` to the `transformerVariantGroup`. If necessary, it should Filter all changed pages based on dependencies on the current page

Simon-He95:
I feel that there is no need to add `ctx.affectedModules.add(id)` in `transformerVariantGroup`, because if you modify the `hasChanged file`, this plug-in will be used. If you do not modify the `hasChanged file`, it will also trigger the style update of `uno.css`.

packages/shared-integration/src/transformers.ts
```diff
@@ -40,7 +40,7 @@ export async function applyTransformers(
   }
 
   if (code !== original) {
-    ctx.affectedModules.add(id)
+    // ctx.affectedModules.add(id)
```
Simon-He95, line 43: It seems that this line of code is not needed. It will cause additional updates. For example, I modified file a, but multiple files may be added to the change list for the first time. Modification of file a will cause hot updates of all files. uno.css is also updated multiple times. If the a file changes, you should only need to initiate the hot update of the a file and the hot update of uno.css.

antfu, line 43:
```suggestion
    // Investigate if this is safe to remove: https://github.com/unocss/unocss/pull/3741
    // ctx.affectedModules.add(id)
```

antfu:
I am still not convinced - but I am ok have it and to experiments a bit

PR #3750, opened 2024-04-22
fix(cjs)!: remove `.default` suffix from d.ts and d.cts files

userquin:
This PR will revert changes in `.cjs` files, patching `d.ts` and `d.cts` instead.

I need to check a few packages, rn next example working with these changes (postcss).

(Maybe we can patch only postcss packages, there is no way yet to configure `postcss` plugin in Next and Webpack (?) with `.default` suffix: `unocss/postcss` and `@unocss/postcss`)

The `dist-verify.ts` script is wrong (in this PR)...

/cc @kwaa

We also have some problems when using old ESLint: check #3751

### DIFF
--- a/packages/unocss/astro.d.ts
+++ b/packages/unocss/astro.d.ts
@@ -1,2 +1,0 @@
-export * from './dist/astro'
-export { default } from './dist/astro'
--- a/packages/unocss/index.d.ts
+++ b/packages/unocss/index.d.ts
@@ -1,2 +1,0 @@
-export * from './dist/index'
-export { default } from './dist/index'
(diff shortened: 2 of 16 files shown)


antfu:
Instead of increasing the complexity of patching, maybe we could expose some proxy cjs file doing something like 

```ts
const mod = require('./dist/index.cjs')

module.exports = mod.default
// this line might not be needed if no named exports (eslint/postcss plugin for example)
Object.assign(module.exports, mod)
```

userquin:
> Instead of increasing the complexity of patching, maybe we could expose some "proxy cjs file" doing something like

This seems to work, I need to test now ESM:

imagen

userquin:
I still need to review `dist-verify` script...

antfu (approved):

PR #3772, opened 2024-05-06
feat(preset-icons): throw error when `prefix` empty

### LINKED ISSUE #3766: Preset icons empty `prefix` makes things very slow in Nuxt
### UnoCSS version

0.59.4 (but applies for older versions as well)

### Describe the bug

Once you use `prefix = ''` for the `presetIcons` package, the startup time goes up by A LOT.

image

### Reproduction

https://stackblitz.com/edit/nuxt-starter-v1rmrn?file=app.vue,unocss.config.ts,nuxt.config.ts

### System Info

Mac M2, Google Chrome, Nuxt

### Validations

zyyv:
Add detection of `prefix` to prevent incorrect use cases.
close #3766

### DIFF
--- a/packages/preset-icons/src/core.ts
+++ b/packages/preset-icons/src/core.ts
@@ -31,6 +31,10 @@ export function createPresetIcons(lookupIconLoader: (options: IconsOptions) => P
       unit,
     } = options
 
+    // eslint-disable-next-line unicorn/prefer-includes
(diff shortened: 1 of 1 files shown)


antfu:
Well, I am not sure about if we should "disallow" this. We could say it's not recommended because it would have impact to the performance - but in the end the cost is paid by the end users, and if they think it's worth it, I suppose we should let them do it

PR #3774, opened 2024-05-06
feat(preset-icons): add svg props customization

### LINKED ISSUE #3744: presetIcons using `?bg` shouldn't also apply 'width', 'height' or 'background-color'
### UnoCSS version

"@unocss/nuxt": "^0.59.3"

### Describe the bug

Perhaps not a bug, but also probably not a feature. 🤷 

When using icons via `presetIcons` it also applies additional CSS which interferes with already set `width`, `height`, and `background-color` values, which I don't think it should be.

This approach has probably been taken so that it gives it the best chance to be 'visible' when first applied, but it's not considerate of existing CSS, and makes it harder to use.

As an example, if I would like to use a specific icon as a background on my 'checked' checkbox (which already has styling) it will override the `width`, `height`, and `background-color`.

### Reproduction

https://unocss.dev/play/?html=DwEwlgbgBAxgNgQwM5ILwCIBGBzAtAOwFMBXAFwCcE5cBWABjqgAdcAWdAPgChgx8myUUgE8mhDDAAWhGAGtMAewAe6WIhQZMuAIxQtSBXDAg9uGIYXlcmRHNjFyBq0wV9ShclARMxCSvhhCAgUiKAB3HUZJSL08MMkwdyhyBWJ8EEIQHVhpOUyALjBcADMEADZcAyMsqRlZAH4cTh4AenAIDiA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAFVcESgF84WKBBBwA5FxUS6dPUiQYAHpFjpsAQy4AbeJhz4iJYKQAUyVFDO9ABccADaVKhhehJcdDA6ALSBThh65IhwUk7QoXq0aHpwagC6VCWydAAW0DCxMHSh-qj6sfFJ1bX1enkwVjCJTqRwfZYD0Pa4pBhwNQBuGFDBowN99k5mqGqyIgqN4ZFyokoqPgCUssLyjCxsnDx8-OeXRwqS0n6HqHQS6xihAEYAHQAJheqAkaFweSqrDAjQA9Aj6CAgdUEZsWmoLuUkNikEA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4XyA

### System Info

Universal issue. But present on MacOS / Chrome.

### Validations

zyyv:
I'll add a hook to add handling of props to the final output. Since `iconify` merges the final properties into `usedProps`, it may not be the properties that the user ultimately wants. 
close #3744


userquin:
I also saw this weird behavior on `roe.dev` (Daniel Roe website), adding variants in linked playground works: `checked:w-10 checked:h-10`.

packages/preset-icons/src/types.ts
```diff
@@ -81,8 +92,14 @@ export interface IconsOptions {
...
    */
   cdn?: string
+
   /**
    * Custom fetch function to provide the icon data.
    */
   customFetch?: (url: string) => Promise<any>
+
+  /**
+   * Custom the finnal additional props.
```
userquin, line 102: `Custom the final additional props.`

zyyv:
> I also saw this weird behavior on `roe.dev` (Daniel Roe website), adding variants in linked playground works: `checked:w-10 checked:h-10`.
> 
> https://github.com/danielroe/roe.dev/pull/1395/files

It is also possible to use additional styles to fix the default behavior, but this may be a bit redundant.😂

userquin:
@zyyv cannot be done via https://unocss.dev/presets/icons#icon-collection-customization ?

packages/preset-icons/src/types.ts
```diff
@@ -81,8 +92,14 @@ export interface IconsOptions {
...
+
   /**
    * Custom fetch function to provide the icon data.
    */
   customFetch?: (url: string) => Promise<any>
+
+  /**
+   * Custom the final additional props.
+   */
+  propsProcessor?: (props: Record<string, string>, collection?: string, icon?: string, svg?: string, mode?: string) => void
```
antfu, line 104: I wonder if we could have a more general purpose `processor` that could modify more than the props, but possibly also the `props`, `svg`, `cssObject`?

zyyv: I fixed it and provided more context for users to use.

packages/preset-icons/src/types.ts
```diff
@@ -81,8 +99,14 @@ export interface IconsOptions {
...
    */
   cdn?: string
+
   /**
    * Custom fetch function to provide the icon data.
    */
   customFetch?: (url: string) => Promise<any>
+
+  /**
+   * processor for the generated CSS object
```
antfu, line 109:
```suggestion
   * Processor for the CSS object before stringify
```

antfu (approved):

PR #3778, opened 2024-05-08
feat(preset-px-to-viewport): add px-to-viewport support

Blithe07:
This PR is implemented for mobile terminal adaptation. You only need to set the corresponding width of the design draft (there is a default width: 375) to convert 'px' into ’vmin‘.

Usage:
```
  <span inline-block w-3.75px h-3.75px text-3.75px -p-3.75px m-3.75px absolute top-3.75px>
    Text
  </span>
```

Generate:
```
{
.w-3.75px{width:1vmin;}
.h-3.75px{height:1vmin;}
.text-3.75px{font-size:1vmin;}
.-p-3.75px{padding:-1vmin;}
.m-3.75px{margin:1vmin;}
.top-3.75px{top:1vmin;}
}
```

### DIFF
--- a/alias.ts
+++ b/alias.ts
@@ -17,6 +17,7 @@ export const alias: Record<string, string> = {
   '@unocss/preset-attributify': r('./packages/preset-attributify/src/'),
   '@unocss/preset-icons': r('./packages/preset-icons/src/'),
   '@unocss/preset-mini': r('./packages/preset-mini/src/'),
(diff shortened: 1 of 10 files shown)


antfu:
Thank you for your contributions! However, please understand that there are so many units in CSS, and we could not possibly provide for every combination. The existing rem-to-px is served more like an example of how you can do that - since it's just a few line, I'd suggest you do the preset locally or publish is as a community preset. Thanks.

Issue #3073, opened 2023-09-03
Enhance `blocklist`


antfu:
### Clear and concise description of the problem

Context: https://github.com/unocss/unocss/issues/3070#issuecomment-1704427138

### Suggested solution

I think we could make `blocklist`:

- Accept a custom function
- Able to provide an additional meta
- Can be contributed by presets (so we could also have blocklist-only presets)
- Integrate with ESLint plugin

The interface I imagine:

```ts
blocklist: [
  'p-1', // plain string
  /$ma-(\d+)$/, // RegExp
  (selector) => { // custom string
    return selector.includes('foo')
  },
  [/$ma-(\d+)$/, { warn: true /* additional meta for this rule */ }], // print out warning
]
```

Open for discussions

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [ ] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

texastoland:
This would solve the user complaint which is primarily autocompleting other styles and possibly enforcing a convention 🙌🏼 The meta would be for tools like ESLint right?

louiss0:
I would like this feature. I wish it gave me access to the rules that were already being enforced. 
That way I could write a condition that would allow me to check for duplicates.

louiss0:
> This would solve the user complaint which is primarily autocompleting other styles and possibly enforcing a convention 🙌🏼 The meta would be for tools like ESLint right?

These are all the differences between preset-windi and preset-uno that I could find. 

| Uno                                     | Windi                                  |
| --------------------------------------- | -------------------------------------- |
| ma-*                                    | m-*                                    |
| pa-*                                    | p-                                     |
| op-*                                    | opacity                                |
| {bg\| text\| border}-{color}-{1-9}      | {bg \| text\| border}-{color}-{50-900} |
| decoration                              | underline                              |
| b                                       | border                                 |
| of                                      | overflow                               |
| position-{absolute \| fixed}            | {absolute \| fixed\| sticky}           |
| flex-{basis\|content\|justify\|items}-* | {basis\|content\|justify\|items}-*     |
| grid-{content\|justify\|items}-*        | {content\|justify\|items}-*            |
| trueGray                                | truegray                               |
|  c-gray                                         | text-gray*  |

[...]

PR #3783, opened 2024-05-09
feat(eslint-plugin): blocklist meta with message

enkot:
#3073 
Added additional meta for blocklist rule and `message` for Eslint plugin.

Example:
Screenshot 2024-05-09 at 23 11 58
Eslint output:
Screenshot 2024-05-09 at 23 13 00

Thinking if "Utility ... is in blocklist" head is ok, or use just message from the meta 🤔

Also thinking about autofix, to replace blocked utility with suggested one (which can be specified in the meta).


enkot:
Fixed eslint-plugin tests, trying to understand why preset-icons tests are failing.

packages/eslint-plugin/src/rules/blocklist.ts
```diff
@@ -15,14 +14,13 @@ export default createRule({
       description: 'Utilities in UnoCSS blocklist',
       recommended: 'recommended',
     },
+    messages: {
+      'in-blocklist': '\"{{name}}\" is in blocklist{{reason}}',
```
enkot, line 18: 💚 Nice

PR #3785, opened 2024-05-10
fix(preset-attributify): fix cannot match `after|before="content-['']"`

Skiyee:


antfu:
We introduced [`content-empty`](https://unocss.dev/interactive/?s=content-empty) for that. The case is that we don't want to sacrifices everybody's performance for an edge case.

Issue #3788, opened 2024-05-10
`--at-apply` Hex colors not working


pdsuwwz:
### UnoCSS version

0.60.0

### Describe the bug

Hex color is invalid after '--at-apply' is used:
```scss
.test-box {
  --at-apply: bg-#f8ce3d; //  <- Not working
  --at-apply: c-red;   // ✓ Correct rendering
}
```

### Reproduction

`v0.60.0` reproduction: https://stackblitz.com/edit/unocss-unocss-2mfxtm?file=package.json,src%2FApp.vue

`v0.59.4` reproduction: https://stackblitz.com/edit/unocss-unocss-7tdkfw?file=package.json,src%2FApp.vue

It seems that Hex colors are not compiling correctly at `v0.60.0`:

![image](https://github.com/unocss/unocss/assets/19891724/3dee8a98-0e6c-40e7-ba41-cd63d3829ff9)

### System Info

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

antfu:
You either use `bg-hex-f8ce3d` or `bg-[#f8ce3d]`

pdsuwwz:
> You either use `bg-hex-f8ce3d` or `bg-[#f8ce3d]`

Thanks, I tried `bg-hex-f8ce3d` and it worked, but `bg-[#f8ce3d]` still didn't work. Did the usage change? I remember the previous version with no problems at all:

![image](https://github.com/unocss/unocss/assets/19891724/a79d3940-dd83-40f3-8652-de1c68964b5b)

antfu:
Try `--at-apply: "bg-[#f8ce3d]"` maybe it's the CSS parser no longer parse `#` as part of the content anymore

15200596294git:
[...]

PR #3790, opened 2024-05-10
fix(transformer-directive): get raw string in directive, support slash & hash

### LINKED ISSUE #3784: `transformer-directives` transforms opacities incorrectly
### UnoCSS version

v0.60.0

### Describe the bug

```css
* {
  --uno: bg-black/5
}
```

Is transformed into

```css
* {
  --un-bg-opacity:1;
  background-color:rgb(0 0 0 / var(--un-bg-opacity));
}
```

### Reproduction

In the UnoCSS Playground:

1. add `transformerDirectives()` to `transformers` in _Config_ panel
2. add `* { --uno: bg-black/5 }` to  _Custom CSS_ panel

And you will see the background is totally black.

### System Info

Edge, Windows

### Validations

zyyv:
Since the ast generated by `csstree` does not meet this scenario, I will manually rearrange the ast. close #3784


antfu:
I guess https://github.com/unocss/unocss/issues/3788 also falls into a similar situation. And maybe `:` as well?

antfu:
I don't know if we could retrieve the `raw` content of the attribute value.

zyyv:
> I guess #3788 also falls into a similar situation. And maybe `:` as well?

For this case, `--uno: text-red:5;` There are two `:`, obviously this is not a correct css syntax.

I'll update the test case see the output.

zyyv:
If the node type was `Declaration`, those children node never has `Raw` type content. fixed #3788 
(The result of manually adjusting ast is hopefully a good solution. 😂)

antfu:
We could actually slice the code to get the raw string :)

antfu (approved):

hooray:
I want to know if this is a bug. Do I have to write it like this in the future  ?

`--at-apply: hover:bg-stone-1 dark:hover:bg-stone-9;` -> `--at-apply: "hover:bg-stone-1" "dark:hover:bg-stone-9";`

zyyv:
> I want to know if this is a bug. Do I have to write it like this in the future ?
> 
> `--at-apply: hover:bg-stone-1 dark:hover:bg-stone-9;` -> `--at-apply: "hover:bg-stone-1" "dark:hover:bg-stone-9";`

When multiple `:` appear without quotes, `css-tree` will parse incorrectly. At the same time, we should also follow the css writing rules. 
```diff
{ 
  prop: value;
}

{
- --at-apply: hover:bg-stone-1 dark:hover:bg-stone-9;
+ --at-apply: 'hover:bg-stone-1 dark:hover:bg-stone-9';
}
```

Issue #3754, opened 2024-04-24
React如果children中直接跟随文本的情况下会出现 Warning: Received `true` for a non-boolean attribute


cqiang1993:
### UnoCSS version

0.59.4

### 环境
vite: 5.2.0
react: 18.2.0
unocss: 0.59.4
typescript: 5.2.2

### 配置
vite.config.ts
```
import { defineConfig } from "vite";
import react from "@vitejs/plugin-react";
import UnoCSS from "unocss/vite";

// https://vitejs.dev/config/
export default defineConfig({
	plugins: [react(), UnoCSS()],
});

```
uno.config.ts
```
import {
	defineConfig,
	presetAttributify,
	presetIcons,
	presetTypography,
	presetUno,
	transformerAttributifyJsx,
} from "unocss";

export default defineConfig({
	theme: {},
	shortcuts: {
		// 这里可以放全局公共样式
		"h-btn": "h-48px w-100% bg-#5C33BE b-none text-white rounded-8px",
		"wh-full": "w-full h-full",
	},
	safelist: [],
	presets: [
		presetUno(),
		presetAttributify(),
		presetIcons({
			extraProperties: {
				display: "inline-block",
				"vertical-align": "middle",
			},
		}),
		presetTypography(),
	],
	transformers: [transformerAttributifyJsx()],
});

```
shimi.d.ts
```
// eslint-disable-next-line @typescript-eslint/no-unused-vars
import * as React from "react";

declare module "react" {
	interface HTMLAttributes<T> extends AriaAttributes, DOMAttributes<T> {
		flex?: boolean;
		relative?: boolean;
		text?: string;
		grid?: boolean;
		before?: string;
		after?: string;
		shadow?: boolean;
		w?: string;
		h?: string;
		bg?: string;
		rounded?: string;
		fixed?: boolean;
		b?: string;
		z?: string;
		block?: boolean;
		"focus:shadow"?: boolean;
	}
	interface SVGProps<T> extends SVGAttributes<T>, ClassAttributes<T> {
		w?: string;
		h?: string;
	}
}

```

### 页面代码
main.tsx
```
import ReactDOM from 'react-dom/client'
import App from './App.tsx'
import './index.css'
import 'virtual:uno.css'
import React from 'react'

ReactDOM.createRoot(document.getElementById('root')!).render(
  <React.StrictMode>
    <App />
  </React.StrictMode>   
)

```
index.css 
```
body {
  width: 100vw;
  height: 100vh;
[...]

Issue #3773, opened 2024-05-06
presetAttributify and transformerAttributifyJsx don't work in react18


tbtnb:
### UnoCSS version

0.58.6

### Describe the bug

when I use presetAttributify  and transformerAttributifyJsx in my project,it appears an error
![image](https://github.com/unocss/unocss/assets/60879811/f923e08d-90c3-4d35-94a1-8b72fa2a7286)
this is my uno.config.ts
```js
import { defineConfig } from 'unocss/vite'
import theme from './src/config/theme'
import {
  transformerAttributifyJsx,
  transformerDirectives,
  presetAttributify,
  presetUno
} from 'unocss'
import presetIcons from 'unocss/preset-icons'

export default defineConfig({
  exclude: [
    'node_modules',
    'dist',
    '.git',
    '.husky',
    '.vscode',
    'public',
    'build',
    'mock',
    './stats.html'
  ],
  presets: [presetUno({ dark: 'class' }), presetAttributify(), presetIcons()],
  transformers: [transformerDirectives(), transformerAttributifyJsx()],
  shortcuts: {},
  theme: {
    colors: {
      ...theme
    }
  }
})

```
this is my vite.config.ts
```js
import { defineConfig } from 'vite'
import react from '@vitejs/plugin-react'
import path from 'path'
import Unocss from 'unocss/vite'

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [Unocss({}), react()],
  resolve: {
    alias: {
      '@': path.resolve(__dirname, 'src') // src 路径
    }
  }
})

```

this is my react code
```tsx
const Home = () => {
  return (
    <div>
      <h1 mt-20>Home</h1>
    </div>
  )
}
export default Home
```
I also add shims.d.ts in the root path,but I don't think it can make this error
```ts
import type { AttributifyAttributes } from '@unocss/preset-attributify'

declare module 'react' {
  interface HTMLAttributes<T> extends AttributifyAttributes {}
}

```

### Reproduction

```tsx
const Home = () => {
  return (
    <div>
      <h1 mt-20>Home</h1>
    </div>
  )
}
export default Home

```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #3791, opened 2024-05-10
fix(transformer-attributify-jsx): matchedRule

Simon-He95:
close: #3754, #3773

First, I included all the syntax of JSX`={}`, and then further split `{}`,
Currently, we can support more complex situations in 
```
<div children={<div></div>} test={
()=> <div text-center>
       <div absolute></div>
     </div>
}></div>
```
However, nesting again is not supported, such as 

```
<div children={
  <div children={<div></div>}></div>
}></div>
```


test/transformer-attributify-jsx.test.ts
```diff
@@ -114,66 +120,98 @@ describe('transformerAttributifyJsx', () => {
...
-    const codeToString = code.toString()
-    blocklist.forEach((rule) => {
-      if (rule instanceof RegExp)
-        expect(new RegExp(`${rule.source}=""`).test(codeToString)).not.true
-      else
-        expect(codeToString).not.toMatch(`${rule}=""`)
-    })
   })
 
+  // it('blocklist', async () => {
```
antfu, line 153: Why commenting this out?

Simon-He95: Ah, I commented while debugging and forgot to cancel it. I will update it later

packages/eslint-plugin/src/types.ts
```diff
@@ -1,11 +1,3 @@
-declare module 'eslint' {
-  interface SharedConfigurationSettings {
-    unocss?: {
-      configPath?: string
-    }
-  }
-}
-
```
Simon-He95, line 8: This statement will affect the packaging failure of eslint-plugin and whether it can be removed.

image

antfu (approved):

Issue #3794, opened 2024-05-12
Directives should ignore inline comments


pauliesnug:
### UnoCSS version

0.60.0

### Describe the bug

Upon migrating to UnoCSS from Tailwind, we experienced an issue where a commented-out section (`/* overflow-hidden */;`) of an `@apply` directive was being applied, making the page unable to be scrollable.

### Reproduction

[...]

PR #3799, opened 2024-05-13
feat(transformer-directives): add removeComments

Simon-He95:
close: #3794

### DIFF
--- a/packages/transformer-directives/src/apply.ts
+++ b/packages/transformer-directives/src/apply.ts
@@ -47,6 +47,8 @@ export async function parseApply({ code, uno, offset, applyVariable }: Transform
   if (!body)
     return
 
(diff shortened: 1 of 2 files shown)


UltraCakeBakery:
Could you add support for `// comment` style comments? In one of my projects we use a different flavor of css that uses these comments instead. We had a similar issue with that.

Simon-He95:
> Could you add support for `// comment` style comments? In one of my projects we use a different flavor of css that uses these comments instead. We had a similar issue with that.您可以添加对 `// comment` 样式注释的支持吗？在我的一个项目中，我们使用了一种不同风格的 css，它使用这些注释。我们也有类似的问题。

sure

zyyv (approved):

Issue #3793, opened 2024-05-11
extractors not working with cli


gorango:
### UnoCSS version

latest (0.60.0)

### Describe the bug

Using @unocss/cli, [extractor-pug](https://github.com/unocss/unocss/tree/main/packages/extractor-pug) is unable to reach the compile step due to `ctx.id` being `undefined`.

In the repro (linked below), there is a reduced clone of the pug extractor that skips the `id` checks which produces the desired results (it is commented out in the `uno.config.js` file).

### Reproduction

https://stackblitz.com/edit/unocss-unocss-hywyra?file=uno.config.js

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #3803, opened 2024-05-14
fix(cli): extractors not working

Simon-He95:
close: #3793

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -145,6 +145,7 @@ export async function build(_options: CliOptions) {
       {
         preflights: options.preflights,
         minify: options.minify,
+        id: postTransform[0].id,
       },
     )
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/index.ts
```diff
@@ -145,6 +145,7 @@ export async function build(_options: CliOptions) {
       {
         preflights: options.preflights,
         minify: options.minify,
+        id: postTransform[0].id,
```
zyyv, line 148: @Simon-He95 Maybe the id at index 0 is not the correct id and may be filtered out.

Simon-He95: This is indeed the case, but postTransform has flattened the codes of all files together. Otherwise, you can provide a parameter to determine whether it is a file type. If not, then use the file to generate css and merge it. You what do you think

PR #3804, opened 2024-05-14
fix(cli): generate css with id

zyyv:
related #3803 #3793

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -140,13 +140,19 @@ export async function build(_options: CliOptions) {
       )
     }
 
-    const { css, matched } = await ctx.uno.generate(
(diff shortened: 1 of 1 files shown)


zyyv:
@Simon-He95 I haven't thought of a better way, but this will ensure that the id is accurate.

Simon-He95:
> @Simon-He95 I haven't thought of a better way, but this will ensure that the id is accurate.

I haven't seen your message. In fact, I have updated the PR to handle this situation. There are still some problems in your PR. The first point is that we can make a judgment in advance based on the filesId. If they all end with the same id, you can use the previous merge strategy, which is an optimization point. In addition, you only considered the merge and deduplication of preflights, and did not consider the deduplication problem on other layers.

PR #3810, opened 2024-05-16
fix(preset-mini): support composite props

### LINKED ISSUE #3807: Properties with multiple values are not matched against shortcuts.
### UnoCSS version

0.60.2

### Describe the bug

When using properties with multiple values (e.g. `transition-colors,opacity`), the vite plugin returns unmatched warnings.

```sh
[unocss] unmatched utility "transition-colors,opacity" in shortcut "input"
```

With the `uno.config.js`:
```js
import { presetUno } from "unocss";

/** @type {import("unocss").UserConfig<import("unocss/preset-uno").Theme>} */
const conf = {
    presets: [
        presetUno(),
    ],
    shortcuts: {
      "input": "w-32 h-10 bg-black hover:bg-red transition-colors,opacity"
    },
};

export default conf;
```

### Reproduction

1. Create a sample app with vite and unocss.
2. Add preset-uno, and a shortcut that depends on a `transition-prop,prop` value. e.g. `transition-colors,opacity` allowed by the `transition-` matcher.

Minimal reproduction: https://github.com/Curstantine/issue-unocss-shortcut-multi-property/

### System Info

OS: Windows 11
Package Manager: `pnpm@9.1.1`
Framework: Not applicable

### Validations

zyyv:
close #3807.


test/assets/preset-mini-targets.ts
```diff
@@ -840,6 +840,9 @@ export const presetMiniTargets: string[] = [
   'transition-property-[border]',
   'transition-200',
   'transition-opacity-200',
+  'transition-colors',
+  'transition-colors,opacity',
+  'transition-colors,opacity-200',
   'transition-color,background-color-200',
```
zyyv, line 846: I don't know why vitest not update those snap.

Curstantine:
Wouldn't this break anything that depended on `h.properties`?
For an example, `transition-height` and etc would completely break.
image
image

Curstantine:
Since this PR adds a composite prop to the CSS props array in handler utils, it more or less breaks the whole prop resolving mechanism.

I think the better approach is to remove the composite prop from that handler array, and guard the `h.properties` value like the codebase has done before. Moreover, since `h.properties` only returns valid CSS values, we don't have to transform it further.

So an approach like this passes the test cases. Though for some odd reason the snapshot matching fails.

```ts
function transitionProperty(prop: string): string | undefined {
  return h.properties(prop) ?? prop.split(',').flatMap(x => transitionPropertyGroup[x]).join(',')
}
```

image

antfu (approved):

PR #3838, opened 2024-05-24
feat: nuxt layers support

### LINKED ISSUE #3837: Automatically merge `uno.config`s from Nuxt layers

Currently we have to manually merge configs from `uno.config` files from Nuxt layers, basically duplicating the `extends` list from the Nuxt config:
```ts
// uno.config.ts
import { mergeConfigs } from 'unocss'
import baseConfig from './base/uno.config'
import uiConfig from './ui/uno.config'
import brandConfig from './brand/uno.config'

export default mergeConfigs([
  baseConfig, 
  uiConfig,
  brandConfig
])
```

### Suggested solution

I suggest automatically generate `uno.config.mjs` file in the `.nuxt` folder, based on the Nuxt config:
```ts
// nuxt.config.ts
export default defineNuxtConfig({
  extends: ['./ui', './base'],
})
```
```ts
// uno.config.mjs
import { mergeConfigs } from '@unocss/core'
import cfg0 from '/Users/enkot/projects/unocss/examples/nuxt3-layers/base/uno.config.ts'
import cfg1 from '/Users/enkot/projects/unocss/examples/nuxt3-layers/ui/uno.config.ts'

export default mergeConfigs([cfg0, cfg1])
```
The only thing is that you should specify different uno.config path for Eslint:
```ts
// eslint.config.js
import unocss from '@unocss/eslint-plugin'

export default [
  unocss.configs.flat,
  {
    settings: {
      unocss: {
        configPath: './.nuxt/uno.config.mjs'
      }
    },
    rules: {
      'unocss/blocklist': 'error',
    }
  }
]
```
We can add `nuxtLayers: true` option to the `@unocss/eslint-config` or create another `@unocss/eslint-config/nuxt` config for Nuxt.

### Alternative

Do not merge the root `uno.config`, so user should reexport it manually and there is no need to change the eslint `configPath` setting:
```ts
// uno.config.ts
import unoConfig from './.nuxt/uno.config.mjs'

export default unoConfig
```

### Additional context

_No response_

### Validations

enkot:
Closes #3837 

Automatically merges configs from each layer and generates `/.nuxt/uno.config.mjs` file based on the Nuxt config's `extends` list. Then  the generated config is passed to the Vite/Webpack plugin.

### DIFF
--- a/packages/nuxt/src/index.ts
+++ b/packages/nuxt/src/index.ts
@@ -97,7 +97,7 @@ export default mergeConfigs([${configPaths.map((_, index) => `cfg${index}`).join
 
     async function loadUnoConfig() {
       const { config: unoConfig } = await loadConfig<UserConfig>(process.cwd(), {
-        configFile: options.configFile,
+        configFile: options.nuxtLayers ? resolve(nuxt.options.buildDir, 'uno.config.mjs') : options.configFile,
       }, [], options)
 
       await nuxt.callHook('unocss:config', unoConfig)
--- a/packages/nuxt/src/index.ts
+++ b/packages/nuxt/src/index.ts
@@ -1,7 +1,7 @@
 import { dirname, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import process from 'node:process'
-import { addComponentsDir, addPluginTemplate, defineNuxtModule, extendWebpackConfig, isNuxt2, isNuxt3 } from '@nuxt/kit'
+import { addComponentsDir, addPluginTemplate, addTemplate, defineNuxtModule, extendWebpackConfig, findPath, isNuxt2, isNuxt3 } from '@nuxt/kit'
 import WebpackPlugin from '@unocss/webpack'
 import type { VitePluginConfig } from '@unocss/vite'
 import VitePlugin from '@unocss/vite'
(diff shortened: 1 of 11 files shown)


packages/nuxt/src/index.ts
```diff
@@ -73,30 +74,58 @@ export default defineNuxtModule<UnocssNuxtOptions>({
...
-      && unoConfig.transformers?.some(t => t.name === '@unocss/transformer-directives' && t.enforce !== 'pre')
-    ) {
-      const preset = nuxt.options.postcss.plugins.cssnano.preset
-      nuxt.options.postcss.plugins.cssnano = {
-        preset: [preset?.[0] || 'default', Object.assign(
-          { mergeRules: false, normalizeWhitespace: false, discardComments: false },
-          preset?.[1],
-        )],
+    if (options.nuxtLayers) {
+      addTemplate({
```
antfu, line 78: Do we need a template for that? We could probably just read configs from all layers and merge them directly in the module and passed the merged object to Unocss?

enkot: But in this case we will not be able tmake it work with Eslint. As I know there is no other way for eslint plugin to read Uno config

enkot: @antfu Do you see any downsides to generating uno.config file? I think we should not leave out the support of Eslint.

antfu (approved):

leosin:
Regarding the merging of layers, is there any specific example?
I tried it and did not find `/.nuxt/uno.config.mjs`

Issue #3847, opened 2024-05-28
0.60.3 上下文中的符号导致不能正确的生成class


Djuvl:
### UnoCSS version

0.60.3

### Describe the bug

更新到0.60.3后 vue模板中某些代码字符的组合方式 导致class不能正确生成

### Reproduction

https://unocss.dev/play/?html=DwZwxgTglgDgLgAhAUzgVxgPgFAHpcKHB4GGEDeABuQL6U0lkDkAtE9sLuNPDsHMgC2MADYBDATiIATKADcEALjDiQIALwAiJmJYAjJpoQwA9hDgQxUOFvgsArABZNUssFlzM5cggtpkCAD8CJQAjABMAMyUSghMkeGhTAg0NJwerunyfLgCwuKSQA&config=PTAEFMGcBsEsDsAuBaAJrSBDARtcz5wAPFOQ0eAe2QCcpEbYBjRcVZWAWwAdKbFIAKC69%2BoAN6DQoVOABmCcAGFK8BQHMANFNDc6kcIgCCiBrGwBXRLDkBPbdL1RDASSarID3fsMBVKtoAvqByNJScoADkFlRMkJCRgoLEoogy8pgW0GmyCoQqarDqABSS0jRZUABcoADaOtK1kUwWkIjhtJWRmhKg7tB8NZF0qJGggQC6XlM6kAAWfIgtAjVl0lEtbR3zi8uRQ6wkyNDqoIcofJjw6uCgCwBu4DRV58ismNDdOoFeTgYrdQa3mciH8lGKAEovI4fMZTIxLNY7JDocD-m4PKUgdJIEwPuAagBGAB0ACZUdImKh4EM5qZuJAqiAoJxifNgF91uMoToZoEIYIgA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4IgLgTghgdgzgMwPYQLYAkyoDYgFwJTZwCmAvkA

portrait中的class没有生成在media里
[...]

PR #3855, opened 2024-05-30
fix(preset-attributify): Some characters in code comments cause abnormal parsing results

Simon-He95:
close: #3847
Maybe he can't solve some special scenarios, but he can reduce some of the scenarios where problems occur.


zyyv:
Closed this as completed in https://github.com/unocss/unocss/pull/4325

PR #3856, opened 2024-05-30
Update postcss.md

HuangZumao:
when in Vue CLI ，Vue CLI has preseted  postcss，the file named ’postcss.config.mjs‘ can't be identified，should use ’postcss.config.js‘，then mjs will be more widely-used


antfu:
Vue CLI is no longer recommended. This is due to it using the legacy PostCSS version - we should put our best effort into moving to ESM, not instead recommanding doing the backward.

Issue #3816, opened 2024-05-17
feat: Add helper comments for extended-colors when previewing.


LoTwT:
### Clear and concise description of the problem

I extend a color in the `uno.config.ts` and use it in my project. When previewing, it displays the color in RGB, the unocss generated results. When developing, I want to check the color to match the figma which provides the HEX instead of RGB.

```ts
// uno.config.ts
export default defineConfig({
  theme: {
    colors: {
      "my-color": "#0070E0",
    },
  },
})
```

![image](https://github.com/unocss/unocss/assets/50254496/1b7b679e-24ee-4965-bb7e-b4a082d2dc6b)

### Suggested solution

![image](https://github.com/unocss/unocss/assets/50254496/47f4e105-eb56-41c3-8d3d-ad8cf8965ed4)

Could it add some helper comments like above?

### Additional context

Same proposal in the [hyoban/tailwindcss-classname-highlight](https://github.com/hyoban/tailwindcss-classname-highlight/issues/28).

_No response_

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #3858, opened 2024-05-30
feat(preset-mini): add rawColorComment

Simon-He95:
close: #3816


zyyv:
@Simon-He95  I think this option is more integrated in the vscode plugin and has nothing to do with the `core` package, and has nothing to do with parsing the output css.

Simon-He95:
> I think this option is more integrated in the vscode plugin and has nothing to do with the `core` package, and has nothing to do with parsing the output css.

Because it is difficult to disassemble, a field that is not used by the client is added to process it. This field is actually only provided for the vscode side.

packages/core/src/types.ts
```diff
@@ -916,4 +920,9 @@ export interface GenerateOptions<T extends boolean> {
    * If return extended "matched" with payload and count
    */
   extendedInfo?: T
+
+  /**
+   * show origin theme color
+   */
+  showOriginThemeColor?: boolean
```
antfu, line 927: We probably don't need a new option for that. We could always do that when `envMode` is `dev`

antfu:
Werid that the test if failing

Issue #3874, opened 2024-06-05
Chaining arbitrary values doesn't seem to work


soerenmartius:
### UnoCSS version

0.60.4

### Describe the bug

Chaining arbitrary values in UnoCSS doesn't seem to work, e.g. 

```
<div class="[&>*:not(:last-child)]:after:content-[',\00a0']">
  <dev>1</dev>
  <dev>2</dev>
</div>
```

This works as intended in TailwindCSS but doesn't seem to work in UnoCSS.

### Reproduction

https://unocss.dev/play/?html=DwEwlgbgBAxgNgQwM5ILwCIDaAyAfAKgC4A7AewBcAKQxJcgWhgAsw4QBKAXUIQDNyApgCdCMUsUET6mAOQAaADoAGJQiUzO6XACgoUUAIi4AjMAD0IQzr0GjAJnOWj2x5FxA&config=PTAEFMGcBsEsDsAuBaAJrSBDARtcz5wAPFOQ0eAe2QCcpEbYBjRcVZWAWwAdKbFIAKC69%2BoAN6DQoVOABmCcAGFK8BQHMANFNDc6kcIgCCiBrGwBXRLDkBPbdL1RDASSarID3fsMBVKtoAvqByNJScoADkFlRMkJCRgoLEoogy8pgW0GmyCoQqarDqABSS0jRZUABcoADaOtK1kUwWkIjhtJWRmhKg7tB8NZF0qJGggQC6XlM6kAAWfIgtAjVl0lEtbR3zi8uRQ6wkyNDqoIcofJjw6uCgCwBu4DRV58ismNDdOoFeTgYrdQa3mciH8lGKAEovI4fMZTIxLNY7JDocD-m4PKUgdJIEwPuAagBGAB0ACZUdImKh4EM5qZuJAqiAoJxifNgF91uMoToZoEIYIgA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4XyA

### Validations

[...]

PR #3875, opened 2024-06-06
fix(extractor-arbitrary-variants): quote in []

Simon-He95:
close: #3874


soerenmartius:
thanks for fixing this!

antfu (approved):

PR #3877, opened 2024-06-06
fix(transformer-variant-group): vscode  autocomplete & annotation

### LINKED ISSUE #2619: VS Code extension: no autocompletion for variant groups or their prefixes
### UnoCSS version

0.15.12, vs code: 0.51.12

### Describe the bug

When using [variant groups](https://unocss.dev/transformers/variant-group), autocomplete does not work inside parentheses:
image
(ignore the error, this is eslint complaining about the order or classes)

Also, the autocompletion for variant prefixes themselves (like`hover:`) does not work.

Also see: #1768 

### Reproduction

Use the current VS Code Plugin and [variant groups](https://unocss.dev/transformers/variant-group).

### System Info

_No response_

### Validations

Simon-He95:
close: #2619

image


packages/core/src/generator/index.ts
```diff
@@ -98,6 +98,10 @@ export class UnoGenerator<Theme extends object = object> {
     raw: string,
     alias?: string,
   ): Promise<StringifiedUtil<Theme>[] | undefined | null> {
+    // eslint-disable-next-line regexp/no-unused-capturing-group
+    if (/(^\(|\)$)/.test(raw))
+      raw = raw.replace(/(^\(|\)$)/g, '')
```
antfu, line 103: This is too specific for core. Core shouldn't aware of this and shouldn't do modifications to `raw`

Issue #2869, opened 2023-07-14
variant-group: left-side group


IlyaSemenov:
### Clear and concise description of the problem

As a developer using UnoCSS with transformer-variant-group, I want to put groups at the left side of the expression, such as in:

```
border-1 (first last)-border-0
```

expanding to:

```
border-1 first-border-0 last-border-0
```

Currently, that does not work, the groups only expand when they are on the right side.

### Suggested solution

-

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

zoeyzhao19:
This is just my poor opinions got from [document](https://unocss.dev/transformers/variant-group). 😄

`Variant group transformer` is introduced the [variant group feature](https://windicss.org/features/variant-groups.html) of Windi CSS, and Windi CSS  said this feature is to **Apply utilities for the same variant by grouping them with a parenthesis.** So I think the point is the same `variant` firstly.

That's what I think it make sense.

PR #3880, opened 2024-06-09
feat(variant-group): support left-side-group

Simon-He95:
close: #2869


antfu:
Thanks for taking time working on it - but I am not sure if it's worth it to support this in tradeoff of Complexity, performance, as well as more surfaces to conflicts/false-positive with other syntaxes in the user code.

antfu:
Oh, btw, the reason it's called variant group is that it applies to variants.

I still feel this is too risky for a little gain. In the linked issue, I guess it's probably better to improve the DX via shortcuts or variants.

Simon-He95:
> Oh, btw, the reason it's called variant group is that it applies to variants.
> 
> I still feel this is too risky for a little gain. In the linked issue, I guess it's probably better to improve the DX via shortcuts or variants.

It is indeed very risky. I did not consider very complex scenarios, such as `c2:(d1 d2):d3:a:(a b)`
If we provide a base version, such as only processing `(?:\s|^)\([\w\s-]+\)[:-]([\w-]+|\[[^\]]+\])`, we point at `(first second):xx` and `(first second):[]` temporary
image

There will be better ways to support more complex situations in the future.

antfu:
I still feel it's a bit too risky to introduce it in the core. Maybe we could design an API that allows a preset to override the variant-group functionality, so such ability could be provided as a preset for users to opt-in (if they want that feature so badly and willing to take the trade-offs)

PR #3890, opened 2024-06-13
feat(preset-mini): support Functionality transition-[...]

Simon-He95:
close: #1303


zyyv:
Refactor the transition parse logic, see #3924

PR #3905, opened 2024-06-17
perf(core): hoist cssvar

Simon-He95:
https://github.com/Simon-He95/css-var-hoist
image


antfu:
Why would we do this? Performance or something else? The reason we have CSS variables is for runtime overriding, like `text-red text-opacity-50`, which with this hoisting seems to no longer work?

Simon-He95:
image
I was wrong, I meant that bg-related colors will have `--un-text-opacity: 1`;
I'm thinking of raising it, which will reduce the size of CSS.

Simon-He95:
But it seems that the result of this improvement is fine. What do you think? If it is feasible, I will modify it again.
image
image

antfu:
This will cause another problem of inheriting:

```html
<div class="text-red text-opacity-40">
  <div class="text-green">This will have the opacity as well</div>
</div>
```

In the end, I think the size of CSS should not be a concern as gzip should take care of the duplications already.

PR #3914, opened 2024-06-19
test: error with mutiple-nest-body shortcuts

zyyv:
This is a test PR.

PR #3915, opened 2024-06-19
fix(core): support mutiple shortcuts with inline css value

zyyv:
close #3914


packages/core/src/generator/index.ts
```diff
@@ -703,9 +702,13 @@ export class UnoGenerator<Theme extends object = object> {
       }
     }
 
-    // expand nested shortcuts
-    if (isString(result))
-      result = expandVariantGroup(result.trim()).split(/\s+/g)
+    if (result) {
+      result = toArray(result).map((s) => {
```
Simon-He95, line 706: I think using `if else` will perform better, although `toArray` will make the code more concise. But I don't want to refactor at the expense of performance.

antfu (approved):

PR #3924, opened 2024-06-26
fix(preset-mini): compatible bracket parse

### LINKED ISSUE #1303: Preset Wind Missing Functionality `transition-[...]`

zyyv:
close #3890, resolve #1303 

This PR will move `transition properties` to `theme`, and support `bracket` parse.

> https://github.com/unocss/unocss/pull/1304#issuecomment-1190252927

When fully customized and using css variables, the default easing and duration time will also be used.


test/assets/preset-mini-targets.ts
```diff
@@ -848,13 +848,18 @@ export const presetMiniTargets: string[] = [
   'transition-colors,opacity-200',
   'transition-color,background-color-200',
   'transition-background-color,color-200',
+  'transition-$variant',
+  'transition-[width_height]',
+  'transition-[width_height_colors]',
+  'transition-[width_height_colors]-200',
```
Simon-He95, line 854: `transition-[width_height_colors]-200` This is a weird way of writing, I don't like it.

test/assets/preset-mini-targets.ts
```diff
@@ -848,13 +848,18 @@ export const presetMiniTargets: string[] = [
...
   'transition',
   'transition-revert-layer',
   'property-none',
   'property-all',
   'property-unset',
   'property-margin,padding',
   'property-padding,margin',
```
Simon-He95, line 861: `property-padding,margin`
 This will look a bit fragmented, and there may be some unknown regular expression parsing problems. It is recommended not to support it and keep `property-[padding, margin]`

test/assets/output/preset-mini-targets.css
```diff
@@ -961,7 +961,11 @@ unocss .scope-\[unocss\]\:block{display:block;}
 .box-content{box-sizing:content-box;}
 .box-inherit{box-sizing:inherit;}
 .transition{transition-property:color,background-color,border-color,outline-color,text-decoration-color,fill,stroke,opacity,box-shadow,transform,filter,backdrop-filter;transition-timing-function:cubic-bezier(0.4, 0, 0.2, 1);transition-duration:150ms;}
-.transition-200{transition-property:color,background-color,border-color,outline-color,text-decoration-color,fill,stroke,opacity,box-shadow,transform,filter,backdrop-filter;transition-timing-function:cubic-bezier(0.4, 0, 0.2, 1);transition-duration:200ms;}
+.transition-\[width_height_colors\]{transition-property:width,height,color,background-color,border-color,outline-color,text-decoration-color,fill,stroke;transition-timing-function:cubic-bezier(0.4, 0, 0.2, 1);transition-duration:150ms;}
```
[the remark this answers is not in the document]
zyyv: This is intentional. Since `colors` has many compose properties, so when using `[xxx]`, you can not only customize the properties, but also use the theme.

test/assets/preset-mini-targets.ts
zyyv: Uno users have been writing `transition` rules in this style for a long time, so I have adopted a compatible way of writing. Otherwise, it will be a breaking change.

test/assets/output/preset-mini-targets.css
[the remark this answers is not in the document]
Simon-He95: In our grammar, `_` is converted to spaces, which may cause ambiguity when used on other attributes.

test/assets/preset-mini-targets.ts
Simon-He95: I looked at the previous tests, well, I haven't seen it used like this before, I thought it was a feature

packages/preset-mini/src/_rules/transition.ts
```diff
@@ -51,9 +49,9 @@ export const transitions: Rule<Theme>[] = [
 
       else if (d != null) {
         return {
-          'transition-duration': theme.duration?.[d] ?? h.time(d),
           'transition-property': theme.transitionProperty?.DEFAULT,
           'transition-timing-function': theme.easing?.DEFAULT,
+          'transition-duration': theme.duration?.[d] ?? h.time(d),
```
Simon-He95, line 54: I think it is best to keep the original location because it will affect the inconsistency of the output file and the problem of cache not being hit.

antfu (approved):

PR #3934, opened 2024-06-28
docs: Fix an example code with nuxt merge config sample

popadicbranislav:
The sample is outdated as the import is now available directly from `unocss` and the signature of function `mergeConfigs` takes an array of configs instead of what the existing example states.

### DIFF
--- a/docs/integrations/nuxt.md
+++ b/docs/integrations/nuxt.md
@@ -82,7 +82,7 @@ export default config
 or modify/extend it:
 
 ```ts
-import { mergeConfigs } from 'unocss'
+import { mergeConfigs } from '@unocss/core'
 import config from './.nuxt/uno.config.mjs'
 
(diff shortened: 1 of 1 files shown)


docs/integrations/nuxt.md
```diff
@@ -82,12 +82,12 @@ export default config
 or modify/extend it:
 
 ```ts
-import { mergeConfigs } from '@unocss/core'
+import { mergeConfigs } from 'unocss'
```
zyyv, line 85: UnoCSS pkg exports a lot of information. I think we don't need to change the source to know where the configuration is imported from.
```suggestion
import { mergeConfigs } from '@unocss/core'
```

zyyv (changes requested):

zyyv:
Thank you for your careful discovery.

zyyv (approved):

Issue #3942, opened 2024-07-02
VSCode extension and autocomplete don't work


vabatta:
### UnoCSS version

0.61.0

### Describe the bug

As soon as I'm editing a `class="<cursor>"` no autocomplete comes up from the UnoCSS VSCode extension as long as I'm not fully extending the default theme. 

I was able to track down the issue to https://github.com/unocss/unocss/blob/e3d3b2ed21f02c213351d2d2c1d46ae57cae808b/packages/autocomplete/src/parse.ts#L68 with the error log from my extension output to be:
```
🛠 Resolving config for /Users/vabatta/Projects/vite-project
🛠 New configuration loaded from
  - /Users/vabatta/Projects/vite-project/uno.config.ts
ℹ️ 1 presets, 714 rules, 0 shortcuts, 44 variants, 0 transformers loaded
🗂️ Enabled context: /Users/vabatta/Projects/vite-project
⚠️ Error on getting autocompletion items
Error: Invalid theme key duration
	at /Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95345:23
	at Array.reduce (<anonymous>)
	at /Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95343:33
	at Array.map (<anonymous>)
	at parts.push.type (/Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95342:37)
	at /Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95312:5
	at Array.forEach (<anonymous>)
	at handleRegexMatch (/Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95308:35)
	at handleNonGroup (/Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95336:5)
	at /Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95370:9
	at handleRegexMatch (/Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95316:5)
	at handleGroups (/Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95360:5)
	at parseAutocomplete (/Users/vabatta/.vscode/extensions/antfu.unocss-0.61.0/dist/index.js:95329:3)
[...]

PR #3946, opened 2024-07-03
fix(vscode): Invalid theme key cause autocomplete not work

Simon-He95:
close: #3942

### DIFF
--- a/packages/autocomplete/src/parse.ts
+++ b/packages/autocomplete/src/parse.ts
@@ -60,15 +60,21 @@ export function parseAutocomplete(template: string, theme: any = {}, extraShorth
       input,
       /\$([\w.|]+)/g,
       (m) => {
-        parts.push({
-          type: 'theme',
-          objects: m[1].split('|').map((i) => {
+        const objects = m[1].split('|').map((i) => {
+          try {
             return i.split('.').reduce((v, k) => {
               if (!k || !v[k])
                 throw new Error(`Invalid theme key ${k}`)
               return v[k]
             }, theme)
-          }),
+          }
+          catch (_) {
+            return null
+          }
+        }).filter(Boolean)
+        parts.push({
+          type: 'theme',
+          objects,
         })
       },
       (str) => {


zyyv:
I think can not modify the `upstream` code directly, but should capture errors in `vscode`, then output error logs in the unocss panel.

packages/autocomplete/src/create.ts
```diff
@@ -6,6 +6,7 @@ import { parseAutocomplete } from './parse'
 import type { AutocompleteOptions, ParsedAutocompleteTemplate, UnocssAutocomplete } from './types'
 import { searchAttrKey, searchUsageBoundary } from './utils'
 
+export const vscodeErrors = new Set<string>()
```
zyyv, line 9: `autocomplete` pkg was a pure package, it should not have any other redundant logic processing. You should match the error output when using it.

packages/autocomplete/src/parse.ts
```diff
@@ -60,15 +63,26 @@ export function parseAutocomplete(template: string, theme: any = {}, extraShorth
...
+              vscodeErrors.add(String(e))
+            else
+              throw e
+
+            return null
+          }
+        }).filter(Boolean)
+        parts.push({
+          type: 'theme',
+          objects,
```
zyyv, line 85: Same as it, restore it.

zyyv (changes requested):

packages/autocomplete/src/parse.ts
Simon-He95: Do you have any good ideas? Because the try catch in the vscode environment cannot intercept the error here.

packages/autocomplete/src/create.ts
antfu: Yeah I think it should not be VSCode specific, and ideally should be handled in the VS code's code

chizukicn:
Maybe we don't need such strict key checking. We could stop throwing errors, or add an option to ignore them.

Issue #3950, opened 2024-07-03
Feature request: validate config at HMR time to prevent server crashes


henrikvilhelmberglund:
### Clear and concise description of the problem

I use autosave in my code editor which causes errors to occur sometimes because of code that isn't finished yet. This is usually not a problem but when editing a `uno.config.ts` config this usually causes a server crash because of `loadConfig()` failing to parse the config. 
You get an error like this:
```
file:///<project>/node_modules/unconfig/dist/index.mjs:177
          })(bundleFilepath);
            ^

Error: ParseError: C:\: Unexpected token
 <project>/uno.config.js:16:24
    at loadConfigFile (file:///<project>/node_modules/unconfig/dist/index.mjs:177:13)
    at async Object.load (file:///<project>/node_modules/unconfig/dist/index.mjs:98:24)
    at async loadConfig (file:///<project>/node_modules/@unocss/config/dist/index.mjs:43:18)
    at async reloadConfig (file:///<project>/node_modules/@unocss/svelte-scoped/dist/vite.mjs:282:33)
```

This feature request is for an additional config validation step to make sure the config is only loaded if it is actually a loadable config, otherwise load a default config until all errors are fixed.

### Suggested solution

Validate the config in `loadConfig()`, if any errors are found use the default config (or maybe the previous valid user config) until the errors are fixed and print a message in the console signaling that the config is invalid.

### Alternative

- disable autosave, works but would rather have it enabled
- just start the server again, also works but takes a few seconds, this feature would make things smoother

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

PR #3953, opened 2024-07-04
feat(config): introduce `createRecoveryConfigLoader`

Simon-He95:
close: #3950


packages/config/src/index.ts
```diff
@@ -56,14 +57,24 @@ export async function loadConfig<U extends UserConfig>(
...
+        ...result.config.configDeps.map(i => resolve(cwd, i)),
+      ]
+    }
+    historyLoadedConfig = result
 
-  return result
+    return result
+  }
+  catch (e) {
+    if (historyLoadedConfig) {
```
antfu, line 74: I don't think this is a robust approach. `loadConfig` should be side-effect free if possible. Also unconditionally cache the previous config would lead to tricky bugs as the arguments could be different, and the cache might be shared across different calls.

I would suggest move this handling to the higher-level abstraction, like `createCachedConfigLoader` where we save the cache inside the instance and the behavior could be more explicit

packages/config/src/index.ts
```diff
@@ -67,3 +67,26 @@ export async function loadConfig<U extends UserConfig>(
 
   return result
 }
+
+const loadConfigCache = new Map<string, LoadConfigResult<UserConfig>>()
+
+export async function createCachedConfigLoader<U extends UserConfig>(
```
antfu, line 73: `createCachedConfigLoader` should be a high-order function that referens a load function, with the cache created inside the function

Simon-He95: Would you like to make it a closure and put the cache inside the function?
image

antfu: Yeah exactly

antfu (approved):

Simon-He95:
`cwd` may be different when executing external and internal functions. I moved the parameters to the internal function.

henrikvilhelmberglund:
Right now `lastResolved` is reset each time so the if statement inside
```ts
if (lastResolved) {
        console.error(e)
        return lastResolved
      }
  throw e
```
never happens, which means that an error is always thrown inside `createRecoveryConfigLoader()`. So it doesn't really do anything right now in Svelte Scoped, does it work elsewhere?

antfu:
What do you mean by reset each time?

henrikvilhelmberglund:
> What do you mean by reset each time?

If I add a log here ([file](https://github.com/unocss/unocss/blob/5d58d365fa5c664e65c385e8e91beaaad1be3407/packages/config/src/index.ts#L86))
```ts
    console.log("lastResolved", lastResolved)
    try {
      const config = await loadConfig(cwd, configOrPath, extraConfigSources, defaults)
```

it will log `undefined`, each time I change something in the config, if `loadConfig()` has an error it stays undefined as the line after is `lastResolved = config` (eg. when a config was resolved) so it will always throw.

antfu:
Yeah, if the config has an error, it won't load - that's expected, isn't it? The point of this util is to reused the previous successfully loaded config during a dev server - if it would never correct, it can't be possible to be loaded.

henrikvilhelmberglund:
For me this is what happens:
1. Start Vite dev server and visit page, works fine
2. Edit `uno.config.ts` adding something that gives a syntax error
3. `createRecoveryConfigLoader()` throws an error and server crashes

(basically the same behavior before the PR)

What I want to happen:
1. Start Vite dev server and visit page, works fine
2. Edit `uno.config.ts` adding something that gives a syntax error
3. 
```
      if (lastResolved) {
        console.error(e)
        return lastResolved
      }
```
triggers and logs the error, returning the last used config
4. Server is still running

antfu:
Yeah - isn't that how it works? Or what do you suggest to change? I feel it's might be something on the callee side

henrikvilhelmberglund:
> Yeah - isn't that how it works? Or what do you suggest to change? I feel it's might be something on the callee side

Right now it's still crashing but I think it's something on the Svelte Scoped side, recreating the context on HMR, I'll see if it's fixable.

PR #3957, opened 2024-07-05
feat(vscode): introduce `autocomplete.strict` config

Simon-He95:


packages/vscode/src/autocomplete.ts
```diff
@@ -109,6 +109,20 @@ export async function registerAutoComplete(
...
 
+      const offset = window.activeTextEditor!.document.offsetAt(position)
+
+      const afterCode = code.slice(offset)
+        .replace(/"[^"<>{}]*"/g, '')
+        .replace(/\{[^}]*\}/g, '')
+        .replace(/'[^'<>{}]*'/g, '')
+      // get style range
+      const start = code.indexOf('<style')
+      const end = code.lastIndexOf('</style')
```
antfu, line 120: Could we extract this as a function like `shouldProvideAutocomplete` and add some unit test/snapshots to this so we can maintain it easier?

Simon-He95: sure

Simon-He95:
have some change to commit…

Simon-He95:
done :)

Issue #3978, opened 2024-07-12
CSS injected by UnoCSS incomplete on first load


KermanX:
### UnoCSS version

0.61.3

### Describe the bug

See https://github.com/slidevjs/slidev/issues/1573, and https://github.com/slidevjs/slidev/issues/1705.

When using UnoCSS with its Vite plugin, it seems that sometimes some UnoCSS class definitions are missing on the first load. Many Slidev users have encountered this, as well as antfu. However, I can't make this bug stably reproduced. Open this issue to track the bug and gather more information from a wider range of UnoCSS users.

### Reproduction

Failed to make it stably reproduced.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

Simon-He95:
I haven't used slidev much, can you provide a minimal demo of the occasional problem? And the steps to follow, what is normal and what is a problem?

KermanX:
> I haven't used slidev much, can you provide a minimal demo of the occasional problem? And the steps to follow, what is normal and what is a problem?

Thanks for your concern. This bug happens very randomly at a low rate. There is a video in https://github.com/slidevjs/slidev/issues/1705, which shows the unexpected behavior.

In the video, the `_uno.css` wasn't fully loaded on the first load, but the other code was loaded properly. After reloading the page, `_uno.css` got fully loaded - you can see the response in the console became much longer.

Theoretically, this can be reproduced in any Slidev project, but at a super low chance (about 1/1000 I think).

There are too many factors inside Slidev, making it too hard to extract which usage of UnoCSS causes this. I think we may leave the issue there and wait for other people to report their findings in other projects.

KermanX:
This is still happening after #3984 is merged 😥 And on my local machine, I think this occurs more frequently than when I opened this issue.

For Slidev, I opened #1969 as a temporary workaround.

PR #3984, opened 2024-07-14
fix(shared-integration): prevent potential data race in `flushTask`

KermanX:
Related to #3978, but it's hard to check whether this PR can fix it.

`flushTask` can be asynchronously called for multiple times at the same time, and previously, if this happens, the `tasks` array would be spliced twice - which may cause problems of losing styles.

### DIFF
--- a/packages/shared-integration/src/context.ts
+++ b/packages/shared-integration/src/context.ts
@@ -108,7 +108,8 @@ export function createContext<Config extends UserConfig<any> = UserConfig<any>>(
   async function flushTasks() {
     const _tasks = [...tasks]
     await Promise.all(_tasks)
-    tasks.splice(0, _tasks.length)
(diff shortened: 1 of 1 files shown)


packages/shared-integration/src/context.ts
```diff
@@ -108,7 +108,8 @@ export function createContext<Config extends UserConfig<any> = UserConfig<any>>(
   async function flushTasks() {
     const _tasks = [...tasks]
     await Promise.all(_tasks)
-    tasks.splice(0, _tasks.length)
+    if (tasks[0] === _tasks[0])
```
KermanX, line 111: This assumes all the tasks are different, which I think is true after searching in the codebase. It is also possible to make `tasks` a `Set`, but this may reduce performance.

antfu (approved):

PR #3995, opened 2024-07-17
chore(svelte-scoped): move tests out of source files into separate files

henrikvilhelmberglund:
This PR moves tests located in the source files into separate `x.test.ts` files.

`if (import.meta.vitest) {` was giving me some troubles using a linked `@unocss/svelte-scoped` package so I figured I could just move them out, also looks a bit cleaner imo.

Edit: @jacob-8 might want to take a look to see if it looks fine.


antfu (approved):
I am good with it :)

jacob-8:
I prefer the way that inline tests provide documentation without having to open up a second file, but if it's causing problems, I see no reason to resist the change. Svelte-scoped is pretty stable too, so I'm not in here often. Go ahead if this helps you use it better.

antfu (approved):

PR #3997, opened 2024-07-18
feat(preset-mini): support `(float|clear)-(start|end)`

OverflowCat:
Add logical support for `float` and `clear`: `float-start`, `float-end`, `clear-start`, and `clear-end`. [Tailwind has added these](https://tailwindcss.com/docs/float).

### DIFF
--- a/packages/preset-mini/src/_rules/position.ts
+++ b/packages/preset-mini/src/_rules/position.ts
@@ -150,6 +150,8 @@ export const floats: Rule[] = [
   // floats
   ['float-left', { float: 'left' }],
   ['float-right', { float: 'right' }],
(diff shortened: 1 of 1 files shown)


antfu (approved):

OverflowCat:
I'm not entirely sure, but it seems like the CI failures might not be directly related to the changes in this PR. `src/_preprocess/transformApply/getUtils.test.ts` has no `clear` or `float` class names.

henrikvilhelmberglund:
That test has a snapshot of the utility count so whenever you add utilities you have to increment the number in the snapshot. Eg. you added 4 utilities so the number went from `86` to `90`.

The test should probably be fixed.

Issue #4003, opened 2024-07-21
The customize option doesn't work.


lvjiaxuan:
### UnoCSS version

0.61.5

### Describe the bug

Following the docs's [saying](https://unocss.dev/presets/icons#global-icon-customization) about `customizations.customize`, It doesn't change the `width` and `height` of icons anyhow.

### Reproduction

[...]

PR #4004, opened 2024-07-21
fix(preset-icons): customize option doesn't work with the destructured default scale

lvjiaxuan:
fixed https://github.com/unocss/unocss/issues/4003


userquin:
The reproduction with wrong customize option: should be inside `customizations` wrapper:
```ts
customizations: { 
  customize(props) { ... } 
}
```

Anyway, the width/height still not being applied, I'm going to check what's happening on iconify utils package.

lvjiaxuan:
Sorry for carelessness.

Play.

> The reproduction with wrong customize option: should be inside `customizations` wrapper:
> 
> ```ts
> customizations: { 
>   customize(props) { ... } 
> }
> ```
> 
> Anyway, the width/height still not being applied, I'm going to check what's happening on iconify utils package.

Sorry for carelessness.

[play](https://unocss.dev/play/?html=DwEwlgbgBAxgNgQwM5ILwCIwFoAOALLBAOxjwHsAnLJMAWxzgFMsAXPMI9KAegD4g&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owCCMMUwARgK4zDoCeZF1tAJIBjAlT6UaMAKpYIcAL5x0UCCDgByNnOFUqGgFChIsOEwCGWKumghUUAGrnmlmAHFVbMMtXqNAAW0IXSoAegsrGyg7KABaADdnYFdYok8wQwNUAA8TeDR0czYAG3yMbDwCYgAKBAM4OCgS6gAuOABtdo1hNioYNVim4tQNMiRRYug2jUpkDUUAXQWSergqAAtoGB6YKja6hobu3v6QWI2tnY1pmByYWOKiMzvY6EsiVDhN%2BPsW2%2Bz7rdzMVRqsFCsGhFrLZ7HsOlCojEnC4sO50tUAJTLVb8KRw9qrBq42iyCCYiGHYn0RjMdicHjkwnkSRCURWWpMhrCZBYabrRhgPahULUEAAOg2oVBh0OPT6amAAC9zJwxPtObKTgrFahqhQIIKMYgNTL9YKxQB3YDIGDrOAAXk0ACZUCBDDKPcyDVQxetUMAiPyHc7Xe7PYdKDA2FAsF7BSbFBSZeCmQoMRTsWmDEA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4IgLgTghgdgzgMwPYQLYgFwKgGzgUwBpxp5k0AJMVHTbPfAXyA)

userquin:
Can you check `examples/vite-vue3`? It is not working, icons without w/h

packages/preset-icons/src/core.ts
```diff
@@ -51,9 +51,9 @@ export function createPresetIcons(lookupIconLoader: (options: IconsOptions) => P
           await customizations.iconCustomizer?.(collection, icon, props)
```
userquin, line 51: revert `scale` changes and set this call before `iconCustomizer` call:
```ts
customizations.customize?.(props as unknown as any)
```

userquin: should be (will work with iconify icons and custom svg icons):
```ts
customizations.customize?.(props as unknown as any)
await customizations.iconCustomizer?.(collection, icon, props)
```

userquin: running playground with previous changes:

imagen

userquin: @cyberalien we need to review iconify icon loader

lvjiaxuan: > should be (will work with iconify icons and custom svg icons):
> 
> ```ts
> customizations.customize?.(props as unknown as any)
> await customizations.iconCustomizer?.(collection, icon, props)
> ```

[check1](https://github.com/iconify/iconify/blob/e60da173c6a25a2cce7eb547519d3093ade82777/packages/utils/src/loader/modern.ts#L39): the **w/h** retruned from `customize()` are not passed to `iconCustomizer()`, which is not like what docs [saying](https://unocss.dev/presets/icons#icon-customizations) - "_apply iconCustomizer with customize customizations, if provided_.". (This should be intentional behaviour? Im not sure.)

[check2](https://github.com/iconify/iconify/blob/e60da173c6a25a2cce7eb547519d3093ade82777/packages/utils/src/loader/modern.ts#L59): And then, the explicit `scale = 1` take precedence over `customize()` to w/h.

So I think it's fine to just drop the destructured default scale value, and the `customize()` should work as expected.

userquin: Check vite-vue3 example, will not work

lvjiaxuan: > Check vite-vue3 example, will not work

Sorry, I can't get your meaning. I have checked this example and it looks fine. What's wrong with it? 

The shorcuts takes precedence as expected.

image

image

lvjiaxuan: Sorry, I have to go to bed now. Don't wait for my reply. I appreciate whatever u will do for this issue.

userquin: The page should have 2 black svg circles,  check the App.vue

userquin: Inspect the pages, both there but missing width/height in css styled (0px)

cyberalien: > @cyberalien we need to review iconify icon loader

Sure, but I can't do that right now. Should be back to computer in 3 days.

lvjiaxuan:
> Can you check `examples/vite-vue3`? It is not working, icons without w/h

https://github.com/iconify/iconify/blob/e60da173c6a25a2cce7eb547519d3093ade82777/packages/utils/src/loader/types.ts#L181

```ts
/**
 * The additional icon properties applied to the svg.
 *
 * The `width` and `height` will not be set to the `svg` if already present on it, and so, the `width` and `height` will be those you configure on the customizations.
 * If you omit the `width/height/scale` options and the `svg` contains the `width` and/or `height`, then, both will be extracted from the `svg`.
 *
 * If you need that properties just add an empty object here, useful for example when using the `svg` on `CSS`.
 */
usedProps?: Record<string, string>;
```

It says that we need to configure the `width/height/scale` options, if the `svg` doesn't contain them.

This PR has removed the scale and caused the icons to lose w/h.

Feeling sad for my PR is not working well. Looking forward to any suggestions.

packages/preset-icons/src/core.ts
cyberalien: Checked it and I'm very confused. Not sure why width/height doesn't work or how this pr fixes it.

cyberalien: > @cyberalien we need to review iconify icon loader

Issue is not in loader. I've added unit tests to test for this specific issue, loader returns correct icon with customised width and height.

See https://github.com/iconify/iconify/commit/84b4fab85f9817eaf91af74f37de2c6d009264eb

Renders `<svg viewBox="0 0 48 48" width="2em" height="1em" ><defs><path id="flatColorIconsUpRight0" fill="#3F51B5" d="M44 19L30 30.7V7.3z"/><path id="flatColorIconsUpRight1" fill="#3F51B5" d="M6 27v13h8V27c0-2.2 1.8-4 4-4h17v-8H18c-6.6 0-12 5.4-12 12"/></defs><use href="#flatColorIconsUpRight0"/><use href="#flatColorIconsUpRight1"/><use href="#flatColorIconsUpRight0"/><use href="#flatColorIconsUpRight1"/></svg>`

cyberalien: Actually, there is an error in loader. It happens when `scale` option is set. It overrides custom value.

Fixed in https://github.com/iconify/iconify/commit/439a862a1a0f0679038a0272bf06a9f277a50756

New behavior scales custom values instead of just returning `${scale}em`

cyberalien: Published `@iconify/utils` version `2.1.26` with fix.

Additional changes to `customize` callback:

It now has 2 more parameters:
- Icon data in `IconifyIcon` format, which is a mutable object. Can be used to get width/height ratio if you want to have different width and height. Because this is a mutable object, it can also be used to completely change icon content (change colors in icon with hardcoded colors, change animation duration for animated icons, add custom stuff and so on).
- Icon name in `${collection}:${name}` format, so you can check if you are customising correct icon.

Returning modified `props` is no longer required (but still there for compatibility). It is a mutable object, so callback can just change it. This is done because new icon data parameter is also a mutable object, so wouldn't make much sense to return props if you are modifying only icon data. It makes more sense to just change objects and not return anything.

cyberalien: Re-fixed in `2.1.28` (and I skipped a version by accident when publishing update, but it doesn't matter).

Option `scale` was broken for non-square icons, but there were no non-square icons in tests to play with, so issue was unnoticed and all icons were square.

In old version when option `scale` was not set, non-square icons were rendered correctly as non-square. If `scale` was set, non-square icons became square. That also affected wide icons by making them shorter than needed.

Now it is fixed, non-square icons have correct aspect ratio.

cyberalien: Added pull request that updates utils and updates unit tests: https://github.com/unocss/unocss/pull/4023

Should also fix this issue.

lvjiaxuan:
Thanks all.

PR #4016, opened 2024-07-26
 feat(playground): supports selecting a specific unocss version

### LINKED ISSUE #3895: 【Feature】Select alternative versions with playground support.

When I want to verify a problem with a version of unocss in my project, I can't reproduce it on playground (because it's always the latest version), so I can't provide the minimum repository when I want to reproduce the problem (the project is private).

### Suggested solution

We can provide playground with the option to select the unocss version in order to restore the problem most easily

### Alternative

_No response_

### Additional context

_No response_

### Validations

Tzyito:
close #3895 
Support selecting a specific unocss version to make regression testing easier, and allow reporting and reproducing bugs from both older and the latest versions

I am not sure how to fully test these changes, I have only tested them locally. If there is anything missing, please let me know


antfu:
Thanks for working on it, but the query doesn't seems to work: 

- https://deploy-preview-4016--unocss.netlify.app/play
- https://deploy-preview-4016--unocss.netlify.app/play/?version=AwOgbAjCBMQ

Tzyito:
> Thanks for working on it, but the query doesn't seems to work:
> 
> * https://deploy-preview-4016--unocss.netlify.app/play
> * https://deploy-preview-4016--unocss.netlify.app/play/?version=AwOgbAjCBMQ

I've resubmitted and fixed it

userquin:
There is small z-index overlap when opening versions selectetor, the panel separator is there:
imagen

Tzyito:
> There is small z-index overlap when opening versions selector, the panel separator is there: imagen

thx, I fixed it:)

henrikvilhelmberglund:
Could the version query string be changed to the actual version number instead of a hash?

Tzyito:
> Could the version query string be changed to the actual version number instead of a hash?

I recommend using a hash, as this prevents the user from manually entering the version and avoids reporting a pull failure error

userquin:
> Could the version query string be changed to the actual version number instead of a hash?

You can try using the version intead encoding it

henrikvilhelmberglund:
> I recommend using a hash, as this prevents the user from manually entering the version and avoids reporting a pull failure error

How? I can still write something else in the query string and fail to load a version.

Anyway not a huge deal, I just thought it would be nice to see the version number in the actual playground link if a user shares multiple playground links with for example a broken repro and a working repro.

Tzyito:
> How? I can still write something else in the query string and fail to load a version.

If an invalid version is entered, the latest version will be loaded by default. The specific version information will be read from the link and will not be directly reflected, as this aligns with the overall rules of the project

henrikvilhelmberglund:
> > How? I can still write something else in the query string and fail to load a version.
> 
> If an invalid version is entered, the latest version will be loaded by default. The specific version information will be read from the link and will not be directly reflected, as this aligns with the overall rules of the project

https://lustrous-klepon-6c72de.netlify.app/play/?version=AwOgbKAsx

Tzyito:
What I'm talking about is that if you enter version: 1.0.0 normally, the latest version will be used by default, instead of deliberately using the encoded version like you did. Deviated from your point of view(The coding is to solve invalid version input, and this is a user-side problem and has nothing to do with the playground.).

henrikvilhelmberglund:
Sure but if I use a version that actually exists:
https://lustrous-klepon-6c72de.netlify.app/play/?version=0.60.0
that doesn't work either but loads the latest version instead. For a user that tried to write the version as a semver string manually that looks like a bug.

Tzyito:
Do you mean to directly get the version value from the link? Wouldn’t this lead to input errors? I think we shouldn’t focus on whether we can manually modify the link to redirect to a specific version, but rather on choosing the version from the page. What do you think? Does anyone else have any suggestions?

henrikvilhelmberglund:
It's up to you but for me it's a bit hard to see what version this is
https://deploy-preview-4016--unocss.netlify.app/play/?version=AwOgbAjCCsQ 
just from the link. If it was
https://deploy-preview-4016--unocss.netlify.app/play/?version=0.61.5 
it's more obvious what version the playground link uses. So you could use the version picker on the page to choose a version that would put the version as for example 0.61.5 in the URL or just write `0.61.4` instead to change the version.

Some REPLs like Svelte works like this https://svelte.dev/repl/hello-world?version=4.1.0 but others like Vue does not.

I think showing the version in the link would be better for sharing repros when you could do something like

``` 
broken: https://deploy-preview-4016--unocss.netlify.app/play/?version=0.61.5 
works: https://deploy-preview-4016--unocss.netlify.app/play/?version=0.61.4
```

I'm also interested in what others think. Maybe I'm being too nitpicky.

antfu:
A few improvements are needed:

- Being able to use the version of the current development (versions that are not yet published, useful for previews) - In that sense I am not sure if an import map would be a good solution for that, also a bit concerned about the browser compatibility
- We should be able to have a link to the version release (link to commit sha if on preview)

Tzyito:
> We should be able to have a link to the version release (link to commit sha if on preview)

Do you mean a preview link? We can use the Netlify link.

> Being able to use the version of the current development (versions that are not yet published, useful for previews) - In that sense I am not sure if an import map would be a good solution for that, also a bit concerned about the browser compatibility

I am not sure if you mean selecting a preview of a specific commit version in the playground for development. I think Netlify can handle this, but I am not certain if it can access the packaged files.

I’ll handle the polyfill later.

antfu:
Screenshot 2024-08-01 at 13 15 48
The "This commit" in https://play.vuejs.org/

antfu:
No worries, I can take over. Thank you!

userquin:
can you change green version selector color?

imagen

antfu (approved):

henrikvilhelmberglund:
Looks great, thanks both of you!

userquin:
@antfu pink contrast on light still not good (same for the hash in dark theme)

PR #4022, opened 2024-07-28
fix: remove generic package export

tobiasdiez:
I got the following error with nuxt:
```
[DEP0166] DeprecationWarning: Use of deprecated double slash resolving ".//index.js" for module request ".//index.js" matched to "./*" in the "exports" field module resolution of the package at D:\Programming\JabRefOnline\node_modules\.pnpm\node_modules\@unocss\preset-wind\package.json
```
(I think this is windows specific)

This is fixed by removing the generic `./*` export field.


antfu:
Can you explain more? We can't simply remove it as it would affects the functionality - I would love to seek for a more robust solution

PR #4023, opened 2024-07-29
fix: update iconify utils to fix icon scaling

cyberalien:
Updated Iconify Utils, which fixes scaling for icons.

Old version of icon loader has a bug with scaling non-square icons:
- If `scale` option was not set, icon had correct aspect ratio
- If `scale` option was set, icon became square

New version correctly applies scaling to width and height, keeping icon's aspect ratio.

Also fixes https://github.com/unocss/unocss/issues/4003

Additionally, new version improves `customize` callback for icons, providing 2 new parameters:
- Mutable icon data that can be used to change icon content, such as changing colors in icons with hardcoded palette, adding shapes and so on.
- Icon name, so user knows which icon is being customised.


cyberalien:
Forgot one more change:
- `cwd` option in loader can now be an array, making it possible to load icon sets from multiple locations (such as main node_modules and package's node_modules in monorepos). See [pull request that implements it](https://github.com/iconify/iconify/pull/317).

PR #4042, opened 2024-08-02
docs(example): add angular postcss example

zcmk123:
add a example: angular + postcss + unocss
original angular project template from https://github.com/unocss/unocss/tree/main/examples/vite-angular

Angular Load tailwindcss with postcss, so we can config postcss, but only with json format config file

https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/utils/postcss-configuration.ts#L20


zyyv:
I think that any framework that postcss can work with can also work with unocss, so there is no need to add additional postcss related cases.

PR #4048, opened 2024-08-03
feat: replace fast-glob with tinyglobby

SuperchupuDev:
relevant: https://github.com/unocss/unocss/pull/3722#issuecomment-2053769072

[`tinyglobby`](https://github.com/es-tooling/module-replacements/blob/main/docs/modules/glob.md#tinyglobby) is meant to be a drop-in replacement to `globby` and `fast-glob`, and it only uses two subdependencies - `fdir` and `picomatch`. it's already being used in some somewhat big packages like [vite-plugin-pwa](https://github.com/vite-pwa/vite-plugin-pwa/pull/733) with no user reports of breaking behavior.

`expandDirectories` is explicitly disabled for better `fast-glob` compatibility


packages/postcss/src/esm.ts
```diff
@@ -105,12 +104,12 @@ export function createPlugin(options: UnoPostcssPluginOptions) {
     const globs = uno.config.content?.filesystem ?? defaultFilesystemGlobs
     const plainContent = uno.config.content?.inline ?? []
 
-    const entries = await fg(isScanTarget ? globs : from, {
+    const entries = await glob(isScanTarget ? globs : [from], {
```
userquin, line 107: cannot add `then` and find stats using `node:fs/promises`, we should add small cache to avoid process unchanged files

SuperchupuDev: i'm not sure i understand, do you mean using `node:fs/promises`' `stats` to use `fileMap` the same way it was used before?

userquin: yes, if the logic was wrong just fix it

SuperchupuDev:
tests should pass now, fixed an edge case around empty patterns

packages/postcss/src/esm.ts
```diff
@@ -135,14 +135,16 @@ export function createPlugin(options: UnoPostcssPluginOptions) {
 
     for (let i = 0; i < entries.length; i += BATCH_SIZE) {
       const batch = entries.slice(i, i + BATCH_SIZE)
-      promises.push(...batch.map(async ({ path: file, mtimeMs }) => {
+      promises.push(...batch.map(async (file) => {
         result.messages.push({
```
userquin, line 139: is this required? I mean, shouldn't be added only if the file has changed?

packages/postcss/src/esm.ts
```diff
@@ -135,14 +135,16 @@ export function createPlugin(options: UnoPostcssPluginOptions) {
...
           type: 'dependency',
           plugin: directiveMap.unocss,
           file: normalize(file),
           parent: from,
         })
 
+        const { mtimeMs } = await stat(file)
+
         if (fileMap.has(file) && mtimeMs <= fileMap.get(file))
           return
```
userquin, line 150: can you remove the `else`? We've return if the file is not changed

packages/postcss/src/esm.ts
SuperchupuDev: hmm, would moving `result.messages.push` after the `fileMap` check be a better option?

userquin: no idea, I'm just asking...

SuperchupuDev: it makes sense to move it (at least to me), and tests seem to pass, so i'll push it

Issue #3844, opened 2024-05-27
pnpm add -D @unocss/nuxt  bug


TomatoDroid:
### UnoCSS version

0.60.3

### Describe the bug

After running pnpm add -D @unocss/nuxt, there are no unocss packages in node_modules, resulting in a unocss not found ts error in uno.config.ts.
<img width="1163" alt="image" src="https://github.com/unocss/unocss/assets/22388686/d13e4ff8-6962-4d87-9ade-d16fd365d762">
<img width="310" alt="image" src="https://github.com/unocss/unocss/assets/22388686/e2cd0bf2-330e-4114-b20e-8c33da6e93ee">

I don't know why, I just find that the correct project with node_modules will have the unocss package。

thank you very much!!

### Reproduction

git@github.com:TomatoDroid/since-then.git

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

antfu:
You need to install `unocss` as devDeps, or enable hoisted mode in pnpm.

brokuka:
@antfu It would be great if you mentioned it in the documentation

Issue #3738, opened 2024-04-17
nuxt3 vscode No code prompt ；console output Error: Cannot find module 'unocss'


fjh1610:
### UnoCSS version

^0.59.3

### Describe the bug

nuxt3 vscode No code prompt ；console output Error: Cannot find module 'unocss'，Dependencies are installed
![image](https://github.com/unocss/unocss/assets/85326708/dd9c3b40-ab87-4a47-8b38-ef35e360ad3e)

![image](https://github.com/unocss/unocss/assets/85326708/cf7b301f-886d-44ee-a756-e875611e9877)

### Reproduction

no link

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

douxuefeng666:
I have the same question , plz help!!

Simon-He95:
We temporarily closed this due to the lack of enough information. We could not identify whether it was a bug or a userland misconfiguration with the given info.
Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) to reopen the issue.
Thanks.

> [Why reproduction is required](https://antfu.me/posts/why-reproductions-are-required)

fjh1610:
> We temporarily closed this due to the lack of enough information. We could not identify whether it was a bug or a userland misconfiguration with the given info.由于缺乏足够的信息，我们暂时关闭了它。我们无法确定这是错误还是给定信息的用户空间配置错误。 Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example) to reopen the issue.请提供最低限度的复制品以重新打开该问题。 Thanks. 谢谢。
> 
> > [Why reproduction is required 为什么需要复制](https://antfu.me/posts/why-reproductions-are-required)

It's done exactly as per the documentation, I don't know what else to provide

Simon-He95:
[...]

PR #4057, opened 2024-08-06
docs: improve intergration nuxt guide

TomatoDroid:
it might be able to reduce the subtype of issues. 
https://github.com/unocss/unocss/issues/3844
https://github.com/unocss/unocss/issues/3738

### DIFF
--- a/docs/integrations/nuxt.md
+++ b/docs/integrations/nuxt.md
@@ -90,6 +90,10 @@ export default mergeConfigs([config, {
 }])
 ```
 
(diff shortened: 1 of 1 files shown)


antfu:
We have already improved the docs https://github.com/unocss/unocss/pull/4015

For hoisting or not, I think it's a general pnpm problem, I would suggest users to fully understand the trade offs to make the decisions, instead of just "turn on this to fix the issue".

Thanks for bringing this up.

PR #4060, opened 2024-08-07
fix(preset-icons): missing default unit `em`

reslear:
I have a strange behavior if you don't set `unit` in config but tsdoc default is em
 https://github.com/unocss/unocss/blob/bdd82c3200c10c3ef3b922b47ebd392155ecb6c4/packages/preset-icons/src/types.ts#L90

style: 

```css
.i-belong-icon-checkmark{
  background:url("data:image/svg+xml;utf8,...") no-repeat;
  background-size:100% 100%;
  background-color:transparent;
  display:inline-block;
  width:24;
  height:25;
}
```

CleanShot 2024-08-07 at 16 57 18@2x

### DIFF
--- a/packages/preset-icons/src/core.ts
+++ b/packages/preset-icons/src/core.ts
@@ -29,7 +29,7 @@ export function createPresetIcons(lookupIconLoader: (options: IconsOptions) => P
       autoInstall = false,
       collectionsNodeResolvePath,
       layer = 'icons',
-      unit,
+      unit = 'em',
       processor,
     } = options
 


antfu:
Can you give a failing test first? I am not sure if the original behavior is actually desired (maybe the docs need to be improved instead).

reslear:
I saw that in the tests the default em and we had the same but then stopped writing em after some updates. 
and who and where controls the default value?

antfu:
The only usage of `unit` is here: https://github.com/unocss/unocss/blob/5d2e67bce40a60b04e727fd230e1e3b70776d154/packages/preset-icons/src/core.ts#L52-L57

Where when not provide, it defaults to what `iconify` outputs, which is `em`.

userquin:
@reslear can you provide the svg or a minimal reproduction?

userquin:
I guess the problem maybe is here https://github.com/iconify/iconify/blob/439a862a1a0f0679038a0272bf06a9f277a50756/packages/utils/src/loader/modern.ts#L68C17-L68C29

I'm 100%  sure since css w/h seems the svg icon w/h

/cc @cyberalien you changed only when there is `scale` and you should check if `iconToSVG` sets the unit

reslear:
https://www.svgviewer.dev/s/ottD7T7J

```html
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="25" fill="none">
  <g clip-path="url(#a)">
    <path fill="#2B6CFB" d="M9.86 18.75a1 1 0 0 1-.73-.32l-4.86-5.17a1 1 0 1 1 1.46-1.37l4.12 4.39 8.41-9.2a1 1 0 1 1 1.48 1.34l-9.14 10a1 1 0 0 1-.73.33h-.01Z"/>
  </g>
  <defs>
    <clipPath id="a">
      <path fill="#fff" d="M0 .75h24v24H0z"/>
    </clipPath>
  </defs>
</svg>

```

cyberalien:
> Where when not provide, it defaults to what `iconify` outputs, which is `em`.

That code will never be ran because props always contain width and height.

There are so many places in code where various customisations are set, it is a mess and that code where unit is set is not even considered by shared code in utils. I don't understand why there is a `unit` option at all, wouldn't it make more sense to provide number and units as one simple string? Set it as height option in props, loader should calculate correct width from icon's width/height ratio and keep units.

userquin:
@cyberalien does this icon https://www.svgviewer.dev/s/ottD7T7J follow iconify json convention, if I just change the svg to use `viewBox="0 0 24 25"` removing width and height attributes it just works:

imagen

cyberalien:
> @cyberalien does this icon https://www.svgviewer.dev/s/ottD7T7J follow iconify json convention?, if I just change the svg to use `viewBox="0 0 24 25"` removing width and height attributes it just works:
> 

If you remove width and height attributes, it will work perfectly when used as background/mask image, but will have weird size when used as svg.

Fill attribute on svg element is not used, it is overwritten by fill on path element.

serkodev:
> Can you give a failing test first? I am not sure if the original behavior is actually desired (maybe the docs need to be improved instead).

@antfu I created a issue (https://github.com/unocss/unocss/issues/4084) with reproduction.

Is it correct that the default SVG size should have the larger dimension (either width or height) set to `1em`, and the other dimension calculated as the ratio of the smaller to the larger dimension (i.e., `min(width, height) / max(width, height)`) ?

antfu:
@serkodev Could you send a PR with some failing tests? (this happens only for the custom icons, right?) This specific PR doesn't have a correct fix I think

serkodev:
Just submitted a draft PR (#4089) that contains failing tests.
Currently investigating on the `loadIcon` behavior in `@iconify/utils` to see if any solution ideas on `presetIcons`.

serkodev:
### Summary of the current research progress:

The issue started with the following commit: https://github.com/iconify/iconify/commit/412c8119b0e1b1a48e58304181d4b4798d59c697

```diff
- props[prop] = `${scale}em`;
+ props[prop] = calculateSize(
+   // Base on result from iconToSVG() or 1em
+   result?.[1] ?? '1em',
+   scale
+ );
```

The intention behind this change was to calculate the correct `width` and `height` based on the icon scale values obtained via `iconToSVG` ([source](https://github.com/iconify/iconify/blob/b6c141ea1412e306ac20338e1e6c76f4bf2e0c39/packages/utils/src/svg/build.ts#L187-L216)).

However, when using custom collections in `loadIcon`, `getCustomIcon` does not call `iconToSVG` to preprocess and calculate the SVG scale values ([source](https://github.com/iconify/iconify/blob/b6c141ea1412e306ac20338e1e6c76f4bf2e0c39/packages/utils/src/loader/loader.ts#L53)).  

As a result, the `calculateSize` function receives the `width` or `height` directly from the SVG, rather than the scale values processed by `iconToSVG`.

Currently, I see two possible directions for a solution:

1. Modify `getCustomIcon` in `@iconify/utils` to preprocess the SVG scale values.
2. Modify `@unocss/preset-icons` to calculate and preprocess the SVG scale values in the `transform` step.

antfu:
While I don't mind to fix it on our side, I guess it might be better to do in the Iconify utils as all the ecosystem would need/benefit from it. WDYT @userquin?

serkodev:
@antfu @cyberalien Thank you! Fixing it through `@iconify/utils` does seem more appropriate.  

By the way, I had earlier attempted to address the issue through `@unocss/preset-icons` and managed to resolve it partially, though some details still needed further work. ([serkodev/unocss#1](https://github.com/serkodev/unocss/pull/1)) I've temporarily closed that effort for now.

If any help is needed with `@iconify/utils`, I'd be more than happy to assist. 👍

userquin:
@serkodev thx

cyberalien:
Actually... won't fixing that bug break Unplugin Icons?

Same loader is shared with Unplugin Icons, I suspect expected behavior there is to keep dimensions from custom icon, not replace them with `1em`.

cyberalien:
I'm not sure changing code in utils is correct way to fix it.

In loader this is intended behavior. Function `mergeIconProps` checks if width and height are set on SVG, restoring them when generating final SVG. If this code is removed, custom icons will break in Unplugin Icons. In CSS those properties are irrelevant.

Problem is in extracting width and height from SVG in UnoCSS. I think it should be changed to use only to calculate width/height ratio, though even that can be irrelevant and easier calculated from viewBox. Then apply scale to it and add units it needed.

However, then customize callback becomes unusable for changing icon size.

So I can't think of a good solution that would work.

serkodev:
@cyberalien How about making it optional?

cyberalien:
> @cyberalien How about making it optional?

I tried to fix it, also considered option to make it optional, but loader code is very confusing, I'm afraid that fixing things will break something else. There are too many sources of props, not sure which one is used in which situation. Then customizer makes things even more confusing, that is used before props are extracted, resulting in different behavior for custom icons and icons loaded from icon sets.

With custom loader it is not just width/height that are broken. Scale option also does not work.

There is a solution though: to use Iconify Tools to import custom icons. It loads icons, convert them to icon set, then icon can be treated by UnoCSS like any other icon from icon set.

See https://github.com/iconify/tools/blob/main/%40iconify-demo/unocss/unocss.config.ts for working example.

latel:
hello everyone, is this problem already solved?

Issue #3660, opened 2024-03-27
[Bug Report] Latest update broke styling for flexbox properties in Nuxt3


rylanharper:
### UnoCSS version

0.58.7

### Describe the bug

Upon updating Unocss to `v0.58.7` in my Nuxt3 portfolio this morning, I notice a number of styles simply no longer work. Notably flexbox properties (and sometimes borders). **This does not seem to have an effect on live production**.

Here is a pic of a section in my portfolio in dev mode vs live - the styles have not changed at all between these two images. It seems that this issue only happens on the initial load of the page, although it can sometimes revert back to this state when I save changes on a file or component:

*These images are showing that `flex`, `items-center`, and `justify-between` not being applied to these elements. Again, styles have not been changed or edited in any way. Only the `unocss` version has changed.

<img width="484" alt="Screenshot 2024-03-27 at 9 19 09 AM" src="https://github.com/unocss/unocss/assets/16216160/22361370-3ac0-445e-ac65-6b7323887e4e">

<img width="647" alt="Screenshot 2024-03-27 at 9 22 07 AM" src="https://github.com/unocss/unocss/assets/16216160/ea74bb11-8672-4686-a5d1-85a979bd99e3">

### Reproduction

I can't seem to reproduce this in a Stackblitz as it seems to be local environment issue.. But here is the gist of what I have in my project, which always seems to be unstyled upon page load:

https://stackblitz.com/edit/nuxt-starter-xy4uvj?file=package.json,unocss.config.ts,app.vue,nuxt.config.ts

### System Info

Nuxt `v3.11.1`
Node `v18.18.2`

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

rylanharper:
Hmm so even when I downgrade versions this seems to still occur sometimes. I cannot seem to pinpoint the core issue here..

rylanharper:
[...]

PR #4078, opened 2024-08-16
fix(transformer-directives): multiple apply in on line

Simon-He95:
close: #3660


packages/shared-integration/src/transformers.ts
```diff
@@ -19,7 +19,8 @@ export async function applyTransformers(
     return
 
   const skipMap = new Map<string, string>()
-  let code = original
+  // There may be multiple @apply statements in one line. In this case, magic-string will process the calculation and modify the same position. Add a space here to handle this.
+  let code = original.replace(/;(@apply|--uno|--at-apply) /g, '; $1 ')
```
antfu, line 23: That would shift the position, which defeats the point of having source maps. Also, those values are customizable. I'd say it would be better to fix the position issue of magic-string usage.

Simon-He95:
This PR can fix the problem that `@apply` is finally compressed and merged into one line, but #3660 issue cannot be fixed because there may be extra `;` in a certain link in svelte, which becomes `@apply: a;;@apply b;`. I think this still needs to be further checked for the problem of extra `;`
image

Simon-He95:
The problem seems to be @sveltejs/vite-plugin-svelte

image

Simon-He95:
> The problem seems to be @sveltejs/vite-plugin-svelte
> 
> image

in svelte 5.0.0-next.225 has fixed :)

Issue #4084, opened 2024-08-20
bug(presetIcons): custom SVG icon loads invalid `width` and `height`


serkodev:
### UnoCSS version

from 0.61.7 to latest (0.62.2)

### Describe the bug

If I load a custom SVG for an `i-custom-arrow` icon with this preset config:

```ts
presetIcons({
  collections: {
    custom: {
      arrow: `<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 256 256"><path fill="currentColor" d="M204 88v104a12 12 0 0 1-12 12H88a12 12 0 0 1 0-24h75L55.51 72.48a12 12 0 0 1 17-17L180 163V88a12 12 0 0 1 24 0"/></svg>`,
    }
  }
})
```

The generated CSS output includes invalid values for `width` and `height`, which causes the icon to not display properly on the webpage:

```css
.i-custom-arrow {
    --un-icon: url(data:image/svg+xml;utf8,....);
    /* ... */
    width: 32; /* Invalid CSS value */
    height: 32; /* Invalid CSS value */
}
```

Before version `0.61.7`, the generated CSS used to be:

```css
width: 1em;
height: 1em;
```

Additionally, I'm unable to set the `width` and `height` using the `customizations.customize` function as described in the [documentation](https://unocss.dev/presets/icons#global-icon-customization).

```ts
// This does not work ❌
presetIcons({
  customizations: {
    customize(props) {
      props.width = '1em';
      props.height = '1em';
      return props;
    }
  }
})
```

However, I can set the dimensions using the `customizations.iconCustomizer` function:

```ts
// This works ✅
presetIcons({
  customizations: {
    iconCustomizer(collection, icon, props) {
      props.width = '1em';
      props.height = '1em';
    },
  }
})
```

---

I'm not sure if it is related to PR #4023, after updating `@iconify/utils` from `v2.1.25` to `v2.1.29`

### Reproduction

https://stackblitz.com/edit/github-59co11?file=uno.config.ts

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
[...]

PR #4089, opened 2024-08-21
test(preset-icons): add tests for loading custom svg icons

serkodev:
Related #4060, #4084


cyberalien:
Sorry, I don't think this can be fixed.

I tried to fix it, but loader code is very confusing, I'm afraid that fixing things will break something else. There are too many sources of props, not sure which one is used in which situation. Then customizer makes things even more confusing, that is used before props are extracted, resulting in different behavior for custom icons and icons loaded from icon sets.

With custom loader it is not just width/height that are broken. Scale option also does not work.

PR #4095, opened 2024-08-22
fix(attributify): add `b` type for jsx

yanhao98:
fix: 
image

fixed:
image

image


antfu:
This was intentionally not included, as `b` can be too generic; also, we want to limit the number of combinations generated; otherwise, TypeScript can go very slow.

yanhao98:
ok

PR #4104, opened 2024-08-27
pref: update `rulemeta` type usage

zyyv:


antfu:
You should really enable Corepack 👀

antfu:
How would this affect the performance?

zyyv:
> You should really enable Corepack 👀

I ensured that I enabled corepack and use the right packageManager version, I just run a `pnpm i` command after the catalog update, I don't know why it made a huge difference.
Anyway, this is a quick PR I made before leaving get off work, I may not have done a quick check. 😅

PR #4112, opened 2024-09-03
feat(runtime): add defineConfig export to `@unocss/runtime`, improve runtime bundler docs

henrikvilhelmberglund:
This PR adds a `defineConfig()` export to `@unocss/runtime` to make the config usage between for example a Vite integration and a runtime bundler integration more consistent.

Before I would have to do something like this

```ts
export default {
  // config here
}
```
```ts
import config from './my-config'
initUnocssRuntime({ defaults: ...config })
```

which works, but it's a big ugly, what I would really like (and what this PR does) is this:

```ts
import { defineConfig } from '@unocss/runtime'

export default defineConfig({
  // config here
})
```
```ts
import config from './uno.config'
initUnocssRuntime({ defaults: config })
```

This PR also improves the runtime bundler docs somewhat by showing how to pass a config to `initUnocssRuntime()` and how to import presets from `esm.sh`.


packages/runtime/src/index.ts
```diff
@@ -1,7 +1,98 @@
-import type { GenerateResult, UnoGenerator, UserConfig, UserConfigDefaults } from '@unocss/core'
+import type { Arrayable, CSSObject, GenerateResult, UnoGenerator, UserConfig, UserConfigDefaults } from '@unocss/core'
 import { createGenerator, isString, toArray } from '@unocss/core'
 import { autoPrefixer, decodeHtml } from './utils'
 
+interface Theme {
+  width?: Record<string, string>
```
antfu, line 6: These theme props are preset-specific, I am not very willing to hard-code them like this here. Could we figure out a way to reference the theme object from the preset somehow?

henrikvilhelmberglund: I changed it to use Theme from preset-uno since it is listed in package.json dependencies anyway `"@unocss/preset-uno": "workspace:*"`, does this seem right?

antfu (approved):

PR #4113, opened 2024-09-04
feat(transformer-directive): support `icon()` directive

### LINKED ISSUE #4111: Feature: icon() function in custom CSS

Sometimes you want to put an icon into a background, especially where it's difficult / undesirable / impossible to introduce additional elements, such as adding icons into forms inputs or styling a native checkbox input.

### Suggested solution

It would be nice to be able to apply just the inline image data-url into custom CSS styles.
I think the intuitive way to do this would be to provide a `icon()` function that is replaced by unocss, similar to `theme()`.

`icon('ph:user')` would be replaced by the data-url generated by this line in icons preset: 
```js
const url = `url("data:image/svg+xml;utf8,${encodeSvgForCss(svg)}")`
```

Example usage: 
```css
input[type=checkbox]:checked {
  background: icon('ph:check') no-repeat, white
}
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

zyyv:
close #4111


packages/rule-utils/src/icon.ts
```diff
@@ -0,0 +1,110 @@
...
+      },
+    },
+  }
+
+  const loader = await createNodeLoader()
+
+  for (const p of toArray(prefix)) {
+    if (icon.startsWith(p)) {
+      icon = icon.slice(p.length)
+      const parsed = await parseIcon(icon, loader, loaderOptions)
```
Simon-He95, line 62: Maybe it would be better to change it to parallel

packages/rule-utils/src/icon.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (body.includes(':')) {
+    [collection, name] = body.split(':')
+    svg = await loader(collection, name, options)
+  }
+  else {
+    const parts = body.split(/-/g)
+    for (let i = COLLECTION_NAME_PARTS_MAX; i >= 1; i--) {
+      collection = parts.slice(0, i).join('-')
+      name = parts.slice(i).join('-')
+      svg = await loader(collection, name, options)
```
Simon-He95, line 95: Maybe it would be better to change it to parallel

Simon-He95 (changes requested):

packages/rule-utils/src/icon.ts
zyyv: What I am considering is:
- In reality, the number of icon prefixes is generally not too large
- Parallel processing will result in redundant loaders for parsing. For my purpose, only one of them needs to have a value is enough.

Simon-He95: This will cause icon loading blockage, which may cause icon rendering to be too slow. In addition, this loader should be cached. In the case of cache, concurrency can have a faster advantage.

packages/rule-utils/src/icon.ts
userquin: We cannot move it to parallel, what Promise.xxx should we use? If we check every collection on every icon request in parallel that will result in blockage.

Anyway, that code will run only when using `i-<collection-name>-<icon-name>`, if you change to `i-<collection-name>:<icon-name>` that code will never be used.

userquin: Maybe it would be better to add a soft cache (iirc there is something like BetterCacheMap or similar) using icon as the key and the value with the collection and icon pair

zyyv:
@userquin I can't use map to cache previous results, because `iconify/utils` adds extra properties during the load process, like `usedProps`.
Do you have any good ideas?

userquin:
> @userquin I can't use map to cache previous results, because `iconify/utils` adds extra properties during the load process, like `usedProps`. Do you have any good ideas?
> 
> Or should I not extract the logic of load icon to `@unocss/presetIcon` and only use cache in `directive`?

You cannot and must no cache the content, you need to cache the original id adding to the soft cache map that id with the collection and the icon from the id once resolved.

packages/rule-utils/src/icon.ts
```diff
@@ -81,27 +88,31 @@ export async function createNodeLoader() {
...
 
-  if (body.includes(':')) {
-    [collection, name] = body.split(':')
-    svg = await loader(collection, name, options)
-  }
-  else {
-    const parts = body.split(/-/g)
-    for (let i = COLLECTION_NAME_PARTS_MAX; i >= 1; i--) {
-      collection = parts.slice(0, i).join('-')
-      name = parts.slice(i).join('-')
```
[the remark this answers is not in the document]
userquin: the load should be always done

userquin:
This is for example when using the external collection: https://github.com/iconify/iconify/blob/main/packages/utils/src/loader/external-pkg.ts#L50-L77

The promise will resolve once, the traverse the iconify collection should be fast to find the icon.

userquin:
fs loader will cache promises here https://github.com/iconify/iconify/blob/main/packages/utils/src/loader/fs.ts#L24

antfu (approved):

antfu:
LGTM. I'll let the others to finish their reviews before merging

packages/preset-icons/src/types.ts
```diff
@@ -4,7 +4,7 @@ import type { Awaitable, CSSObject } from '@unocss/core'
 
 interface IconMeta {
   collection: string
-  icon: string
+  name: string
```
userquin, line 7: is this a breaking change?

packages/rule-utils/src/icon.ts
```diff
@@ -0,0 +1,110 @@
+import type { IconsOptions } from '@unocss/preset-icons'
+import { encodeSvgForCss, type IconifyLoaderOptions, type UniversalIconLoader } from '@iconify/utils'
+import { type Preset, toArray, type UnoGenerator } from '@unocss/core'
```
userquin, line 3: use `import type {}` whenever possible

userquin:
We should review `rule-utils` build config file, we can have some problems, check current preset icons build config file https://github.com/unocss/unocss/blob/main/packages/preset-icons/build.config.ts#L3-L17

packages/rule-utils/src/icon.ts
userquin: (split types and modules into 2 static imports)

packages/preset-icons/src/types.ts
zyyv: In the #3774, I added handling of icon meta for process, but in this PR, referring to the previous variable naming, I think `name` is a better choice.
https://github.com/unocss/unocss/blob/31e6709c43068a3ff216c36c669007ab38fb33d3/packages/preset-icons/src/core.ts#L73-L75

Or I can revert the changes here.

userquin: We should avoid breaking changes.

I won't be available until next Tuesday.

antfu (approved):

jods4:
Fantastic, and you even added a parameter for changing the fill color!!
This opens up so many new possibilities!
🎉🙏 Thank you @zyyv

jods4:
I have the impression that `handleFunction` expects nodes of type `string`, does it exclude nesting a `theme('colors.red')` as the 2nd parameter?
The color string is then replaced unchanged into the inlined svg, so I assume `icon('check', 'red.400')` isn't working.
I don't see a test case that uses a theme color either, nor a mention in the docs.

It would be really nice to be able to use theme colors.
As the 2nd parameter is supposed to be a color, maybe `icon()` could attempt to resolve it as a `theme.colors` first and used that if it succeeds, so `icon('check', 'green.400')` could work.

PR #4115, opened 2024-09-04
docs: add how to enable inspector

userquin:


zyyv:
Hi @userquin 
The `inspector` is enabled by default and is not enabled in `uno.config`

```ts
// vite.config.ts
import { defineConfig } from 'vite'
import UnoCSS from 'unocss/vite'

export default defineConfig({
  plugins: [
    // other plugin...
    UnoCSS({
      inspector: false,
    }),
  ],
})
```

Issue #4021, opened 2024-07-27
The scoped attribute of the vue style tag seems to cause the failure of adding unocss @apply styles.


ying-tch:
### UnoCSS version

v0.61.6

### Describe the bug

When presetUno is configured with important
```
presets: [
    presetUno({
      important: '#app',
    }),
    presetAttributify(),
  ],
```
When style in the vue file has `scoped`, the `@apply` directive of unocss will fail.
![image](https://github.com/user-attachments/assets/c221e29d-3a8c-4aa9-af09-199bac8f806f)

### Reproduction

https://stackblitz.com/edit/vitejs-vite-yzjmbw?file=src%2FApp.vue

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

wkeylin:
You can use `mode:'vue-scoped' `to solve this problem

wkeylin:
https://unocss.dev/integrations/vite#vue-scoped
You can use `mode:'vue-scoped'` to solve this problem

ying-tch:
> https://unocss.dev/integrations/vite#vue-scoped You can use `mode:'vue-scoped'` to solve this problem

After I used `{ mode: 'vue-scoped' }`, I found that this problem was not solved, and there was also a problem with the normal use of unocss in the tag. I am not sure whether this is a problem with my configuration. This is the modified example:
https://stackblitz.com/edit/vitejs-vite-yzjmbw?file=vite.config.js
![image](https://github.com/user-attachments/assets/77730250-3915-4c40-aed9-4424767c30c0)

Simon-He95:
This is because vue-scoped adds scoped processing after uno processing, which causes problems

### after uno transformed:

<img width="226" alt="image" src="https://github.com/user-attachments/assets/bacbcdbd-e18d-47c0-a801-3aa9e909a18b">

### after `vue-scoped` transformed
[...]

PR #4124, opened 2024-09-06
fix(transformer-directives): vue-scoped generate css code

Simon-He95:
close: #4021

image


packages/transformer-directives/src/apply.ts
```diff
@@ -83,7 +84,8 @@ export async function parseApply({ code, uno, applyVariable }: TransformerDirect
     if (parent || (selectorOrGroup && selectorOrGroup !== '.\\-')) {
       let newSelector = generate(node.prelude)
       const className = code.slice(node.prelude.loc!.start.offset, node.prelude.loc!.end.offset)
-      if (selectorOrGroup && selectorOrGroup !== '.\\-') {
+      // if mode is 'vue-scoped', we don't need to scope the selector
+      if ((uno.userConfig as VitePluginConfig).mode !== 'vue-scoped' && selectorOrGroup && selectorOrGroup !== '.\\-') {
```
zyyv, line 88: This `mode` property is not on `userConfig`

Simon-He95: It seems that there are some problems with the type, but it does exist at runtime. It should be merged by the userConfig from vitePlugin.

zyyv: https://github.com/unocss/unocss/blob/243623ddb7decd7002e4110ef3674e1feeef6297/packages/config/src/index.ts#L59-L60

It seems that when loading the config, the vite plugin config is not merged into the `userConfig`

Simon-He95: https://github.com/unocss/unocss/blob/main/packages/vite/src/index.ts#L35
Go down from here and there is a merge operation

image

image

zyyv: image
image

vite plugin config is **not** injected into the `context`.

Simon-He95: https://github.com/unocss/unocss/blob/main/packages/vite/src/index.ts#L48
`ConfigHMRPlugin` will trigger reloadConfig and then inject
image

zyyv: I think `reloadConfig` still won't inject vite config in `userConfig`, it is a configuration item of `unocss` not `vite` plugin option. 🧐

Simon-He95: image

image
If you debug, you can see that userConfig already has `mode: vue-scoped` before transformerDirectives are executed.

zyyv:
I will finish the review today, give me some time. ❤️🥹

zyyv:
@Simon-He95 I'll close this pr because it has been resolved upstream.

Issue #4133, opened 2024-09-10
unocss hangs when using tsconfig paths imports in config file


rafael-lua:
### UnoCSS version

0.62.2

### Describe the bug

When I import a file using the tsconfig paths, unocss extension seems to hang and stop working until I reload the vscode window. It does not come back even if I change back to a normal import. The code runs fine, types and autocomplete for file imports works, but unocss extension just becomes nonfunctional. It does not print any error on vscode outputs (unocss, extensions, etc...)

I'm using a basic monorepo setup.

https://github.com/user-attachments/assets/f18073ce-de83-4eab-9f04-785b57393e6c

Here is the tsconfig I'm using:

```
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "compilerOptions": {
    "target": "ESNext",
    "module": "ESNext",
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "jsx": "preserve",
    "jsxImportSource": "solid-js",
    "types": ["vite/client"],
    "noEmit": true,
    "isolatedModules": true,
    "skipLibCheck": true,

    "baseUrl": ".",
    "paths": {
      "~": ["."],
      "~/*": ["./*"],
      "@": ["../.."],
      "@/*": ["../../*"]
    }
  },
  "include": ["**/*", "../../utils"]
}
```

### Reproduction

https://github.com/rafael-lua/repro-unocss-extension-issue

### System Info

Windows 11 (wsl2), Solid Framework, Monorepo with Changesets, typescript 5.3.3

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

zyyv:
I don't think there's a problem. Your tsconfig configuration doesn't provide any resolution for unconfig; you should use relative path references in unocss.config.

When loadContext fails, the vscode extension should stop working.

PR #4135, opened 2024-09-10
fix(vscode): when loadContextInDirectory error not add contextsMap

Simon-He95:
close: #4133
If an error is reported, do not append it to this.contextsMap, otherwise it will be cached and subsequent updates will not take effect.
By the way, the @ method of `unconfig` using alias does not seem to be supported.

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -68,6 +68,8 @@ export async function loadConfig<U extends UserConfig>(
   return result
 }
 
+export const errorCwdMap = new Set()
+
 /**
  * Create a factory function that returns a config loader that recovers from errors.
  *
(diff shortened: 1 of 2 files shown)


liangmiQwQ:
Met the same issue, hope it can be merged soon ❤️ (/cc @zyyv)

~~(Can also replace this line with `if (cached !== undefined && cached !== null)` to fix)
https://github.com/unocss/unocss/blob/6c99b3201f6abaacb9db29ce327d5d957d1936ea/packages-integrations/vscode/src/contextLoader.ts#L118~~

zyyv:
I think #4133 is fine and shouldn't be fixed.
@liangmiQwQ  If you have a new issue report, you can create a new pull request. Since VS Code extensions are inconvenient to test, please record relevant videos locally so I can review them.

PR #4143, opened 2024-09-18
feat(core): add symbols.layer

### LINKED ISSUE #4131: proposal: support dynamic layer (and layer order?) in rules

I'd like to be to have rules which put CSS in an arbitrary layer based on the class name, and set the layer order based on the class name. Just like we can currently use the rule regex to set parent, selector, property, and value, I'd like to be able to use the rule regex to set layer and to set that layer's order.

### Suggested solution

I've thought of three possible solutions

1. Have the `layers` config generate a block at-rule.
    _Supports a subset of possible needs_

    ```js
    export default defineConfig({
      layers: {
        stuff: 1,
        things: 2,
      },
      rules: [
        [
          /^m-([a-z]+)-(\d+)$/,
          ([, layer, size], { symbols }) => ({
            [symbols.parent]: `@layer ${layer}`,
            margin: `${size / 4}rem`
          }),
        ],
      ],
      // must leave `outputToCssLayers` false
    })
    ```
    ```html
    <div class="m-stuff-4"></div>
    ```
    ```diff css
    + @layer stuff, things;

    @layer stuff {
      .m-stuff-4 {
        margin: 1rem;
      }
    }
    ```
    Downsides:
    - `symbols.parent` can't be used for anything else in the same rule
    - `outputToCssLayers` can't be used (from UnoCSS's perspective, the CSS is in the `default` layer)
    - layer ordering can't itself be dynamic

    Benefits:
    - smallest change
3. Teach `symbols` about layers.
    🌟 _I like this forth best_
    ```js
    export default defineConfig({
      layers: {
        stuff: 1,
        things: 2,
      },
      rules: [
        [
          /^m-([a-z]+)-(\d+)$/,
          ([, layer, size], { symbols }) => ({
            [symbols.layer]: layer,
            margin: `${size / 4}rem`
          }),
        ],
      ],
      // optional: add `outputToCssLayers: true,`
    })
    ```
    ```html
    <div class="m-stuff-4"></div>
    ```
    ```css
    /* layer: stuff */
    .m-stuff-4 {
      margin: 1rem;
    }
    ```
    Downsides:
    - Two handles for controlling layer, one in the rule's second argument and one in the third argument.

    Benefits:
    - can use `symbols.parent` for other things
    - can choose to use `outputToCssLayers` or not
    - potential for teaching symbols.layers about order
        🌟🌟 _I like this third best_
        ```js
        export default defineConfig({
          layers: {
            stuff: 1,
            things: 2,
          },
          rules: [
            [
              /^m-([a-z]+)(-\d+)?-(\d+)$/,
              ([, layer, layerOrder, size], { symbols }) => {
                return {
                  [symbols.layer]: [layer, layerOrder], // would also accept `[symbols.layer]: [layer]`. maybe also accept `[symbols.layer]: layer`? or maybe use an object instead?
                  margin: `${size / 4}rem`,
                }
              },
            ],
          ],
          // optional: add `outputToCssLayers: true,`
        })
        ```
        ```html
        <div class="m-stuff-4"></div>
        <div class="m-other-0-4"></div>
        <div class="m-more-3-4"></div>
        ```
        ```css
        /* layer: other */
        .m-other-4 {
          margin: 1rem;
        }
        /* layer: stuff */
        .m-stuff-4 {
          margin: 1rem;
        }
        /* layer: more */
        .m-stuff-4 {
          margin: 1rem;
        }
        ```
        Would have to decide whether layers listed in the `layers` config get special treatment. I think probably no: simply override the config… In the above example, adding `<div class="m-stuff-4-4"></div>` would move the "stuff" layer to order 4.
4. Support dynamic `layer` values in rules' third arguments. Like the second argument, the rule regex's capture groups are available.
    🌟🌟🌟 _I like this second best_
    ```js
    export default defineConfig({
      layers: {
        stuff: 1,
        things: 2,
      },
      rules: [
        [
          /^m-([a-z]+)-(\d+)$/,
          ([,, size]) => ({
            margin: `${size / 4}rem`,
          }),
          ([, layer]) => ({
            layer: layer,
          }),
        ],
        // third argument would still accept the current shape
        [
          /^test$/,
          {
            margin: '1rem',
          },
          {
            layer: 'things',
          },
        ],
      ],
      // optional: add `outputToCssLayers: true,`
    })
    ```
    ```html
    <div class="m-stuff-4"></div>
    ```
    ```css
    /* layer: stuff */
    .m-stuff-4 {
      margin: 1rem;
    }
    ```
    Benefits
    - only one handle for controlling layer, in the rule's third argument
    - can use `symbols.parent` for other things
    - can choose to use `outputToCssLayers` or not
    - potential for teaching symbols.layers about order
        🌟🌟🌟🌟 _I like this best_
        ```js
        export default defineConfig({
          layers: {
            stuff: 1,
            things: 2,
          },
          rules: [
            [
              /^m-([a-z]+)(-\d+)?-(\d+)$/,
              ([,,, size]) => ({
                margin: `${size / 4}rem`,
              }),
              ([, layer, order]) => ({
                layer: [layer, order], // would also accept `layer: [layer]`. maybe also accept `layer: layer`? or maybe use an object instead?
              })
            ],
            // third argument would still accept the current shape
            [
              /^test$/,
             {
                margin: '1rem',
              },
              {
                layer: 'things',
              },
            ],
          ],
          // optional: add `outputToCssLayers: true,`
        })
        ```
        ```html
        <div class="m-stuff-4"></div>
        <div class="m-other-0-4"></div>
        <div class="m-more-3-4"></div>
        ```
        ```css
        /* layer: other */
        .m-other-4 {
          margin: 1rem;
        }
        /* layer: stuff */
        .m-stuff-4 {
          margin: 1rem;
        }
        /* layer: more */
        .m-stuff-4 {
          margin: 1rem;
        }
        ```
        Again, would have to decide whether layers listed in the `layers` config get special treatment. I think probably no: simply override the config… in the above example, adding `<div class="m-stuff-4-4"></div>` would move the "stuff" layer to order 4.

### Alternative

_No response_

### Additional context

_No response_

### Validations

henrikvilhelmberglund:
This PR adds `symbols.layer` which can be used to specify a layer either with an string/function or a regex match. `symbols.layer` does not make it possible to change the order of the layer, for that use `layers:` as before.

```ts
rules: [
      [/^color-(.*)$/, function * ([, color], ctx) {
        yield {
          color,
          [ctx.symbols.layer]: color,
        }
      }],
    ],
```
`color-red`:

```css
"/* layer: red */
.color-red{color:red;}"
```

fixes #4131

### DIFF
--- a/packages/core/src/generator/index.ts
+++ b/packages/core/src/generator/index.ts
@@ -10,6 +10,7 @@ export const symbols: ControlSymbols = {
   variants: '$$symbol-variants' as unknown as ControlSymbols['variants'],
   parent: '$$symbol-parent' as unknown as ControlSymbols['parent'],
   selector: '$$symbol-selector' as unknown as ControlSymbols['selector'],
+  layer: '$$symbol-layer' as unknown as ControlSymbols['layer'],
 }
 
 export class UnoGenerator<Theme extends object = object> {
@@ -614,6 +615,12 @@ export class UnoGenerator<Theme extends object = object> {
             else if (entry[0] === symbols.selector) {
               variants = [
                 { selector: entry[1] },
+                ...variants,
+              ]
+            }
+            else if (entry[0] === symbols.layer) {
+              variants = [
+                { layer: entry[1] },
                 ...variants,
               ]
             }
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -77,6 +77,7 @@ declare const SymbolShortcutsNoMerge: unique symbol
 declare const SymbolVariants: unique symbol
 declare const SymbolParent: unique symbol
 declare const SymbolSelector: unique symbol
+declare const SymbolLayer: unique symbol
 
 export interface ControlSymbols {
   /**
@@ -95,6 +96,10 @@ export interface ControlSymbols {
    * Selector modifier
    */
   selector: typeof SymbolSelector
+  /**
+   * Layer modifier
+   */
+  layer: typeof SymbolLayer
 }
 
 export interface ControlSymbolsValue {
@@ -102,6 +107,7 @@ export interface ControlSymbolsValue {
   [SymbolVariants]: VariantHandler[]
   [SymbolParent]: string
   [SymbolSelector]: (selector: string) => string
+  [SymbolLayer]: string
 }
 
 export type ObjectToEntry<T> = { [K in keyof T]: [K, T[K]] }[keyof T]
--- a/packages/core/test/__snapshots__/extended-info.test.ts.snap
+++ b/packages/core/test/__snapshots__/extended-info.test.ts.snap
@@ -29,6 +29,7 @@ Map {
             ],
           ],
           "symbols": {
+            "layer": "$$symbol-layer",
             "parent": "$$symbol-parent",
             "selector": "$$symbol-selector",
             "shortcutsNoMerge": "$$symbol-shortcut-no-merge",
@@ -75,6 +76,7 @@ Map {
             ],
           ],
           "symbols": {
+            "layer": "$$symbol-layer",
             "parent": "$$symbol-parent",
             "selector": "$$symbol-selector",
             "shortcutsNoMerge": "$$symbol-shortcut-no-merge",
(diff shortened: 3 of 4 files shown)


olets (commented):
That was fast! Cool.

Putting both yields in a single rule would let this file stay one-test-per-symbol.

antfu (approved):

Issue #4148, opened 2024-09-19
@unocss/preset-icons   does not support class names with capital letters


shenjunjian:
### UnoCSS version

0.62.4 

### Describe the bug

@unocss/preset-icons   does not support class names with capital letters。 
Regardless of whether the icon name is written directly in collections or in the iconify-json file, as long as it contains uppercase letters, the corresponding icon cannot be generated on demand

### Reproduction

[...]

PR #4149, opened 2024-09-20
fix(preset-icons): support captital letters in className

shenjunjian:
close: #4148 

support  captital letter in className

### DIFF
--- a/packages/preset-icons/src/core.ts
+++ b/packages/preset-icons/src/core.ts
@@ -67,7 +67,7 @@ export function createPresetIcons(lookupIconLoader: (options: IconsOptions) => P
       options,
       layers: { icons: -30 },
       rules: [[
-        /^([a-z0-9:_-]+)(?:\?(mask|bg|auto))?$/,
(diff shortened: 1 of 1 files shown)


zyyv:
We keep the lowercase syntax in all regular expressions by default, and why don't you use lowercase syntax anymore?

The capitalized syntax is weird when connected after `-` and it doesn't follow the camelCase naming convention. The lowercase icon name is consistent with `iconify`.

shenjunjian:
“We keep the lowercase syntax in all regular expressions by default”
You can encourage such behavior, but there's no need to restrict it to only such behavior. For example, when browsing the Rust standard library, the style there is not at all in line with JavaScript conventions, yet Rust is still a successful language.

zyyv:
> Isn't it a better choice to gain greater freedom with just a little change? @zyyv

@shenjunjian  I actually don't want to see any breakthroughs in this area, as it would create an update burden, such as `flex` --> `Flex` etc...  which makes UnoCSS more likely to be extracted, resulting in some invalid utilities.

We get icons from iconify, and I think we should follow the upstream naming conventions.

zyyv (approved):

Issue #4052, opened 2024-08-05
Lit, 'shadow-dom' mode, support @apply


felbry:
### Clear and concise description of the problem

https://stackblitz.com/edit/vitejs-vite-cgorxw?file=src%2Fmy-element.js

### Suggested solution

https://stackblitz.com/edit/vitejs-vite-cgorxw?file=src%2Fmy-element.js

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Read the `README.md` of using the package.
- [X] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

felbry:
考虑了下，这也可能是个伪需求，
至少针对当前情况：
```html
    <style>:host { @apply text-green text-2xl; }</style>
    <div class="text-red text-xl">
      我是第一行
    </div>
    <div>我是第二行，想继承host的文本样式</div>
    ...
```
只要保证root始终是一个div包裹即可，就像下边这样：
```html
<div class="text-green text-2xl">
  <div class="text-red text-xl">
    我是第一行
  </div>
  <div>我是第二行，想继承host的文本样式</div>
</div>
```
因此关闭这个问题

PR #4155, opened 2024-09-23
feat(transformer-directives): enable transformer directives to apply to embed style

enie123:
Enabling transformer directives for embed styles (e.g. <style> [CSS Code] </style> ) allows rules to be transformed in web components

Changes:
1. Change file filtering so that `ctx.filter(code, id)` is a fallback if a file fails to satisfy the `idFilter` which checks for `.css` 
2. Handle embed style via regex `/(<style[^>]*>)([\s\S]*?)<\/style>/gi`: 
     - Leverage the ability of `css-tree` can parse from offsets
3. Refactor tests to support both embed style and separate stylesheet cases: 
     - `describe.each` only supports FileSnapshots: `InlineSnapshot -> FileSnapshot`

Considerations: 
- Performance implications of file filtering changes
- Regex for style matching


zyyv:
Can you splite the commit for refactored the tests or move tests to another PR?

hannoeru:
I'm wandering what is the use case for this PR?
I think most component in Vite will be transform to dedicate css file to process, so you don't need extra tools to process the embed style, the only case I thought s to process embed style at pure HTML or using with CLI tool.

enie123:
> I'm wandering what is the use case for this PR? I think most component in Vite will be transform to dedicate css file to process, so you don't need extra tools/steps to process the embedded style, the only use case I thought is to process embedded style at pure HTML or using UnoCSS with CLI tool.

This would be a partial solution to #4052 (the top portion of https://stackblitz.com/edit/vitejs-vite-cgorxw?file=src%2Fmy-element.js) and also would allow for `@apply` when writing native web components when using `shadow-dom` mode: https://unocss.dev/integrations/vite#shadow-dom. 

Is there an alternative solution for this case?

hannoeru:
I think this should be a optional feature, because it introduce a lot of overhead.
Can you add some optional that user can turn it on or off? and I personally prefer it disabled by default.

enie123:
Yeah makes sense will fix when I get a chance.

antfu:
I am not sure about this either, ideally, UnoCSS should be as language-agnostic as possible - for such cases, I feel it should be handled by user-provided transformers (by wrapping the official transformer or somehow)

PR #4160, opened 2024-09-25
feat(preset-mini): support `bg-[image:*]`

JianJroh:
To support the following case:

```html
<div
  :style="{ '--hero': 'url(https://picsum.photos/200)' }"
  class="w-50 h-50 bg-[image:--hero]"
></div> 
```

Why not use `bg-[url(*)]` like this?

```html
<div
  :style="{ '--hero': 'https://picsum.photos/200' }"
  class="w-50 h-50 bg-[url(var(--hero))]"
></div> 
```

Because we cannot interpolate css variables with url like `url(var(--hero))`.

```css
.foo {
  --hero: 'https://picsum.photos/200';
  background-image: url(var(--hero));  /* Not Work */
}
```


zyyv:
LGTM. And we generally use `$variable` syntax to use css variables.

zyyv (approved):

packages/preset-mini/src/_rules/color.ts
```diff
@@ -23,8 +23,13 @@ export const bgColors: Rule[] = [
       return { 'background-size': h.bracketOfLength(d)!.split(' ').map(e => h.fraction.auto.px.cssvar(e) ?? e).join(' ') }
     if ((isSize(d) || bgPositionRE.test(d)) && h.bracketOfPosition(d) != null)
       return { 'background-position': h.bracketOfPosition(d)!.split(' ').map(e => h.position.fraction.auto.px.cssvar(e) ?? e).join(' ') }
-    if (bgGradientRE.test(d) || bgImageRE.test(d))
-      return { 'background-image': h.bracket(d) }
+    if (bgGradientRE.test(d) || bgImageRE.test(d)) {
+      const s = h.bracket(d)
+      if (s) {
+        const url = s.startsWith('http') ? `url(${s})` : h.cssvar(s)
```
JianJroh, line 29: The `url` seem to unnecessary for `bgGradientRE` case. 
How about it only for `bgImageRE` case?

zyyv: `bgGradientRE` and `bgImageRE` both produce `background-image` css, we can handle them together.

PR #4168, opened 2024-09-26
feat(vite): apply transformers to preflights during build

### LINKED ISSUE #4167: transformerDirectives in preflight not applied on Astro build
### UnoCSS version

0.62.4

### Describe the bug

If `@apply` is used in the UnoCSS preflights, it is not getting compiled when building an Astro application. It works fine in dev mode.

Example config: 
```ts
export default defineConfig({
  presets: [
    presetUno(),
  ],
  transformers: [transformerDirectives()],
  preflights: [
    {
      getCSS: () => `body {@apply text-orange;}`,
    },
  ],
});
```
Output when running `pnpm run dev`:
```css
body {
  --un-text-opacity: 1;
  color: rgb(251 146 60 / var(--un-text-opacity)) /* #fb923c */;
}
```

Output when running `pnpm run build`:
```css
body{@apply text-orange;}
```

See repo for reproduction

### Reproduction

https://github.com/HagenMorano/unocss-astro-directives-bug/tree/main

### System Info

Node 22.5.1
pnpm 9.11.0

### Validations

sibbng:
Fixes #4167

### DIFF
--- a/packages/vite/src/modes/global/build.ts
+++ b/packages/vite/src/modes/global/build.ts
@@ -225,9 +225,16 @@ export function GlobalModeBuildPlugin(ctx: UnocssPluginContext<VitePluginConfig>
           return null
         }
         const result = await generateAll()
-        const mappedVfsLayer = Array.from(vfsLayers).map(layer => layer === LAYER_MARK_ALL ? layer : layer.replace(/^_/, ''))
-        const importStatements = result.getLayer(LAYER_IMPORTS)
+        const importsLayer = result.getLayer(LAYER_IMPORTS) ?? ''
         const fakeCssId = `${viteConfig.root}/${chunk.fileName}-unocss-hash.css`
+        const preflightLayers = ctx.uno.config.preflights?.map(i => i.layer).concat(LAYER_PREFLIGHTS).filter(Boolean)
+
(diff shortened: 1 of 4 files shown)


packages/vite/src/modes/global/build.ts
```diff
@@ -226,13 +227,31 @@ export function GlobalModeBuildPlugin(ctx: UnocssPluginContext<VitePluginConfig>
...
+
+          let layerContent
+          if (layer === LAYER_MARK_ALL) {
+            layerContent = result.getLayers(undefined, [...mappedVfsLayer, LAYER_IMPORTS])
+          }
+          else {
+            layerContent = result.getLayer(layer.replace(/^_/, '')) || ''
+          }
+
+          if (layer === LAYER_PREFLIGHTS || layer === LAYER_MARK_ALL) {
```
antfu, line 244: While I understand that the ALL layer would contain preflights, but I am not sure if passing the whole generated CSS back to transformers is a good idea (for both performance and correctness concern).

I am not even sure if https://github.com/unocss/unocss/issues/4167 is a good idea to transform the preflights. But if we absolutely need it, maybe we should apply that in the preflight generation (before the layers been generated). But again I am not super sure about using directives in the preflights yet.

antfu (approved):

PR #4170, opened 2024-09-27
ci: use `--only-templates`

JianJroh:
Only show templates to avoid long comments.

Reference: https://github.com/stackblitz-labs/pkg.pr.new/pull/172


zyyv (approved):

zyyv:
I think this is good for now, although in the future we'd like `prg.pr.new` to provide a display of the main packages, with the rest collapsed.
At that time, we can update it again.

PR #4173, opened 2024-09-27
feat(webpack): support rspack/rsbuild

### LINKED ISSUE #4142: Is there a plan to support rsbuild?

Rsbuild has released version 1.0, and although PostCSS can be used to use Unocss, I think it is similar to Vite. It may be more convenient to create a plugin.

### Suggested solution

Develop a plugin

### Alternative

Configure PostCSS

### Additional context

_No response_

### Validations

sibbng:
Fixes #4142

Usage:
```tsx
import { defineConfig } from '@rsbuild/core';
import { pluginVue } from '@rsbuild/plugin-vue';
import { UnoCSSRspackPlugin } from "unocss/webpack";

export default defineConfig({
  plugins: [pluginVue()],
  tools: {
    rspack(config, ctx) {
      ctx.prependPlugins(UnoCSSRspackPlugin());
      config.optimization ??= {};
      config.optimization.realContentHash = true;
    }
  }
});
```


packages/webpack/src/index.ts
```diff
@@ -184,8 +184,23 @@ export default function WebpackPlugin<Theme extends object>(
...
 }
+
+export default function WebpackPlugin<Theme extends object>(
+  configOrPath?: WebpackPluginOptions<Theme> | string,
+  defaults?: UserConfigDefaults,
+) {
+  return unplugin(configOrPath, defaults).webpack()
+}
+
+export function UnoCSSRspackPlugin<Theme extends object>(
```
antfu, line 197: Named export will make the CJS interop hard - Maybe we should expose a new sub module with default export instead.

antfu: It can be a minor breaking change, I can bump a new minor version to be safe.

antfu (approved):

wtto00:
@sibbng Hello, How to use it?

Use it like in webpack: Add `import 'uno.css'` in `src/index.ts`？

```ts
import { createApp } from 'vue'
import App from './App.vue''

import 'uno.css'

import './index.css'

createApp(App).mount('#root')
```

But I encountered a problem: The unocss I wrote in `style.css` cannot be processed.

```css
/* src/style.css */
.body {
  --uno: m-0;
  @apply: bg-red;
}
```

The generated result is:
```css
.body {
  --uno: m-0;
}
```

But should be:
```css
```css
.body {
  margin: 0;
}
```

I have added the css file in the filesystem and configured the transformerDirectives plugin.
```ts
import { defineConfig, presetUno, transformerDirectives, transformerVariantGroup } from 'unocss'

export default defineConfig<Theme>({
  content: {
    filesystem: ['src/**/*.{html,js,ts,jsx,tsx,vue,svelte,astro,css}'],
  },
  presets: [
    presetUno(),
  ],
  transformers: [transformerVariantGroup(), transformerDirectives()],
})
```

There is the same issue in the postcss plugin.

But this works fine in Vite.

wtto00:
I create a example project: https://codesandbox.io/p/github/rspack-contrib/rsbuild-codesandbox-example/main

```js
// uno.config.mjs
import { defineConfig, presetUno } from 'unocss';

export default defineConfig({
  content: {
    filesystem: ['./src/**/*.{html,js,ts,jsx,tsx,css}'],
  },
  presets: [presetUno()],
  preflights: [
    {
      getCSS: ({ theme }) => `
        body {
          color: blue;
          --uno: bg-red;
        }
      `
    }
  ]
});
```

```css
/* src/index.css */
@unocss preflights;
@unocss default;

@unocss;

body {
    --uno: font-italic;
    color: green;
}
```

The results are:

- The output result of `font-size: 36px;` in `preflights` is correct, which is `font-size: 36px;`.
- The output result of `--uno: bg-red;` in `preflights` is not correct, which is `--uno: bg-red;` with no processing.
- The output result of `color: green;` in `index.css` is correct, which is `color: green;`.
- The output result of `--uno: font-italic` in `index.css` is not correct, which is `--uno: font-italic` with no processing.

wtto00:
I found a new issue, `@unocss/transformer-variant-group` is not working.

```jsx
<span className="hover:(c-red text-16) hover-font-bold hover:bg-pink">default style</span>
```

- `hover-font-bold hover:bg-pink` works well.
- `hover:(c-red text-16)` is not working.

wtto00:
Thank you for your reply. I will start trying right away.

wtto00:
Perfact, it works! Thanks a lot.

By the way, hope to add the usage and precautions to the document.

wtto00:
@Asukaaaaaa 
> Also you need to set `transformerDirectives({ enforce: "pre" })`

Asukaaaaaa:
that works fine in react, but looks not good in vue2😭
image

Asukaaaaaa:
Previously, in order to solve this problem in vue2, I had to use both the rspack plugin and the postcss plugin at the same time. But when there were many styles, doing so would cause some disturbing problems.

Asukaaaaaa:
> that works fine in react, but looks not good in vue2😭 image

Previously, in order to solve this problem in vue2, I had to use both the rspack plugin and the postcss plugin at the same time. But when there were many styles, doing so would cause some disturbing problems.

snecker:
`@apply` doenst work in vue3

rsbuild.config.ts
```javascript
import {defineConfig} from '@rsbuild/core';
import {pluginVue} from '@rsbuild/plugin-vue';
import {UnoCSSRspackPlugin} from "@unocss/webpack/rspack";
import {presetAttributify, presetMini, transformerDirectives, transformerVariantGroup} from "unocss";


export default defineConfig({
    plugins: [pluginVue()],
    tools: {
        rspack(config, ctx) {
            ctx.prependPlugins(UnoCSSRspackPlugin({
                    content: {
                        pipeline: {
                            include: [/\.(css|vue|svelte|[jt]sx|mdx?|astro|elm|php|phtml|html)($|\?)/],
                            exclude: [],
                        }
                    },
                    presets: [presetMini(), presetAttributify()],
                    transformers: [
                        transformerVariantGroup(),
                        transformerDirectives({enforce: "pre"}),
                    ]
                })
            );
            config.optimization ??= {};
            config.optimization.realContentHash = true;
        }
    }
});
```

index.ts
```typescript
import { createApp } from 'vue';
import App from './App.vue';
import "uno.css"
import './index.css';

createApp(App).mount('#root');
```

App.vue
```vue
<template>
  <div class="content text-green-500">
    Rsbuild with Vue3
  </div>
</template>

<style scoped>
.content {
  border:1px solid gray;   /* worked */
  @apply text-red-400;  /* doesnt worked */
  --at-apply: text-yellow-400; /* doesnt worked */
}
</style>

```

PR #4196, opened 2024-10-11
fix(transformer-attributify-jsx): exclude `node_modules` by default

### LINKED ISSUE #4195: transformerAttributifyJsx caused vite Pre-transform error
### UnoCSS version

0.63.4

### Describe the bug

Running `transformerAttributifyJsx` with `solid-sonner` produces the following error:
```sh
[vite] Pre-transform error: /Users/_/Desktop/ui/node_modules/.pnpm/solid-sonner@0.2.8_solid-js@1.9.2/node_modules/solid-sonner/dist/dev.jsx: Unexpected token, expected "," (440:58)

  438 |         return;
  439 |       }
> 440 |       toastRef?.style.setProperty("--swipe-amount", "0px=""");
      |                                                           ^
  441 |       setSwiping(false);
  442 |     }}
  443 |     onPointerMove={(event) => {
```

I checked `solid-sonner/dist/dev.jsx`, `transformerAttributifyJsx` may be converting this code
```jsx
toastRef?.style.setProperty("--swipe-amount", "0px");
```
into
```jsx
toastRef?.style.setProperty("--swipe-amount", "0px=""");
```

After uninstalling transformerAttributifyJsx, vite was successfully compiled.

### Reproduction

https://stackblitz.com/edit/unocss-unocss-j46jgm?file=src%2FApp.tsx

### System Info

_No response_

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).

zyyv:
close #4195

For this transformer, `node_modules files` should be excluded by default, otherwise there may be additional side effects.

If there is any special case I haven't considered, please feel free to close the PR.

### DIFF
--- a/packages/transformer-attributify-jsx/src/index.ts
+++ b/packages/transformer-attributify-jsx/src/index.ts
@@ -64,7 +64,7 @@ export default function transformerAttributifyJsx(options: TransformerAttributif
 
   const idFilter = createFilter(
     options.include || [/\.[jt]sx$/, /\.mdx$/],
-    options.exclude || [],
+    options.exclude || [/node_modules/],
   )
 
   return {


Simon-He95:
The key to the problem should not be here. Maybe we still need to solve this problem from the 		`transformerAttributifyJsx`.
`toastRef?.style.setProperty("--swipe-amount", "0px");` -> `toastRef?.style.setProperty("--swipe-amount", "0px=""");`

zyyv:
> The key to the problem should not be here. Maybe we still need to solve this problem from the `transformerAttributifyJsx`. `toastRef?.style.setProperty("--swipe-amount", "0px");` -> `toastRef?.style.setProperty("--swipe-amount", "0px=""");`

Yes, that needs to be done, I will record it and fix it in another PR.

Issue #3179, opened 2023-09-27
eslint-config: treat svelte conitional class as utilities


sunxyw:
### UnoCSS version

0.55.0

### Describe the bug

Eslint config/plugin will treat conditional classes inside the Svelte file as utilities and try to sort them, breaking the code.

## Example
![image](https://github.com/unocss/unocss/assets/31698606/49807c24-ad1f-4d02-8537-078c3b8eae3a)

Expected result:
```html
<li class="relative cursor-pointer select-none py-2 pl-3 pr-9 {active ? 'bg-primary10 text-primary-fg' : 'text-hue12'}">
```

Current result:
```html
<li class="relative cursor-pointer select-none py-2 pl-3 pr-9{active ? 'bg-primary10 text-primary-fg' : 'text-hue12'}">
```

> The space between `pr-9` and `{active` will be removed after `eslint --fix`.

### Reproduction

please see "describe the bug".

### System Info

@unocss/eslint-config: 0.56.4
@unocss/svelte-scoped: 0.55.0

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

jacob-8:
This is still a bug that'd be good to fix, but I want to mention that while `svelte-scoped` supports inline conditionals, I've entirely stopped writing them myself. When you lean into using the `class:bg-primary10={active}` syntax you'll find your code more readable and also it avoids these sorts of edge-case bugs like we have here.

PR #4205, opened 2024-10-16
fix(eslint-plugin): improve class ordering for Svelte components

vineryap:
related #3179 

- Add recursive checking for nested ternary expressions in Svelte class attributes
- Preserve spacing between class literals and mustache tags


packages/eslint-plugin/src/rules/order.test.ts
```diff
@@ -43,3 +44,61 @@ run({
     },
   ],
 })
+
+run({
+  name: 'order',
```
antfu, line 49:
```suggestion
  name: 'order-svelete',
```

antfu (approved):

zyyv (approved):

PR #4207, opened 2024-10-16
feat(eslint-plugin): support utilities ordering in Vue attributify attributes

### LINKED ISSUE #4007: Eslint plugin isn't working with attributify
### UnoCSS version

0.61.5

### Describe the bug

Eslint plugin isn't working with attributify. Utility classes are sorting inside `class` attribute, but not inside other attributes.
See reproduction, in `App.vue` wrong order classes order inside `class` and `un-before` attributes, but eslint warns only about `class`.

### Reproduction

https://stackblitz.com/edit/unocss-unocss-cfjwep?file=uno.config.ts

Go to terminal and execute:
eslint ./src/App.vue

```
❯ eslint ./src/App.vue

/home/projects/unocss-unocss-cfjwep/src/App.vue
  3:11  warning  UnoCSS utilities are not ordered  unocss/order

✖ 1 problem (0 errors, 1 warning)
  0 errors and 1 warning potentially fixable with the `--fix` option.
```

### System Info

_No response_

### Validations

vineryap:
- Add ATTRIBUTIFY_PREFIXES constant to include common UnoCSS-related prefixes
- Update VAttribute implementation

fix #4007 

https://stackblitz.com/edit/unocss-unocss-ptjzrg?file=src%2FApp.vue,patch%2F%40unocss%2Feslint-plugin%400.63.4%2Fdist%2Findex.mjs,package.json

https://github.com/user-attachments/assets/757114cb-e94a-42e5-96de-49cc41afb2d3

### DIFF
--- a/packages/eslint-plugin/src/constants.ts
+++ b/packages/eslint-plugin/src/constants.ts
@@ -1,2 +1,5 @@
 export const CLASS_FIELDS = ['class', 'classname']
 export const AST_NODES_WITH_QUOTES = ['Literal', 'VLiteral']
+
+// supports the default attributify prefix 'un-', and includes other potential UnoCSS-related prefixes for broader compatibility
+export const ATTRIBUTIFY_PREFIXES = ['un-', 'uno-', 'unocss-', 'un:', 'uno:', 'unocss:']
--- a/packages/eslint-plugin/src/rules/order.ts
+++ b/packages/eslint-plugin/src/rules/order.ts
@@ -1,7 +1,7 @@
 import type { TSESTree } from '@typescript-eslint/types'
 import type { ESLintUtils } from '@typescript-eslint/utils'
 import type { RuleListener } from '@typescript-eslint/utils/ts-eslint'
-import { AST_NODES_WITH_QUOTES, CLASS_FIELDS } from '../constants'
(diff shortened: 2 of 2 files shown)


packages/eslint-plugin/src/constants.ts
```diff
@@ -1,2 +1,5 @@
 export const CLASS_FIELDS = ['class', 'classname']
 export const AST_NODES_WITH_QUOTES = ['Literal', 'VLiteral']
+
+// supports the default attributify prefix 'un-', and includes other potential UnoCSS-related prefixes for broader compatibility
+export const ATTRIBUTIFY_PREFIXES = ['un-', 'uno-', 'unocss-', 'un:', 'uno:', 'unocss:']
```
zyyv, line 5: For attribute mode, we only have the default prefix `un-`
I don't know if this prefix list can be passed in via configuration.

zyyv (approved):

packages/eslint-plugin/src/constants.ts
vineryap: I haven’t looked deeply into the code yet, but as far as I know right now, we don’t have access to the user’s configuration, which makes the prefixes not configurable at the moment.
I'll explore it more when I'm free.

antfu: Yeah, I would feel this should be configurable, and the default to only be `un-` to match the preset's default

Issue #4208, opened 2024-10-16
remove commented code make compile pass


oluceps:
### UnoCSS version

0.63.4

### Describe the bug

As titled, deleting code that has already been commented could make the compile pass.

```patch
--- a/src/ingredients/quick-link.tsx
+++ b/src/ingredients/quick-link.tsx
@@ -24,13 +24,7 @@ export const QuickLinks: ParentComponent<QuickLinksProps> = (props) => {
 	// 	setInnerW(0);
 	// };
 
-	// const innerWText = createMemo(() => `w-[${innerW()}%]`);
 
-	// onCleanup(() => clearTimeout(timeoutId));
-
-	// <a href={props.href} target="_blank">
-	// 	<span class="absolute -inset-px rounded-xl" />
-	// </a>
 	return (
 		<div/>
 	);
-- 
```

BTW I know the commented code actually didn't work, thus commented.

### vinxi build

```
> pnpm build

> 's blog@ build /home/elen/Src/blog.nyaw.xyz
> pnpm gen && vinxi build


> 's blog@ gen /home/elen/Src/blog.nyaw.xyz
> node ./scripts/writeData.mjs && node ./scripts/writeSitemap.mjs && node ./scripts/time.mjs

Data successfully written to data.ts
Sitemap generated and saved as sitemap.xml
written time.json
vinxi v0.4.3
The CJS build of Vite's Node API is deprecated. See https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.


⚙  Building your app...
removing /home/elen/Src/blog.nyaw.xyz/.vinxi/build/ssr
vinxi v0.4.3
The CJS build of Vite's Node API is deprecated. See https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.


⚙ Building your router ssr...


📦 Compiling ssr router...
vinxi building router ssr in http mode
vite v5.4.9 building SSR bundle for production...
✓ 234 modules transformed.
x Build failed in 9.98s

[03:09:23]  ERROR  [unocss:global:build:scan] [postcss] /home/elen/Src/blog.nyaw.xyz/assets/ssr-!~{001}~.js-unocss-hash.css:1:26577: Unknown word

  at Input.error (node_modules/.pnpm/postcss@8.4.47/node_modules/postcss/lib/input.js:106:16)
[...]

PR #4212, opened 2024-10-18
feat(generator): skip comment

Simon-He95:
close: #4208 #4217


Simon-He95:
This may cause some people to add Unocss to the file as a comment in order to avoid using some rules in the `safelist` every time, but in theory, comments should be skipped and parsed. No one knows what strange rules the comment will have, which may cause more difficult to control problems, leading to a regular infinite loop or incorrect parsing. It is recommended to replace the comment with a placeholder before parsing, and then restore it after conversion. In addition, for the above non-standard usage, you may consider adding a comment rule, perhaps `// @uno-safe-list w-10px` 🤣

packages/shared-integration/src/constants.ts
```diff
@@ -5,4 +5,5 @@ export const CSS_PLACEHOLDER = '@unocss-placeholder'
 export const SKIP_START_COMMENT = '@unocss-skip-start'
 export const SKIP_END_COMMENT = '@unocss-skip-end'
 // eslint-disable-next-line regexp/no-super-linear-backtracking, regexp/no-useless-lazy, regexp/optimal-quantifier-concatenation
-export const SKIP_COMMENT_RE = new RegExp(`(\/\/\\s*?${SKIP_START_COMMENT}\\s*?|\\/\\*\\s*?${SKIP_START_COMMENT}\\s*?\\*\\/|<!--\\s*?${SKIP_START_COMMENT}\\s*?-->)[\\s\\S]*?(\/\/\\s*?${SKIP_END_COMMENT}\\s*?|\\/\\*\\s*?${SKIP_END_COMMENT}\\s*?\\*\\/|<!--\\s*?${SKIP_END_COMMENT}\\s*?-->)`, 'g')
+export const SKIP_UNO_COMMENT_RE = new RegExp(`(\/\/\\s*?${SKIP_START_COMMENT}\\s*?|\\/\\*\\s*?${SKIP_START_COMMENT}\\s*?\\*\\/|<!--\\s*?${SKIP_START_COMMENT}\\s*?-->)[\\s\\S]*?(\/\/\\s*?${SKIP_END_COMMENT}\\s*?|\\/\\*\\s*?${SKIP_END_COMMENT}\\s*?\\*\\/|<!--\\s*?${SKIP_END_COMMENT}\\s*?-->)`, 'g')
+export const SKIP_COMMENT_RE = /^\s*(<!--.*?-->|\/\/.*)/gm
```
antfu, line 9: This would create a lot of false-negative, RegExp is not turning complete - removing comments correctly would require a proper parser - where UnoCSS is designed to be language-agnosic (+parsing would cost a lot computational power as well as bundle size), we shouldn't do this.

In the end, Uno's approach is trying to sacrifice a bit false-positive (which is just extra necessary bits) to avoid false-negative (which would cause user apps to break unexpectedly). I believe this is the best workaround so far. For the most accuracy, I'd say users can provide custom extractors for each language and use a proper parser to extract - but that would be out of scope of Uno I'd say.

In that case I would suggest to not have this PR.

Simon-He95: It suddenly occurred to me that this implementation of skipping comments can be implemented by users themselves in transformers

PR #4235, opened 2024-10-31
docs: Update arbitrary-variants extractor example

jonathonadams:
The arbitrary variants extractor is not callable, it just needs to be imported

### DIFF
--- a/docs/extractors/arbitrary-variants.md
+++ b/docs/extractors/arbitrary-variants.md
@@ -34,7 +34,7 @@ import { defineConfig } from 'unocss'
 
 export default defineConfig({
   extractors: [
-    extractorArbitrary(),
+    extractorArbitrary,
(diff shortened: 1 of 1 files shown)


hannoeru:
Maybe we should align variants extractor with others by make it callable?
cc: @zyyv

zyyv:
> Maybe we should align variants extractor with others by make it callable? cc: @zyyv

I agree with you.

@jonathonadams Would you like to change it being a function?

jonathonadams:
Sure. I initially thought that would be the preferred option but for some reason decided against it.

jonathonadams:
@zyyv Given this is a breaking change, do you still want me to target the main branch for the PR?

jonathonadams:
Closing in favour of #4239

PR #4239, opened 2024-10-31
refactor(extactor-attributify)!: make arbitrary variants extractor callable

jonathonadams:
See #4235 for further details.

This PR aligns the arbitrary variant extractor with the others and makes it callable.

Note the docs already show it as callable, so they do not need to be updated.


hannoeru (approved):
LGTM, I think this will not effect normal user, as people usually using preset-mini as starter which enabled the extractorArbitraryVariants by default. But might effect some library authors, from GitHub search is not a lot (maybe 2 or 3), so I think is fine.

zyyv (approved):

antfu (approved):

PR #4252, opened 2024-11-06
feat(core, preset-mini): generate preflights on demand

### LINKED ISSUE #3521: Let `preflight-base` variables be generated on-demand and reusable.

The previous context in a discussion: <https://github.com/unocss/unocss/discussions/3513>.

Currently, UnoCSS will generate `preflight-base` variables as the default behavior with the `preset-mini`.

When building in the default mode `global`  without any utility, some `preflight-base` variables will always be generated to avoid potential errors in case of the usage of some utilities. However, the project doesn't always use some of these utilities. It may increase some extra computational and transferring costs.

In the modes such as `vue-scoped`, `shadow-dom`, and `dist-chunk`, these variables will be generated into each generated CSS file. It seems to be an unexpected behavior. Also in the mode `vue-scoped`, the selectors have a hash.

I haven't been using Svelte, so I don't know what the mode `svelte-scoped` behaves like.

### Suggested solution

The scopes may include:

- generate `preflight-base` variables on-demand when related utilities are in use
- clarify the modes in which these variables should be included in each generated CSS file and if they should be hashed in selectors
- how to reuse these variables in different CSS files

### Alternative

_No response_

### Additional context

_No response_

### Validations

zyyv:
resolve #3521 

~~In `Rulemeta`, specify the `preflight key` you need to depend on and generate preflights on demand. This will reduce irrelevant CSS output.~~


packages/core/src/types.ts
```diff
@@ -202,6 +209,11 @@ export interface RuleMeta {
    * @private
    */
   __hash?: string
+
+  /**
+   * Depend on preflight keys
+   */
+  preflightKeys?: Arrayable<string>
```
antfu, line 216: I am a bit worried that if this would be too coupled with our implementation of preset-mini (meaning it might not be that useful for other custom presets)

I wonder if we could put this in rule's meta, and instead of passing a `keys`, we pass a list of "Active rules"? So `preflightKeys` can be an internal convention we use for `preset-mini`, while it's also possible to pass more information other than just keys. WDYT?

zyyv: There is some coupling～😅, I will continue to explore new implementation methods.

docs/presets/mini.md
```diff
@@ -178,7 +178,11 @@ Prefix for CSS custom properties.
...
-- **Type:** `boolean`
+- **Type:** `boolean` | `on-demand`
 - **Default:** `true`
 
-Generate preflight.
+Generate preflight css. It can be:
+
+- `true`: always generate preflight.
+- `false`: no preflight.
+- `on-demand`: only generate preflight for used utilities.
```
zyyv, line 188: Add`on-demand` mode for `preset-mini` `preflight` option, which can generate the dependency css of theme preflight on demand.

packages/core/src/config.ts
```diff
@@ -171,13 +171,13 @@ export async function resolveConfig<Theme extends object = object>(
...
         prefixes.forEach((prefix) => {
-          rulesStaticMap[prefix + rule[0]] = [i, rule[1], rule[2], rule]
+          rulesStaticMap[prefix + rule[0]] = [i, rule[0], rule[1], rule[2], rule]
         })
         // delete static rules so we can't skip them in matching
         // but keep the order
         return undefined
       }
-      return [i, ...rule]
+      return [i, rule[0], rule[1], rule[2], rule]
```
antfu, line 180: I wonder if we can just return `[i, rule]`, or even the `i` is not needed?

antfu: I made a PR to this branch with https://github.com/unocss/unocss/pull/4304

Seems to work fine?

antfu (approved):

PR #4268, opened 2024-11-20
feat(core)!: deprecate `new UnoGenerator`, make `createGenerator()` async

antfu:
This is a breaking change that makes the creation of UnoGenerator async. This is a preparatory work for the future to support async presets and async configs resolution.

As class constructor does not support async, `new UnoGenerator` is deprecated in favor of the factory function `createGenerator()`.

⚠️ Breaking: `createGenerator()` now returns `Promise<UnoGenerator>` instead of `UnoGenerator`

This change should only affect programmatically usages, normal users with integrations should not be affected by this.

References:

- Projects that use `new UnoGenerator`: https://github.com/search?q=%22new+UnoGenerator%22++NOT+is%3Afork&type=code

### DIFF
--- a/docs/tools/core.md
+++ b/docs/tools/core.md
@@ -26,7 +26,7 @@ The core engine of UnoCSS without any presets: `@unocss/core`. It can be used as
 ```ts
 import { createGenerator } from '@unocss/core'
 
-const generator = createGenerator(
+const generator = await createGenerator(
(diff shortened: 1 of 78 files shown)


zyyv (approved):

hannoeru (approved):
LGTM

Issue #4269, opened 2024-11-20
Postcss @apply changes rule order


ameinhardt:
### UnoCSS version

0.64.1

### Describe the bug

The order of resolved declarations after postprocessing rules with @apply has changed.

### Reproduction

- ✅ [Playground without postcss and expected result](https://uno.antfu.me/play/#html=DwEwlgbgBAxgNgQwM5ILwCIAuBTJmoAWAtAGYCuccUJc2AHlGDgLZJEzYB2OATlAFZk8YEgE92XXugB8AFXqZgAenARpAKCA&config=Q&css=HQFwpgziAEDeBQ1oAECGAHdAbAntARgOYC0ATmACbECsADLQNyIoQCumA9qSBNABQBjDli4AuaEJGliAWwCWADz5yAdtA4BrLKnwAaAtoEb9%2BQxoCU5uMyT5URwqQ6sVVSWOiOwYFUyQBfeH8gA&options=N4IgzgLgTglgxhEAuaBXApgGhNAhgOzADMB7KAW2SNwBswscoDizyBhVSE9gZR6tr1seQqQoAJCORoC66AL5A)
- ⚡ [Reproduction with postcss and unexpected result](https://stackblitz.com/edit/ameinhardt-uno-post-daisy-v8dnj6?file=package.json)

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4270, opened 2024-11-20
fix(postcss): keep rule order

ameinhardt:
Addresses case of #4269. Not sure if more areas change the order


packages/postcss/src/apply.ts
```diff
@@ -82,7 +82,7 @@ export async function parseApply(root: Root, uno: UnoGenerator, directiveName: s
           css.walkDecls((declaration) => {
             declaration.source = source
           })
-          rule.parent.append(css)
+          rule.parent.insertAfter(rule, css)
```
hannoeru, line 85: Can we add some test for this one?

zyyv (approved):

hannoeru (approved):
LGTM

Issue #2531, opened 2023-04-21
当我同时设置autoInstall=true,以及自定义图标时，它总是尝试去帮我安装@iconify-json/xx


xiankaiqun:
### UnoCSS version

0.51.5

### Describe the bug

虽然会尝试去安装，但它安装失败后 ，自定义图标是可以正常显示的

### Reproduction

```typescript
import transformerAttributifyJsx from '@unocss/transformer-attributify-jsx';
import { defineConfig, presetAttributify, presetIcons, presetTypography, presetUno } from 'unocss';

import jimei from './iconfont/jimei.json';

export default defineConfig({
  presets: [
    presetUno(),
    presetAttributify({
      prefix: 'un-',
      prefixedOnly: true,
    }),
    presetIcons({
      autoInstall: true,
      collections: {
        jm: (iconName) => {
          return jimei.find((e) => e.font_class === iconName)?.svg;
        },
      },
    }),
    presetTypography(),
  ],

  transformers: [transformerAttributifyJsx()],
});
```
![image](https://user-images.githubusercontent.com/26294948/233624057-5ab70403-cffa-40cf-94d8-0699d146efdb.png)

### System Info

window11
Microsoft Edge 版本 112.0.1722.48 (正式版本) (64 位)
node: v18.14.2
vite: 4.3.1

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

xiankaiqun:
这是我的json文件
我的vite.config.ts
```json
[
 {
    "icon_id": "35113287",
    "name": "celiang2",
    "font_class": "celiang2",
    "unicode": "e6df",
    "unicode_decimal": 59103,
[...]

PR #4278, opened 2024-11-21
fix(preset-icons): validate collection name before load icon svg data

ntnyq:
### Description

This PR checks if the collection name is valid before load icon SVG data.

### Context:

With class `i-foo-bar-baz`, presetIcons will try to install icon data order by order in bellow and got 3 errors.

1. `@iconify-json/foo-bar-baz`
2. `@iconify-json/foo-bar`
3. `@iconify-json/foo`

### Related issue

#2531

### DIFF
--- a/packages/preset-icons/src/core.ts
+++ b/packages/preset-icons/src/core.ts
@@ -201,16 +201,30 @@ export async function parseIconWithLoader(body: string, loader: UniversalIconLoa
   let name = ''
   let svg: string | undefined
 
+  const allCollections = [
+    ...icons,
+    ...Object.keys(options.customCollections || {}),
+  ]
+
   if (body.includes(':')) {
     [collection, name] = body.split(':')
+
+    if (!allCollections.includes(collection))
+      return
+
     svg = await loader(collection, name, options)
   }
   else {
     const parts = body.split(/-/g)
     for (let i = COLLECTION_NAME_PARTS_MAX; i >= 1; i--) {
       collection = parts.slice(0, i).join('-')
+
+      if (!allCollections.includes(collection))
+        continue
+
       name = parts.slice(i).join('-')
       svg = await loader(collection, name, options)
+
       if (svg)
         break
     }


zyyv:
Great, I think we can sense in advance whether it comes from a custom icon collection.

zyyv (approved):

onmax:
I use a custom icon set. In order to auto install it in UnoCSS, I used to install my icon set like this:

```
"@iconify-json/nimiq": "https://pkg.pr.new/onmax/nimiq-ui/nimiq-icons@f27621c",
```

This no longer works. Any idea how could I do it now? 

Thanks!

userquin:
Check if the pkg-pr-new has been removed, if the pr including the collection has been merge the link will not work: if so, replace pkg-pr-new version with the corresponding version

onmax:
Nope. I can see the icons.json being installed in node_modules. That's not the issue

userquin:
can you check if the icon is in the css? maybe it is a problem with the icon size

onmax:
No. The CSS is not being generated. The VSCode extension also does not detect the utility `i-nimiq:bell`

image

userquin:
show me the div with the icon not the parent

onmax:
Yes, sorry. Here is the `div` with the icon

image


And yes, `nimiq` is not part of `@iconify-json`, but in order to install it automatically in UnoCSS, I renamed it in `package.json` as follows:

```
    "@iconify-json/nimiq": "https://pkg.pr.new/onmax/nimiq-ui/nimiq-icons@24e0317",
```

You can see the source code [here](https://github.com/nimiq/developer-center/blob/main/package.json#L45). With `unocss@0.64.0` this technique works,

This allowed me to install it automatically without adding any extra config to `uno.config.ts`, which was quite handy since I use this method in more than 10 repos...

As an alternative, could I open a feature request to add a config param in `presetIcons` to allow this or any other alternative method to load custom icons without writing a custom function?

Thanks in advance!

userquin:
@onmax can you try with an icon without `hidden: true`? 

Where are the styles in the devtools?

imagen

onmax:
I'm not sure why `"hidden": true` is there, as I don't need it. However, it doesn't seem to be the problem:

image

- [`flags-it-hexagon`](https://github.com/onmax/nimiq-ui/blob/main/packages/nimiq-icons/dist/icons.json#L1018): Does not have `hidden: true` and is not displayed (same for `i-nimiq:flags-vo-hexagon`)
- [`bell`](https://github.com/onmax/nimiq-ui/blob/main/packages/nimiq-icons/dist/icons.json#L142): Does have `hidden: true`.

I also checked older commits, for example [8 months ago](https://github.com/onmax/nimiq-ui/blob/0db62e3def94b54cb898022870174ec6ddf2de13/packages/nimiq-icons/dist/icons.json). If I change the `unocss` version to `0.64.0` works.

I also have an icon explorer which uses `iconify` vue component and shows the latest icons: 
https://www.nimiq.com/developers/build/ui/design/icons

userquin:
the cons are shown because you're rendering the svg, unocss sets the icon in the css.

Can you show me the css props of the icon? Should be something like this:

imagen

onmax:
That's the problem. The icon is not being recognized automatically in `0.65.1`. 

Here are the different cases

| UnoCSS Version | Screenshot |
|--------|--------|--------|
| 0.65.1 | image |
| 0.65.1 (implicit install) | image  |
| 0.64.0 | image |

Hope this helps to show the issue.

PR #4283, opened 2024-11-22
Add space to fix ::marker pseudo class

chuckntaylor:
I see that using the marker pseudo class stopped working between versions 0.63 and 0.64. The difference I see is the lack of a space before `::marker` in the resulting css.

### DIFF
--- a/packages/preset-mini/src/_variants/pseudo.ts
+++ b/packages/preset-mini/src/_variants/pseudo.ts
@@ -70,7 +70,7 @@ const PseudoClasses: Record<string, string> = Object.fromEntries([
   ['before', '::before'],
   ['after', '::after'],
   ['selection', ' ::selection'],
-  ['marker', '::marker'],
+  ['marker', ' ::marker'],
   ['file', '::file-selector-button'],
 ].map(key => Array.isArray(key) ? key : [key, `:${key}`]))
 


zyyv:
This error was caused by https://github.com/unocss/unocss/pull/4221.

Thanks for finding this error, but the normal behavior should not be just the pseudo-element with marker in children

The same error also occurs with the `::selection` pseudo-element. I will send a PR to fix it together

Right generated css: 

```css
.marker\:bg-red *::marker {
  --un-bg-opacity: 1;
  background-color: rgb(248 113 113 / var(--un-bg-opacity));
}
.marker\:bg-red::marker {
  --un-bg-opacity: 1;
  background-color: rgb(248 113 113 / var(--un-bg-opacity));
}

.selection\:bg-red *::selection {
  --un-bg-opacity: 1;
  background-color: rgb(248 113 113 / var(--un-bg-opacity));
}
.selection\:bg-red::selection {
  --un-bg-opacity: 1;
  background-color: rgb(248 113 113 / var(--un-bg-opacity));
}
```

zyyv:
Let me mark this as a draft first, since this PR doesn't solve a real problem.
Or you can update more resolve ways about this PR.

antfu:
After https://github.com/unocss/unocss/pull/4291 get merged, it should allow this to be implemented a lot easier

PR #4298, opened 2024-11-28
fix(preset-mini): spread mutiple pseudo-elements for correct variant handling

zyyv:
close #4283
Before this, the error handling method was related to #4221 #3771

### DIFF
--- a/packages/preset-mini/src/_variants/pseudo.ts
+++ b/packages/preset-mini/src/_variants/pseudo.ts
@@ -69,8 +69,6 @@ const PseudoClasses: Record<string, string> = Object.fromEntries([
   ['placeholder', '::placeholder'],
   ['before', '::before'],
   ['after', '::after'],
-  ['selection', ' ::selection'],
-  ['marker', '::marker'],
   ['file', '::file-selector-button'],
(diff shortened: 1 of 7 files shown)


packages/preset-mini/src/_variants/pseudo.ts
```diff
@@ -89,6 +87,11 @@ const PseudoClassFunctions = [
   'has',
 ]
 
+const PseudoClassesMulti: Record<string, string[]> = Object.fromEntries([
```
zyyv, line 90: Because it is different from other pseudos that require `multiPass`, it is classified as a new category.

packages/preset-mini/src/_variants/pseudo.ts
```diff
@@ -89,6 +87,11 @@ const PseudoClassFunctions = [
...
+  ['selection', ['::selection', ' *::selection']],
+  ['marker', ['::marker', ' *::marker']],
+])
```
zyyv, line 93: Because it is different from other pseudos that require `multiPass`, it is classified as a new category.

antfu (approved):

PR #4299, opened 2024-11-28
feat(core): output layer order css

zyyv:


packages/core/test/__snapshots__/use-css-layer.test.ts.snap
```diff
@@ -1,20 +1,22 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`use-css-layer > change layer name 1`] = `
-"/* layer: shortcuts */
+"@layer utilities.shortcuts, utilities;
```
antfu, line 4: This should PR be a feat I suppose

antfu (approved):

Issue #4297, opened 2024-11-28
[rspack] hmr not work


thinke5:
### UnoCSS version

^0.64.1

### Describe the bug

unplugin in respack `plugin.__vfsModules` not  includes `plugin.__virtualModulePrefix`

This line of code will have a bug

https://github.com/unocss/unocss/blob/01994e57508c0bfa53ebff992210991032a408e4/packages/webpack/src/unplugin.ts#L160

After changing to `String.replace` , proceed to the next process normally
```js
        let path = decodeURIComponent(id.replace(plugin.__virtualModulePrefix,''));
```

### Reproduction

not

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4300, opened 2024-11-29
fix(rspack): optimize unplugin rspack plugin.__vfsModules id not includes __virtualModulePrefix

thinke5:
fix(rspack): optimize unplugin rspack plugin.__vfsModules id not includes __virtualModulePrefix #4297

### DIFF
--- a/packages/webpack/src/unplugin.ts
+++ b/packages/webpack/src/unplugin.ts
@@ -157,7 +157,7 @@ export function unplugin<Theme extends object>(configOrPath?: WebpackPluginOptio
       lastTokenSize = tokens.size
       Array.from(plugin.__vfsModules)
         .forEach((id) => {
-          let path = decodeURIComponent(id.replace(plugin.__virtualModulePrefix, ''))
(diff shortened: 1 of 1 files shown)


packages/webpack/src/unplugin.ts
```diff
@@ -157,7 +157,7 @@ export function unplugin<Theme extends object>(configOrPath?: WebpackPluginOptio
       lastTokenSize = tokens.size
       Array.from(plugin.__vfsModules)
         .forEach((id) => {
-          let path = decodeURIComponent(id.slice(plugin.__virtualModulePrefix.length))
+          let path = decodeURIComponent(id.replace(plugin.__virtualModulePrefix, ''))
```
antfu, line 160: I wonder if we should do `id.startsWith(plugin.__virtualModulePrefix) ? id.slice(plugin.__virtualModulePrefix.length) : id`

thinke5: Yes, it is better. I updated

antfu (approved):

Issue #4303, opened 2024-11-29
Custom sorting for multi-selector rules


ameinhardt:
### Clear and concise description of the problem

For making daisyui components work efficiently with unocss, I'd like to create related rules in a specific order.
My first attempt of having many dynamic rules for one selector = component doesn't seem to work, as unocss just seems to evaluate the first match for a token.
The second attempt of having multi-selector rules for a component doesn't fully work, as it re-groups css rules. Sort seem to only work on rule level, not on multi-selector level.

### Suggested solution

Introducing another special symbol `selectors.sort`, applying to each entries meta in `UnoGenerator.parseUtil()`, for being able to define sort order of multi-selector rules within their layer

### Alternative

Apply all matching dynamic rules for one token

### Additional context

[...]

PR #4308, opened 2024-12-01
feat: allow fine-grained sorting order on multi-selector rules

ameinhardt:
Addresses #4303


antfu (approved):

packages/core/src/generator/index.ts
```diff
@@ -659,6 +660,7 @@ class UnoGeneratorInternal<Theme extends object = object> {
 
             // Extract variants from special symbols
             let variants = variantHandlers
+            let entryMeta = { ...meta }
```
antfu, line 663:
```suggestion
            let entryMeta = meta
```

I think we can avoid copy when not necessary

antfu (approved):

PR #4319, opened 2024-12-08
feat(preset-icons,transformer-directives): allow add safe icons collections

userquin:
Changes included in https://github.com/unocss/unocss/pull/4278 will prevent using/auto-installing iconify collections not listed in default collections. In that case we need to use `createExternalPackageIconLoader` from `@iconify/utils/lib/loader/external-pkg` or node `fs` loader, check https://github.com/unocss/unocss/discussions/4318#discussioncomment-11499122.

This PR adds `safeCollectionsNames` option to allow use external collections including `auto-install` support: I'll add some hint in the jsdocs since this will work only with `@iconify-json/*` collections (✔️ done).

When there are new `@iconify-json/*` collections missing in the `collections.json` file, the consumer can use this new option until we release a new version with the new iconify collections included in the json file (transient solution: instead using static import + custom icon loader the consumer can just add the new collection name to the new option).

Tested this PR on my local using https://github.com/nimiq/developer-center/tree/main (after updating UnoCSS to 0.65.1, using 0.64.0 just works without this PR).

@zyyv Feel free to close this PR since there is an option to use external packages, the user using some hack to move the `nimiq` collection under `@iconify-json/nimiq` package name (`pnpm` is adding the package inside `@iconify-json/nimiq` folder).

### DIFF
--- a/packages/preset-icons/src/types.ts
+++ b/packages/preset-icons/src/types.ts
@@ -56,7 +56,16 @@ export interface IconsOptions {
   warn?: boolean
 
   /**
-   * Safe collections to use (will be also auto installed when missing and `autoInstall` enabled).
+   * Safe `@iconify-json` collections to use (will be also auto installed when missing and `autoInstall` enabled).
+   *
+   * This should be used only when default there are new collections not listed in the default icons preset collections.
+   *
+   * Adding external collections will not work, you should use `FileSystemIconLoader` from
+   * `@iconify/utils/lib/loader/fs` or `createExternalPackageIconLoader` from
+   * `@iconify/utils/lib/loader/external-pkg` instead.
+   *
+   * @see https://unocss.dev/presets/icons#filesystemiconloader
+   * @see https://unocss.dev/presets/icons#externalpackageiconloader
    */
   safeCollectionsNames?: string[]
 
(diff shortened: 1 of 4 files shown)


zyyv (approved):
Thanks, it is compatible with collections that are not synced to `iconify`.

packages/preset-icons/src/types.ts
```diff
@@ -55,6 +55,20 @@ export interface IconsOptions {
...
+   * This options should be used only when there are new `@iconify-json` collections not listed in the default icons preset collection names.
+   *
+   * Adding external collections will not work, you should use `FileSystemIconLoader` from
+   * `@iconify/utils/lib/loader/fs` or `createExternalPackageIconLoader` from
+   * `@iconify/utils/lib/loader/external-pkg` instead.
+   *
+   * @see https://unocss.dev/presets/icons#filesystemiconloader
+   * @see https://unocss.dev/presets/icons#externalpackageiconloader
+   */
+  safeCollectionsNames?: string[]
```
antfu, line 70:
```suggestion
  iconifyCollectionsNames?: string[]
```

In this context, I think `safe` is a bit ambiguous and confusing.

userquin: looks like npm registry is down

userquin: I'm going to add the new option to the docs

PR #3380, opened 2023-11-26
feat!: move to ESM-only and fix package exports

kwaa:
## Description

`unbuild` outputs `.d.mts` and `.d.cts` at build time, but `unocss` doesn't use it.

This PR updates the export map for most packages to export types correctly.

## References

https://publint.dev/unocss@0.57.7
https://arethetypeswrong.github.io/?p=unocss
https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md
https://github.com/microsoft/TypeScript/issues/50466
https://github.com/unjs/unbuild/issues/238

PR #4322, opened 2024-12-10
test: add additional attw test for built packages

userquin:
This PR includes:
- adds tests for subpackage exports types using [arethetypeswrong/cli](https://github.com/arethetypeswrong/arethetypeswrong.github.io/tree/main/packages/cli)
- fixes `@unocss/webpack/rspack`, broken node10 types (https://arethetypeswrong.github.io/?p=%40unocss%2Fwebpack%400.65.1)
- patches `arethetypeswrong/cli` to use `pnpm pack` instead `npm pack`: cannot run the script on my local Windows (check https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/230)
- adds a new test in ci `Check types` to run `nr test:attw`

This is the continuation of moving UnoCSS to ESM only, maybe we can remove the protection in the scripts: https://github.com/unocss/unocss/pull/3380#issuecomment-1898378339

(This PR is inspired by the work done by Daniel at Nuxt here https://github.com/nuxt/nuxt/pull/30206)

/cc @kwaa


antfu (approved):

kwaa (approved):

package.json
```diff
@@ -24,11 +24,13 @@
...
     "test:update": "vitest -u",
     "test:integration": "pnpm -F svelte-scoped test:integration",
-    "test:ci": "nr build && nr typecheck && nr lint && nr test && nr test:integration"
+    "test:ci": "nr build && nr typecheck && nr lint && nr test && nr test:integration",
+    "test:attw": "pnpm --filter=./packages/* --filter=!cli --filter=!core --filter=!reset --filter=!shared-* test:attw"
   },
   "devDependencies": {
     "@antfu/eslint-config": "catalog:",
     "@antfu/ni": "catalog:",
+    "@arethetypeswrong/cli": "0.17.1",
```
userquin, line 33: change this to use `catalog:`

package.json
```diff
@@ -24,11 +24,13 @@
     "test": "vitest",
     "test:update": "vitest -u",
     "test:integration": "pnpm -F svelte-scoped test:integration",
-    "test:ci": "nr build && nr typecheck && nr lint && nr test && nr test:integration"
+    "test:ci": "nr build && nr typecheck && nr lint && nr test && nr test:integration",
+    "test:attw": "pnpm --filter=./packages/* --filter=!cli --filter=!core --filter=!reset --filter=!shared-* test:attw"
```
userquin, line 28: not sure why I excluded core, it should be there

zyyv (approved):
@userquin Feel free to merge PR when you feel good. 😃

zyyv:
@userquin Would you like to modify the cli command so that each package uses a common attw.json?
Because they are all avoiding `cjs` checks.

```bash
--config-path <path>                    Path to config file (default: ./.attw.json)
```

userquin:
Nuxt package uses false-esm ( cjs masquearading), and some package don't require the file (those with dual cjs/esm: postcss, eslint-*...). 

I guess it is better to have its own .attw.json file instead changing all the scripts.

I need to check if we can add excluding rules per subpackage export, for example in unocss package.

userquin:
I Will change all esm only packages to use only 1 attw json file.

zyyv (approved):

PR #4323, opened 2024-12-11
feat!: upgrade to `unplugin@2`

jardenliu:
try unocss/webpack `UnoCSSRspackPlugin` in rspack with unocss

 plugin.__vfs is undefined 

unplugin rspack fix it above 2.0.0 https://github.com/unjs/unplugin/commit/3c4b79e6f935d85befc2805bbce321e596e80d28


antfu:
We would need to do a minor version for this

antfu:
Thank you but it seems that we already upgrade to v2.

PR #4325, opened 2024-12-13
fix(shared-common): getMatchedPositions excludeRegex

Simon-He95:
excludeRegex will interfere with includeRanges. I think excludeRanges should be eliminated first, and then includeRegex should be matched.

like this case：

- before modification
image

- after modification
image


antfu:
Could we have some test to free the behavior? thanks

Simon-He95:
> Could we have some test to free the behavior? thanks

sure

zyyv (approved):

antfu (approved):

PR #4347, opened 2024-12-21
chore: remove `attw` patch

userquin:
This PR removes the `attw` patch included in this PR #4322 and updates the version.

This PR also updates the ci workflow to run `test:attw` script only on `Linux` and `node lts`, the same env. used to release the packages.

I was using old `npm` cli 10.2.3 version, updating to latest 10 version (`npm install npm@^10 -g`) `npm pack` works on my Windows.


userquin:
Looks like old Node 18.18.2 using old npm cli, we still need the patch or we can skip the `test:attw` on it, I'll check if we can exclude it.

PR #4350, opened 2024-12-24
refactor(transformDirectives): handleScreen

Simon-He95:


antfu:
Could you explain a bit what's this change for and why it's necessary?

Simon-He95:
> Could you explain a bit what's this change for and why it's necessary?

Because this function handles screen separately, I hope to judge in advance, rather than entering this function to judge whether it is screen. I think it should make the code more readable

antfu (approved):

PR #4361, opened 2024-12-30
feat(inspector): show url on Vite's dev server start

ferferga:
Recently, I learned that UnoCSS has an inspector. However, it's not "advertised" anywhere and I think that it could be showcased to users way better

Inspired by Vue Devtools plugin, a message is now shown to the user when spinning up Vite's development server:

image


packages/inspector/package.json
```diff
@@ -45,6 +45,7 @@
     "@unocss/core": "workspace:*",
     "@unocss/rule-utils": "workspace:*",
     "gzip-size": "catalog:",
+    "kolorist": "^1.8.0",
```
zyyv, line 48: We have `colorette` dependency in pnpm workspace, you can use it with `catalog` rather than install a new dependency.
https://github.com/unocss/unocss/blob/d32841ce9e1ada8b1fd81d980daba5d2c6cf9ab1/pnpm-workspace.yaml#L75

zyyv (approved):

packages/inspector/src/index.ts
```diff
@@ -105,6 +107,19 @@ export default function UnocssInspector(ctx: UnocssPluginContext): Plugin {
...
+    const _printUrls = server.printUrls
+    const colorUrl = (url: string) =>
+      cyan(url.replace(/:(\d+)\//, (_, port) => `:${bold(port)}/`))
+
+    server.printUrls = () => {
+      _printUrls()
+      for (const url of server.resolvedUrls?.local ?? []) {
+        const inspectorUrl = url.endsWith('/') ? `${url}${baseUrl}/` : `${url}/${baseUrl}/`
+        // eslint-disable-next-line no-console
+        console.log(`  ${green('➜')}  ${bold('UnoCSS Inspector')}: ${green(`Open ${colorUrl(`${inspectorUrl}`)} as a separate window`)}`)
```
antfu, line 120:
```suggestion
        console.log(`  ${green('➜')}  ${bold('UnoCSS Inspector')}: ${colorUrl(`${inspectorUrl}`)}`)
```

packages/inspector/package.json
ferferga: @zyyv Great catch, thank you!

PR #4362, opened 2025-01-03
feat(vite): add option `checkImport`

ntnyq:
This PR adds an option `checkImport` to vite plugin.

## why

Framework like [WXT](https://github.com/wxt-dev/wxt) supports dynamic entrypoints and only partial of them need style.

Using UnoCSS as a vite plugin will get the warning every time run dev and build.

It's kind of annoying.

### DIFF
--- a/packages/vite/src/modes/global/build.ts
+++ b/packages/vite/src/modes/global/build.ts
@@ -281,7 +281,7 @@ export function GlobalModeBuildPlugin(ctx: UnocssPluginContext<VitePluginConfig>
           if (replaced)
             return
 
-          if (!(await getConfig() as VitePluginConfig).disableEntryCheck) {
+          if ((await getConfig() as VitePluginConfig).checkImport) {
             this.warn(MESSAGE_UNOCSS_ENTRY_NOT_FOUND)
           }
 
--- a/packages/vite/src/modes/global/build.ts
+++ b/packages/vite/src/modes/global/build.ts
@@ -19,6 +19,7 @@ import {
   resolveId,
   resolveLayer,
 } from '../../integration'
+import { MESSAGE_UNOCSS_ENTRY_NOT_FOUND } from './shared'
 
 // https://github.com/vitejs/vite/blob/main/packages/plugin-legacy/src/index.ts#L742-L744
 function isLegacyChunk(chunk: RenderedChunk, options: NormalizedOutputOptions) {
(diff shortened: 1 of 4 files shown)


zyyv (approved):
👀

packages/vite/src/types.ts
```diff
@@ -57,4 +57,11 @@ export interface VitePluginConfig<Theme extends object = object> extends UserCon
    * @default 'cors'
    */
   fetchMode?: 'cors' | 'navigate' | 'no-cors' | 'same-origin'
+
+  /**
+   * Disable `import 'uno.css'` existing check.
+   *
+   * @default false
+   */
+  disableEntryCheck?: boolean
```
hannoeru, line 66:
```suggestion
  checkImport?: boolean
```

Maybe use `checkImport` and default to `true`? that give us more flexibility for the feature that we can add different check modes for example.

packages/vite/src/modes/global/build.ts
```diff
@@ -279,8 +279,12 @@ export function GlobalModeBuildPlugin(ctx: UnocssPluginContext<VitePluginConfig>
           // to replace on current build pipeline, we can skip the warning.
           if (replaced)
             return
-          const msg = '[unocss] Entry module not found. Did you add `import \'uno.css\'` in your main entry?'
-          this.warn(msg)
+
+          if (!(await getConfig() as VitePluginConfig).disableEntryCheck) {
+            const msg = '[unocss] Entry module not found. Did you add `import \'uno.css\'` in your main entry?'
+            this.warn(msg)
```
hannoeru, line 285:
```suggestion
            this.warn('[unocss] Entry module not found. Did you add `import \'uno.css\'` in your main entry?')
```

packages/vite/src/types.ts
ntnyq: Consider future usage, `checkImport?: boolean | { entry?: boolean }` would be better. 
But for now, since there isn't any other import check. It's may be over designed.

packages/vite/src/modes/global/build.ts
ntnyq: Message has been extracted to a shared constant as it's been used three times within two files.

packages/vite/src/types.ts
hannoeru: I think just rename it will be fine for now.

antfu (approved):

zyyv (approved):

PR #4371, opened 2025-01-09
chore: sync reset css

zhangmo8:
sync reset css for tailwind v4


antfu:
We shouldn't do it, as the presets are not ready for Tailwind 4. @zyyv is working on the Tailwind 4 preset, depends on the implementation we may not even need the reset file - but either way, we need to keep supporting Tailwind 3 and 4 at the same time, so even we do, we will need to have it in another file.

PR #4387, opened 2025-01-15
fix(preset-mini): update `inline/block` screen unit

zyyv:
fixed #4385


hannoeru (approved):
Might consider be a breaking change.

antfu (approved):

PR #4392, opened 2025-01-16
feat(preset-mini): allow using multiple selectors for dark mode

XiNiHa:
This was found while porting [Tailwind config of `solid-ui`](https://github.com/stefan-karger/solid-ui/blob/12a0685a627030264e70ca7d4bc7cd1c5d187e2d/packages/cli/src/utils/templates.ts#L19) to UnoCSS.


packages/rule-utils/src/variants.ts
```diff
@@ -12,13 +12,17 @@ export function variantMatcher(name: string, handler: (input: VariantHandlerCont
...
+          matcher,
           handle: (input, next) => next({
             ...input,
             ...handler(input),
           }),
-        }
+        }))
+        return handlers.length === 1
+          ? handlers[0]
+          : handlers
```
XiNiHa, line 25: For backward compat reasons, this ensures that changes from this PR don't affect the function's output if the input doesn't change. Always returning arrays broke one of our tests so it looks like it could also happen on user code.

antfu (approved):

zyyv:
Can you add some test for it?

XiNiHa:
> Can you add some test for it?

Done!

XiNiHa:
No idea what's happening inside the CIs... 😇

XiNiHa:
😇 https://github.com/nodejs/corepack/issues/612

PR #4397, opened 2025-01-19
fix: cjs package exports

userquin:
**DON'T MERGE YET**

This PR adds a modified Rollup plugin from this PR https://github.com/unjs/unbuild/pull/475 pinning `unbuild` to `v3.1.0.

This PR should fix the `test:attw` script, right now it is broken, fixing all CJS default package exports (d.ts and d.cts files).

Once this PR ready, we should:
- remove scripts/postbuild.ts` script
- remove root `xbuild` script from this PR


userquin:
Looks like pkg-pr-new is not working:

imagen

package.json
```diff
@@ -8,7 +8,8 @@
     "prepare": "esno scripts/prepare.ts",
     "taze": "taze minor -wIr && pnpm -r --parallel run update-post",
     "bench": "npm -C bench run bench",
-    "build": "rimraf packages/*/dist --glob && esno scripts/copy-files.ts && pnpm -r --filter=./packages-*/* run build && pnpm -r run build-post && esno scripts/postbuild.ts",
+    "build": "rimraf packages/*/dist --glob && esno scripts/copy-files.ts && pnpm -r --filter=./packages-*/* run build && pnpm -r run build-post",
```
hannoeru, line 11: We already have clean options set to true on unbuild config, is there any reason to add rimraf to build script?

pnpm-workspace.yaml
```diff
@@ -137,7 +137,7 @@ catalog:
   tsup: ^8.3.5
   tsx: ^4.19.2
   typescript: ^5.7.3
-  unbuild: ^3.3.1
+  unbuild: 3.1.0
```
antfu, line 140: We do need to pin it? Or should we fix upstream directly?

userquin: The PR (https://github.com/unjs/unbuild/pull/475) is there, I only need to push my changes to include `export { default } from '<identifier>'` fix.

I will try to remove the cjs plugin included with version 3.3.1 in the afternoon:
- https://github.com/unjs/unbuild/blob/main/src/builders/rollup/build.ts#L85
- https://github.com/unjs/unbuild/blob/main/src/builders/rollup/plugins/cjs.ts#L18

antfu: Oh, awesome! How about we do a patch from your PR with the latest version?

userquin: In the meantime, I'm going to submit another pr to fix cjs exports temporarily.

userquin: @antfu @hannoeru new PR here https://github.com/unocss/unocss/pull/4401

userquin:
superseded by #4401

PR #4407, opened 2025-01-26
build: change `catalog:` entries before releasing `pkg-pr-new`

userquin:
@jacob-8 @Aslemammad


antfu:
I would wish pkg-pr-new handles it automatically instead of us patching it

userquin:
> I would wish pkg-pr-new handles it automatically instead of us patching it

I'm talking with @Aslemammad, this is a temporary patch here... maybe we can add some utility since we need to provide the packages globs to pkg-pr-new to patch the packages before publishing

antfu:
Would running `pnpm publish` instead of `npm publish` in `pkg-pr-new` solve this? `pnpm publish` automatically resolves the catalog and workspace tag

userquin:
The problem is about ignoring the package.json file when running `pnpm pack`, it is not being used and pkg-pr.new should merge the package.json dependencies with the generated by the pack command.

userquin:
https://github.com/stackblitz-labs/pkg.pr.new/pull/318

PR #4427, opened 2025-02-11
docs: introduce unocss community

zyyv:
https://github.com/unocss/unocss/discussions/4426


docs/presets/community.md
```diff
@@ -1,54 +0,0 @@
-# Community presets
```
antfu, line 1: I think maybe we could keep this page, but replace the content to link the community

Issue #4432, opened 2025-02-16
Missing documentation: accessibility / a11y


GauBen:
### UnoCSS version

all

### Describe the bug

Hi dear maintainers!

Not really a bug _per se_, but there's no template for documentation issues, so I file it as a bug.

Neither https://unocss.dev/presets/icons nor https://antfu.me/posts/icons-in-pure-css document how to make CSS icons accessible, which is important when not supported by a text node.

The https://unocss.dev/presets/icons page should have an Accessibility category, presumably above Credits, teaching developers how to make their icons accessible.

I'm opening this issue to make it easier for you to track the documentation task. If it does not, feel free to close it :)

This task should also update the playground with best practices, as it's not the case today:

<img width="390" alt="Image" src="https://github.com/user-attachments/assets/ca10e523-2446-44e0-9441-b979ea309fa6" />

### Reproduction

- Go to https://unocss.dev/presets/icons
- Look for best practices

### System Info

:)

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4433, opened 2025-02-16
docs: icon accessibility

GauBen:
First hit at #4432!


GauBen:
Is adding a text node directly inside an icon ok? Something like `<span class="i-ph:github-logo-duotone">My GitHub Profile</span>`

zyyv (approved):

zyyv:
> Is adding a text node directly inside an icon ok? Something like `<span class="i-ph:github-logo-duotone">My GitHub Profile</span>`

Icons are usually applied to `block` elements.

antfu (approved):

PR #4435, opened 2025-02-18
feat(core)!: switch to nested css output

zyyv:
This pr will change the following:

- change css to `nested` css for output
- according to writing habits, parse variants in order to obtain a clear hierarchy

Advantages:
- clearer hierarchy

Disadvantages:
- compared with before, the same parent structure will not be reused, which will increase the size of CSS.

Consideration:
- For `presetWind3`, add the nested attribute switch to support browser compatibility Polyfill
- For `presetWind4`, always keep it enabled by default


packages-engine/core/src/generator.ts
```diff
@@ -496,7 +497,7 @@ class UnoGeneratorInternal<Theme extends object = object> {
   ): UtilObject {
     const handler = variantHandlers.slice()
       .sort((a, b) => (a.order || 0) - (b.order || 0))
-      .reduceRight(
+      .reduce(
```
zyyv, line 500: Change the order in which variants are parsed will be a breaking change.

antfu (commented):
Could we support both modes in the engine, and make it configurable on the variant side? We should keep tailwind 3 as-is (as well as user variants using the same approach), while make it possible to do nested in tailwind 4 presets

PR #4451, opened 2025-02-26
ci: replace node version `lts` to `22.14.0` temporarily

zyyv:
https://github.com/pnpm/pnpm/issues/9029#issuecomment-2650658230

After my investigation, I found that `actions/setup-node@v4` did not update the node `lts` version in time, which caused the `corepack` version to be unable to be upgraded to `0.31.0`, resulting in a ci error.

Temporarily fix node to the latest lts version, wait for the upgrade of `actions/setup-node@v4`, and then revert this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,12 +18,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node_version: [18.18.2, lts/*]
+        node_version: [18.18.2, 22.14.0]
         include:
           - os: macos-latest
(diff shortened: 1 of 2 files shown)


hannoeru:
@zyyv Maybe you want to take a look on this one: #4452.

I will also do some improvement on CI after #4452 merged.

zyyv:
> @zyyv Maybe you want to take a look on this one: #4452.
> 
> I will also do some improvement on CI after #4452 merged.

I saw it, disable corepack and abandon the built-in package manager. This is another solution. 👀

PR #4459, opened 2025-02-26
chore: remove default CJS exports fix

userquin:
This PR includes:
- update `unbuild` to `v3.5.0` including the default CJS export fix
- move `collections.json` to  `collections.ts` in `preset-icons`: tried with Node 20.10.0, 22.13.1 and 23.8.0 on my local and cannot build the preset
- update iconify collections update script to move the content to the ts module

image


packages-presets/preset-icons/build.config.ts
```diff
@@ -19,6 +19,9 @@ export default defineBuildConfig({
   declaration: true,
   externals,
   rollup: {
+    json: {
+      include: './src/collections.json',
+    },
```
userquin, line 24: remove this

zyyv (approved):

Issue #4503, opened 2025-03-12
Lowering specificity of Tailwind reset for buttons


Ericlm:
### UnoCSS version

66.1.0-beta.3

### Describe the bug

There is two way of resetting the styles with Tailwind, but the "base" one has a specificity that is too high, making it impossible to have for example a `bg-blue-500` for an idle button.

The second reset makes it possible by unsetting the background-color.

I think the ideal solution would be to lower the specificity of the selector making the background-color transparent.

Today, it is 0,1,0 while for Tailwind it is 0,0,1 thanks to the following selector : 
`button, input:where([type="button"]), input:where([type="reset"]), input:where([type="submit"])`.

This could also simplify the Tailwind reset by having only one reset available.

See #2127 and https://github.com/unocss/unocss/issues/2127#issuecomment-1712725688

### Reproduction

I think the provided informations are enough, but I could provide a reproduction if needed.

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4504, opened 2025-03-12
feat(reset): lower specificity on button

Ericlm:
Improves the Tailwind reset by lowering the specificity of the button reset from 0,1,0 to 0,0,1 to allow easier overwrite.

I think this could combine both of tailwind and tailwind-compat resets. If accepted, one of them could be removed.

See #4503


zyyv:
But we don't know what element effected with this attribute. `button` or `input`?

Ericlm:
> But we don't know what element effected with this attribute. `button` or `input`?

These are `button` and `input` that look like buttons (`submit`, `reset` and `button`)

Simon-He95:
https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/preflight.css#L363

Ericlm:
@zyyv what do you think? 🙂

zyyv:
> @zyyv what do you think? 🙂

Sorry, I can't accept PR for now. I'm not sure if the reset package will continue to exist in the future, because I plan to merge it into a specific preset as preflight css. Just like presetWind4.

I have already achieved this in `presetWind4`.
https://github.com/unocss/unocss/blob/main/packages-presets/preset-wind4/src/preflights/reset.ts

Regarding the previous reset style(for presetWind3), I don't want to find major changes.
If the difference is really smoothed out, I won't delete the other one.
According to https://github.com/search?q=@unocss/reset/tailwind-compat&type=code survey, the usage is more.

Ericlm:
Juste took a look at presetWind4 and it has the perfect reset, so thank you for it! 🙂

Issue #4523, opened 2025-03-21
`presetWind4` not working for `font-[xxx]`


lisonge:
### UnoCSS version

66.1.0-beta.6

### Describe the bug

https://tailwindcss.com/docs/font-weight

```text
font-(<custom-property>)         font-weight: var(<custom-property>);
font-[<value>]                   font-weight: <value>;
```

---

but got `.font-\[500\]{font-family:500;}` by unocss

### Reproduction

https://unocss.dev/play/#version=66.1.0-beta.6&html=DwEwlgbgBAxgNgQwM5ILwCIBmB7AdgFwFoBtAVgAZyBddAPgCgonmBVXbAYQGUv7gB6cBFpA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAOp40AFkoBfOFigQQcAORdcEAMZ06qpEgwAPSLHTYAhlwA28TDnxESwUgApkw2gxh0AXHABtKk9RGAlcaVcASiFUai9GFjZOHj5%2BaNiAXTkopCA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4IgLgTghgdgzgMwPYQLYAkyoDYgFwJTZwCmAvkA

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

[...]

PR #4526, opened 2025-03-22
fix(preset-mini): font-(.*)

Simon-He95:
close: #4523

reference: https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/src/utilities.ts#L2988


zyyv:
Sorry, I think this rule does not need to be changed, and changing the code is a bit complicated.

PR #4535, opened 2025-03-25
chore(core): skip empty token in `safelist`

Simon-He95:


zyyv:
Can you tell me the advantages of doing this? What utility is it specifically intended to capture?

zyyv (approved):

Issue #4522, opened 2025-03-20
How to enable suggestions for Pug's shorthand syntax?


shadowdreamer:
### Clear and concise description of the problem

I’m using @unocss/extractor-pug along with the VSCode extension, and  autocomplete works when defining classes inside  `class="..."`, but not when using shorthand syntax like `div.text-red....` in Pug；

Is there a way to enable suggestions for `div.text-...`  in Pug? It's works fine in Tailwindcss extension

### Suggested solution

Ideally, `.text-  .border- etc..` should also receive autocomplete suggestions, just like `class="text-red-500"`.

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Read the `README.md` of using the package.
- [x] Already used the [Interactive Docs](https://unocss.dev/interactive/).
- [x] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

shadowdreamer:
With Tailwindcss extension

![Image](https://github.com/user-attachments/assets/84ea9c9d-0543-4e62-af5e-837e8ed6621d)

With Unocss extension

![Image](https://github.com/user-attachments/assets/b75a3f7f-5335-4a9e-b67d-9e54eb031cea)

shorthand suggestions not work, only works in `class=""`

![Image](https://github.com/user-attachments/assets/0762a010-1ea5-4022-b226-01ba050e84cc)

PR #4543, opened 2025-03-27
fix(vscode): fix the vscode autocomplete in vue pug template

zhangmo8:
image

I tried to fix this issue #4522  and it worked, but I don't think this implementation is good, actually, I see another package of @unocss/extractor-pug dealing with pug-related stuff, but I don't know how to integrate it, I might need some help

@antfu @Simon-He95 @zyyv


packages-integrations/vscode/src/autocomplete.ts
```diff
@@ -74,13 +74,47 @@ export async function registerAutoComplete(
...
-        const result = await autoComplete.suggestInFile(code, doc.offsetAt(position))
+        const cursorPosition = doc.offsetAt(position)
+        let result: SuggestResult | undefined
+
+        // Special treatment for Pug Vue templates
+        if (isPug) {
+          // get content from cursorPosition
+          const textBeforeCursor = code.substring(0, cursorPosition)
+          // check the dot
+          const dotMatch = textBeforeCursor.match(/\.\w*-*$/)
```
Simon-He95, line 93: It is inaccurate to obtain valid text in this way.
UnoCSS is very flexible, such as .-w- or .w-10

Simon-He95: It can also be `.opacity-[0.1]`

zhangmo8: So I need to get the match rule of the current unocss, where should I find it

Simon-He95: Maybe you can match the regular expression to a valid word, and then split the last item not in `[]` or `()`? Because `autoComplete.suggest` will handle the following problems. I don't particularly understand the syntax of pub, will there be any strange things?

zhangmo8: There seems to be nothing wrong at the moment, but I actually just matched `.`, The match here doesn't care about [] and (), that's what the extractor does, and it does it well.

I just think this logic may not be appropriate here, and I have not found a suitable place to place this logic

shadowdreamer: Pug's abbreviation syntax does not support `w-[100px]` such that it will report an error; it is sufficient to obtain the basic presets in the shorthand, I think

Simon-He95: Then there should be no need to process the `.` in `[]` and `()`. Just match the last `.` of the valid content before the cursor position to the content at the cursor position.

zhangmo8: Yes, it is now.

Simon-He95: const dotMatch = textBeforeCursor.match(/\.\w*-*$/)
When you type `.-w-` (startsWith `-`) or `.w-1` there is no further prompt .w-1 -> [.w-12,.w-14,...]

zhangmo8: fixed

zhangmo8:
Actually, I found the code related to highlight `pug` [here](https://github.com/unocss/unocss/blob/main/virtual-shared/integration/src/match-positions.ts#L21)

But for highlights, the character will be a complete match className, and autocomplete only needs a part of it to expand the prompt, so I don't know where to find a suitable rule

Simon-He95:
I think there should be no problem with the functionality at present. Can we separate this logic into a getSuggestResult function? The code will look cleaner.

Simon-He95 (approved):

zyyv:
@Simon-He95 If you think it's ok, feel free to merge it.

PR #4546, opened 2025-03-28
fix(preset-wind4): shadow property

antfu:
I didn't look into the code very carefully yet, but I found this due to an syntax error in the generated css:

CleanShot 2025-03-28 at 13 05 02@2x

And it seems to be an typo?


zyyv:
Yes, there are some typo and fixed something in https://github.com/unocss/unocss/pull/4539

zyyv (approved):

hannoeru (approved):

PR #4547, opened 2025-03-28
feat(preset-wind4): hoist css properties

antfu:
WIP, I need to implement a way to make `noMerge` check for exact duplications

This PR introduces `symbols.noMerge`, and converts `defineProperty` from raw rules to CSS object

--- 

Another approach might be that we make it possible to assign layers to raw rules


antfu:
Should be good now

zyyv (approved):

PR #4548, opened 2025-03-28
perf(preset-wind4): hoist remRE

Simon-He95:
The regular expression object gets recreated every time the resolver function is called, which is less efficient than reusing the imported regex from ./handlers.

### DIFF
--- a/packages-presets/preset-wind4/src/utils/handlers/index.ts
+++ b/packages-presets/preset-wind4/src/utils/handlers/index.ts
@@ -5,5 +5,3 @@ export const handler = createValueHandler(valueHandlers)
 export const h = handler
 
 export { valueHandlers }
-
-export * from './regex'
--- a/packages-presets/preset-wind4/src/utils/handlers/index.ts
+++ b/packages-presets/preset-wind4/src/utils/handlers/index.ts
@@ -5,3 +5,5 @@ export const handler = createValueHandler(valueHandlers)
 export const h = handler
 
 export { valueHandlers }
+
+export * from './regex'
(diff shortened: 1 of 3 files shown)


packages-presets/preset-rem-to-px/package.json
```diff
@@ -38,6 +38,7 @@
     "test:attw": "attw --pack --config-path ../../.attw-esm-only.json"
   },
   "dependencies": {
-    "@unocss/core": "workspace:*"
+    "@unocss/core": "workspace:*",
+    "@unocss/preset-wind4": "workspace:*"
```
zyyv, line 42: I think it doesn't need to install additional dependency.

Simon-He95: if remove it，will build fail

zyyv: I mean, keep `preset-rem-to-px` the original unchanged and make it a pure package.
I don't think it has anything relationship with `preset-wind4`.

packages-presets/preset-rem-to-px/src/index.ts
```diff
@@ -1,6 +1,5 @@
 import { definePreset } from '@unocss/core'
-
-const remRE = /(-?[.\d]+)rem/g
+import { remRE } from '@unocss/preset-wind4/utils'
```
zyyv, line 2: Abandon the import from preset-wind4

packages-presets/preset-rem-to-px/package.json
Simon-He95: Get

packages-presets/preset-wind4/src/utils/handlers/index.ts
```diff
@@ -5,3 +5,5 @@ export const handler = createValueHandler(valueHandlers)
 export const h = handler
 
 export { valueHandlers }
+
+export * from './regex'
```
zyyv, line 9:
```suggestion
```

I don't think it makes sense to export it for external use. It attacks our internal rules.

zyyv (changes requested):

zyyv (approved):

PR #4554, opened 2025-03-31
feat(postcss): allow content from node_modules

HomyeeKing:
I think the `content.filesystem` options is like `@source` in [tailwindcss](https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources)

sometimes we may need to include the packages from `node_modules`, like [@llamaindex/chat-ui](https://www.npmjs.com/package/@llamaindex/chat-ui)

so it'd be better to be able access the node_modules

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -187,7 +187,7 @@ The usage extracted from each source will be **merged** together.
 - **Type:** `string[]`
 - **Default:** `[]`
 
-Glob patterns to extract from the file system, in addition to other content sources.
+Glob patterns to extract from the file system, in addition to other content sources. `node_modules` are ignored by default, but unocss will scan from it when you specify the path includes `node_modules`.
 
 In dev mode, the files will be watched and trigger HMR.
 
--- a/packages-integrations/postcss/src/esm.ts
+++ b/packages-integrations/postcss/src/esm.ts
@@ -103,12 +103,13 @@ export function createPlugin(options: UnoPostcssPluginOptions) {
     }
 
     const globs = uno.config.content?.filesystem ?? defaultFilesystemGlobs
+    const needCheckNodeMoudules = globs.some(i => i.includes('node_modules'))
     const plainContent = uno.config.content?.inline ?? []
 
     const entries = await glob(isScanTarget ? globs : [from], {
       cwd,
       absolute: true,
-      ignore: ['**/node_modules/**'],
+      ignore: needCheckNodeMoudules ? undefined : ['**/node_modules/**'],
       expandDirectories: false,
     })
 


johndunderhill:
@HomyeeKing @antfu This may be important for us.  We are converting from Tailwind and use a number of custom 'source' components in our Monorepo.  We install these via NPM and consume them by importing the TypeScript source directly, including the Atomic CSS inside our TSX components.  We build everything out using Vite.

Moreover, like many Monorepo users, our CSS configuration files and stylesheets, which comprise our Design System, are stored in a package, not in the the app itself.  This way those files can be shared among many apps.

For Atomic CSS tools to find these source files via the filesystem, they have to be able to scan `node_modules`, where the source components are installed, or the classes will not be seen.

This may be moot if they are seen via the Vite pipeline, but I'm concerned that this may not work, and I may need this as a fallback, and possibly for other use cases, including finding the configuration files and CSS files themselves.  We use a focused patturn within `node_modules` to find our components, so it's efficient enough:

```
// Part of tailwind.config.ts
const extensions = '.{astro,html,js,jsx,md,mdx,ts,tsx}';

const config = {
	content: {
		files: ['*' + extensions, 'src/**/*' + extensions, 'node_modules/@costvine/*/*' + extensions]
	},
	plugins: [forms, typography],
	darkMode: 'class',
	future: {
		hoverOnlyWhenSupported: true
	},
	theme: {
		colors: defaultColors
	}
};
```

Elsewhere:

```
// Part of another tailwind.config.ts
import config from '@costvine/colors/tailwind.config.js';
```

And in a package:

```
// Cardbox.tsx
import { Show, type JSX } from 'solid-js';
// Resolved via node_modules, may contain Atomic CSS classes
import Header from '@costvine/banners/Header.jsx';
import Rule from '@costvine/banners/Rule.jsx';

export default function (props: { title?: string; rule?: boolean; children: JSX.Element }) {
	return (
		<div class="w-full border-b px-6 py-5 sm:w-fit sm:mx-auto sm:border-none sm:shadow sm:rounded-lg sm:px-7 sm:py-6 lg:px-8 lg:py-7 cardbox-color">
			<Show when={props.title}>
				<Header>{props.title}</Header>
				<Show when={props.rule}>
					<Rule />
				</Show>
			</Show>
			{props.children}
		</div>
	);
}

```

If the pipeline works, then that would be great.  Just for your information, Tailwind 4 can't get their pipeline scanner to work, so they solved that problem by removing it, which is causing many breaking changes for Monorepo users:

https://github.com/tailwindlabs/tailwindcss/pull/16631

We are hoping to escape this drama.  We're a pure SPA--no SSR of any kind--so we're not concerned about the concurrent rendering problems which caused this disaster, although perhaps even those problems could have been solved.  We just want to get away from Tailwind, because they're hurting us.

It's seems gratuitous for Uno CSS to force `node_modules` to be ignored if the user explicitly includes it.  I have reviewed the changes in this PR and it looks good to me.

Appreciate all of you very much.

John

HomyeeKing:
@antfu @zyyv  CC 👀

zyyv (approved):

johndunderhill:
Thanks, everyone.

PR #4562, opened 2025-04-03
chore: adding llms.txt

imkarthikk:
- Adding LLMs file to better support LLMs to understand and use UnoCSS.

### DIFF
--- a/docs/public/llms.txt
+++ b/docs/public/llms.txt
@@ -1,0 +1,67 @@
+# UnoCSS - AI Reference Guide
+
+> UnoCSS is an atomic-CSS engine for high-performance, on-demand utility generation. It is highly customizable, extensible, and designed for modern frameworks like Vite, Astro, Nuxt, and more.
+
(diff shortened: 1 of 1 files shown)


antfu:
Thanks for the PR! But I wish it to be generated instead of requiring us to manually maintain it

imkarthikk:
Understood. I thought this would help a lot of people adopt Uno even better. Especially when vibe coding.Sent from my iPhoneOn 4 Apr 2025, at 8:57 AM, Anthony Fu ***@***.***> wrote:﻿
  Thanks for the PR! But I wish it to be generated instead of requiring us to manually maintain it—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



antfu left a comment (unocss/unocss#4562)
Thanks for the PR! But I wish it to be generated instead of requiring us to manually maintain it

zhangmo8:
So maybe we should write a script to get this information from the current markdown docs

Issue #4568, opened 2025-04-06
PresetWind4 and transformerDirectives conflict issues


wuiyang:
### UnoCSS version

66.1.0-beta.9

### Describe the bug

When using `presetWind4` and transformers `transformerDirectives`, using some classes (such as `ease`) would produce `TypeError: x.children is undefined` in browser playground, and `Pre-transform error: Cannot read properties of undefined (reading 'forEach')` in node environment.

This issue occurs from 66.1.0-beta.9 and 66.1.0-beta.10, but has no issue in 66.1.0-beta.8.

The error occurs in this line of code.
https://github.com/unocss/unocss/blob/a7ec2e4aee969a87e074828a183d14fa232951f9/packages-presets/transformer-directives/src/apply.ts#L99-L101

In v66.1.0-beta.9, VSCode UnoCSS intellisense shows the CSS code below for `ease-out`. Showing

```css
/* layer: default */
.ease-out {
  --un-ease: var(--ease-out);
  transition-timing-function: var(--ease-out);
}
.ease-out {
  syntax: "*";
  inherits: false;
}
```

![Image](https://github.com/user-attachments/assets/d2cddf25-72c4-4aa6-b2db-1477e86bc2a1)

In v66.1.0-beta.8, VSCode UnoCSS intellisense shows the CSS code below for `ease-out`.

```css
/* layer: default */
.ease-out {
  --un-ease: var(--ease-out);
  transition-timing-function: var(--ease-out);
}
@property --un-ease {
  syntax: "*";
  inherits: false;
}
```

![Image](https://github.com/user-attachments/assets/730da80e-de44-46c1-ad11-86b00d19a5ec)

### Reproduction

[...]

PR #4584, opened 2025-04-10
fix(4568): check children in selectorList

chb-dlght:
try to resolve this issue: [#4568](https://github.com/unocss/unocss/issues/4568). 

Just adding a check on children array.


zyyv:
Maybe it's not that simple, it's still lack of test.

PR #4592, opened 2025-04-15
fix(preset-wind4)!: abandon css theme variables hyphenatable

### LINKED ISSUE #4579: preset-wind4 colors declaration opacity is incorrect
### UnoCSS version

66.1.0-beta.10 (`@unocss/preset-wind4`)

### Describe the bug

If I declare a color in wind4, the corresponding class is created, but it contains an invalid setting, which is why the color doesn't appear.

### Reproduction

**uno.config.ts**
```ts
import presetWind4 from '@unocss/preset-wind4';
import { defineConfig } from 'unocss';

export default defineConfig({
  presets: [presetWind4()],
  theme: {
    colors: {
      veryCool: '#0ff', // here
    },
  },
});
```

* https://stackblitz.com/edit/vitejs-vite-tjji3wus?file=uno.config.ts

Please excuse me, it's not a professional reproduction, I just found an existing one on the internet, added the wind4, and it's visible that the div placed in the index.html doesn't display the correct color. However, when checking the developer tools, it's clear that the class was created:

```css
:root, :host {
  --colors-very-cool: #0ff;
}

.text-very-cool {
  color: color-mix(in oklch, var(--colors-very-cool) var(--un-text-opacity), transparent);
}
```

But this way, the basic syntax of CSS doesn't allow the code to run; the reason for this is:

> `--un-text-opacity` is not defined

The perfect solution would be a fallback 100% value:

```css
:root, :host {
  --colors-very-cool: #0ff;
}

.text-very-cool {
  color: color-mix(in oklch, var(--colors-very-cool) var(--un-text-opacity, 100%), transparent);
}
```

I found the error on Stack Overflow and answered it there. You can also quickly check the executable code snippets there:
* [Add custom theme spacing to unocss for tailwind 4 preset](https://stackoverflow.com/a/79561728/15167500)

### System Info

_No response_

### Validations

zyyv:
close #4579 

This pr changes are as follows：
The theme preflights preset by unocss wind4 is different from tailwind4. We will generate all subkeys of the practical key of the theme and keep the original naming rules of the key.

For example, `--fontWeight-DEFAULT` will not `hyphenate` on `fontWeight`, and will not remove `DEFAULT`, so as to ensure that all subkeys are generated and unocss will not interfere with the user's subsequent customization of the theme.


NikitaVelixJob:
@zyyv hello! When it will be merged? I have the same issue with different colors. Thank you!

zyyv:
> @zyyv hello! When it will be merged? I have the same issue with different colors. Thank you!

Could you open a new issue report it?

Issue #4594, opened 2025-04-15
translating negative value


minkoonim:
### UnoCSS version

66.1.0-beta.10

### Describe the bug

I've been putting `-` at the front for negative value

for example `class="-m-1 -p-1" or "-m1 -p1"`
resulting in
```
.-m-1 {
  margin: calc(calc(var(--spacing) * 1) * -1);
}

.-p1 {
  padding: calc(calc(var(--spacing) * 1) * -1);
}

```

but for translate,
`-translate-x-50%`
results in positive value since double negative multiply
`-50% * -1 = 50%`
`translate-x--50%` works though
```
.-translate-x-50\% {
  --un-translate-x: -50%;
  translate: calc(var(--un-translate-x) * -1) var(--un-translate-y);
}

.translate-x--50\% {
  --un-translate-x: -50%;
  translate: var(--un-translate-x) var(--un-translate-y);
}
```

### Reproduction

cause and result is clear

### System Info

win11 wsl2 ubuntu24
bun latest unocss latest

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

henrikvilhelmberglund:
Needs repro, I get 

```ts
.-translate-x-50\%,
.translate-x--50\%{--un-translate-x:-50%;transform:translateX(var(--un-translate-x)) translateY(var(--un-translate-y)) translateZ(var(--un-translate-z)) rotate(var(--un-rotate)) rotateX(var(--un-rotate-x)) rotateY(var(--un-rotate-y)) rotateZ(var(--un-rotate-z)) skewX(var(--un-skew-x)) skewY(var(--un-skew-y)) scaleX(var(--un-scale-x)) scaleY(var(--un-scale-y)) scaleZ(var(--un-scale-z));}
```
for both.

minkoonim:
sorry I forgot to mention it's wind4
[...]

PR #4601, opened 2025-04-18
fix(core,preset-wind4): add `operators` type to `CSSEntry` & resovle negative value in `translate`

Simon-He95:
close: #4594


packages-engine/core/src/types.ts
```diff
@@ -15,7 +15,7 @@ export type PartialByKeys<T, K extends keyof T = keyof T> = FlatObjectTuple<Part
 export type RequiredByKey<T, K extends keyof T = keyof T> = FlatObjectTuple<Required<Pick<T, Extract<keyof T, K>>> & Omit<T, K>>
 
 export type CSSObject = Record<string, string | number | undefined>
-export type CSSEntry = [string, string | number | undefined]
+export type CSSEntry = [string, string | number | undefined, '$$mini-no-negative'?]
```
zyyv, line 18: I got your means. Cause the `negative` variant will process the css entry all of utility, considering the future scalability of the entry, have you optimized the design?

Simon-He95: Currently, the second parameter is used as the basis for judgment. Considering the scalability, the last parameter can be used to judge whether it is `$$mini-no-negative`, or do you have any better suggestions?

zyyv (approved):

Issue #4606, opened 2025-04-19
[transformer-directives]: mangled `@apply` chained breakpoints


pauliesnug:
### UnoCSS version

v66.0.0

### Describe the bug

When using chained breakpoints (i.e., `sm:lg:w-[40em]`) with an `@apply` directive, UnoCSS seems to mangle the code and use `$$` in the transformed CSS.

For example, if I make the following stylesheet and transform it with UnoCSS:
```css
body {
  @apply sm:lg:w-[40em];
}
```

I expect to see the following code (which Tailwind v3 does produce):
```css
@media (min-width: 640px) {
  @media (min-width: 1024px) {
    body {
      width: 40em;
    }
  }
}
```

Instead, the following code is transpiled:
```css
body {
}@media (min-width: 1024px) $$ @media (min-width: 640px){body{width:40em;}}
```

Seemingly, the `$$` syntax doesn't even exist, and the VSCode extension also gets confused and previews different code. Is this intentional? Does `transformer-directives` not support chaining breakpoints?

### Reproduction

[...]

PR #4614, opened 2025-04-22
fix(transformer-directives): mangled @apply chained breakpoints

Simon-He95:
close: #4606


packages-presets/transformer-directives/src/apply.ts
```diff
@@ -119,6 +127,7 @@ export async function parseApply({ code, uno, applyVariable }: TransformerDirect
         code.appendRight(childNode!.loc!.end.offset + semicolonOffset, body)
     }
   }
+  // todo: after transformered remove empty css like `selector{}`
```
zyyv, line 130: Do you want to continue completing this TODO in this PR?

Simon-He95: This function has nothing to do with this pr. I just think it is a point that can be optimized and can be put in the later pr.

PR #4637, opened 2025-05-05
feat(preset-wind4): add noscript variant in Tailwind 4.1

OverflowCat:
https://tailwindcss.com/blog/tailwindcss-v4-1#new-noscript-variant

Not adding it into the mini preset, as there are 3 values of the media feature. Do we need to support all of them?

> The scripting feature is specified as a keyword value chosen from the list below.
>
> [none](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/scripting#none)
Scripting is completely unavailable on the current document.
>
> [initial-only](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/scripting#initial-only)
Scripting is enabled during the initial page load, but not afterwards.
>
> [enabled](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/scripting#enabled)
Scripting is supported and active on the current document.

### DIFF
--- a/packages-presets/preset-wind4/src/variants/default.ts
+++ b/packages-presets/preset-wind4/src/variants/default.ts
@@ -10,7 +10,7 @@ import { variantColorsMediaOrClass, variantColorsScheme } from './dark'
 import { variantDataAttribute, variantTaggedDataAttributes } from './data'
 import { variantLanguageDirections } from './directions'
 import { variantImportant } from './important'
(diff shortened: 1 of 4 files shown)


zyyv (approved):

zyyv:
> Do we need to support all of them?

@OverflowCat  We can support all, but the default is align with `tailwind` when no value is given. Default is `scripting: none`

antfu (approved):

zyyv (approved):

PR #4653, opened 2025-05-11
fix(vscode-ext): display oklch colors inline

### LINKED ISSUE #4651: VSCode extension - Colors in oklch not displayed
### UnoCSS version

66.1.1

### Describe the bug

When using the Uno VSCode extension, it seems that the colors that come from oklch are not displayed as a "badge" along the text : 

Image

 
However, the white color has no problem, I think because it is correctly "translated" into hexa : 

Image

Which is not the case of the pink-500 : 

Image

And here is an example of the result of the Tailwind extension:

Image

### Reproduction

https://github.com/Ericlm/uno-extension-colors

### System Info

macOS 15.4.1 (24E263)

**VSCode :** 
Version: 1.100.0 (Universal)
Commit: 19e0f9e681ecb8e5c09d8784acaa601316ca4571
Date: 2025-05-07T12:48:53.763Z
Electron: 34.5.1
ElectronBuildId: 11369351
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

### Validations

Ericlm:
This PR is for the VSCode extension, allowing it to display colors inline with the text when they are in oklch.

The only change is to make `cssColorRE` match oklch colors.

Fix #4651

### DIFF
--- a/packages-integrations/vscode/src/utils.ts
+++ b/packages-integrations/vscode/src/utils.ts
@@ -6,7 +6,7 @@ import prettier from 'prettier/standalone'
 
 const remUnitRE = /(-?[\d.]+)rem(\s+!important)?;/
 const matchCssVarNameRE = /var\((?<cssVarName>--[^,|)]+)(?:,(?<fallback>[^)]+))?\)/g
-const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:rgb|hsl|oklch)a?\(.*\)/g
+const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|((?:rgb|hsl)a?|oklch)?\(.*\)/g
 const varFnRE = /var\((--[^,|)]+)(?:,([^)]+))?\)/
 
 export function throttle<T extends ((...args: any) => any)>(func: T, timeFrame: number): T {
--- a/packages-integrations/vscode/src/utils.ts
+++ b/packages-integrations/vscode/src/utils.ts
@@ -6,7 +6,7 @@ import prettier from 'prettier/standalone'
 
 const remUnitRE = /(-?[\d.]+)rem(\s+!important)?;/
 const matchCssVarNameRE = /var\((?<cssVarName>--[^,|)]+)(?:,(?<fallback>[^)]+))?\)/g
-const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:rgb|hsl)a?\(.*\)/g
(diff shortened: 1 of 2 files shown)


packages-integrations/vscode/src/utils.ts
```diff
@@ -6,7 +6,7 @@ import prettier from 'prettier/standalone'
 
 const remUnitRE = /(-?[\d.]+)rem(\s+!important)?;/
 const matchCssVarNameRE = /var\((?<cssVarName>--[^,|)]+)(?:,(?<fallback>[^)]+))?\)/g
-const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:rgb|hsl)a?\(.*\)/g
+const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:rgb|hsl|oklch)a?\(.*\)/g
```
hannoeru, line 9: Can you move oklch to other capture group? becuase oklch should not include `oklcha` when matching.

hannoeru (approved):
LGTM! Thanks.

packages-integrations/vscode/src/utils.ts
```diff
@@ -6,7 +6,7 @@ import prettier from 'prettier/standalone'
 
 const remUnitRE = /(-?[\d.]+)rem(\s+!important)?;/
 const matchCssVarNameRE = /var\((?<cssVarName>--[^,|)]+)(?:,(?<fallback>[^)]+))?\)/g
-const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:rgb|hsl)a?\(.*\)/g
+const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|((?:rgb|hsl)a?|oklch)?\(.*\)/g
```
zyyv, line 9:
```suggestion
const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:(?:rgb|hsl)a?|oklch)?\(.*\)/g
```
It doesn't look like it needs capture group.

zyyv (changes requested):

zyyv:
And can you provide a screenshot of a successful run?

Ericlm:
Hmmm… Actually, it displays a white bubble, I think because of oklch. I guess I messed up with the config of the project where I was testing and that's why I believed I was using tailwind 4, and oklch…

So I think I'm gonna need to make `convertToRGBA` support oklch? That will be useful for showing them in the completion menu too. @zyyv

antfu (approved):

zyyv (approved):

PR #4657, opened 2025-05-13
feat(virtual-shared): update defaultPipelineInclude to support vue-vine

zhangmo8:
I tried using unocss in vue-vine, but unocss does not support `.ts|js` by default, so I hope unocss can support vine.ts by default

Minimum reproduction path [here](https://stackblitz.com/edit/unocss-unocss-59z5bv85?file=uno.config.ts)

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -208,7 +208,7 @@ Set `false` to disable.
 ##### include
 
 - **Type:** `FilterPattern`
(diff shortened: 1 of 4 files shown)


antfu (approved):

antfu:
I am ok adding it. Just note it's a feat not a chore

PR #4698, opened 2025-05-31
feat(postcss): add `@unocss !<layer-name>` to exclude a layer

### LINKED ISSUE #4697: PostCSS: At-rule to exclude layers (`@unocss !<layer-name>`)

Layers cannot be individually excluded from `@unocss` at-rule.

### Suggested solution

Currently, we can inject each layer individually using `@unocss <layer-name>` at-rule with the layer name, e.g.: `@unocss preflights`. This will replace the at-rule placeholder with only the generated `preflights` layer (and add the layer to the exclude list to not generate it again in the fallback `@unocss` at-rule).

I suggest being able to specify layers to be excluded in `@unocss` at-rule.
Maybe a syntax like: `@unocss !<layer-name> ...`

**Example**

This should generate all the rules, _except_ the `preflights` layer:

```postcss
@unocss !preflights;
@unocss;
```

### Alternative

_No response_

### Additional context

I'm trying to separate the `properties` layer from the rest of the styles into a file (ref: `presetWind4`).

**File: `properties.css`**

```postcss
@unocss properties;
```

**File: `style.css`**

```postcss
@unocss;
```

**File: `main.ts`**

```ts
...
import "./properties.css"
import "./style.css"
...
```

But the `@unocss` at-rule only operates at the file-level; So, it won't exclude the `properties` layer specified in another file.

With this proposal, I can adjust the `style.css` file to be like this, and it should work:

```postcss
@unocss !properties;
@unocss;
```

### Validations

songpola:
Close #4697


zyyv:
Would you like to add some tests?

songpola:
I haven't written test before, but I will try.

songpola:
@zyyv I have added the test. Is it ok?

PR #4682, opened 2025-05-20
fix(preset-wind4): support numeric values in pseudo variants

onmax:
Related #4679 

Not sure how to run the playground/tests to make sure this change does not break anything.

I have installed dependencies in the root level and run `nr play`.

I get this error:

```
failed to load config from /home/maxi/personal/unocss/playground/vite.config.ts
error when starting dev server:
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/maxi/personal/unocss/node_modules/.pnpm/jiti@2.4.2/node_modules/.pnpm/jiti@2.4.2/node_modules/jiti/lib/jiti.mjs' imported from /home/maxi/personal/unocss/node_modules/.pnpm/jiti@2.4.2/node_modules/jiti/lib/jiti.mjs
    at finalizeResolution (node:internal/modules/esm/resolve:275:11)
    at moduleResolve (node:internal/modules/esm/resolve:932:10)
    at defaultResolve (node:internal/modules/esm/resolve:1056:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:654:12)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:603:25)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:586:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:242:38)
    at onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:546:36)
    at TracingChannel.tracePromise (node:diagnostics_channel:344:14)
    at ModuleLoader.import (node:internal/modules/esm/loader:545:21)
```

I also tried to prune store.

Any advice on how to fix this issue and test this PR is appreciated it :)

PR #4720, opened 2025-06-08
refactor(preset-wind4, preset-mini): deduplicate pseudo variant code

onmax:
Task: https://github.com/unocss/unocss/pull/4682#pullrequestreview-2899464656

deduplicate the 'pseudo' variant code between `preset-mini` and `preset-wind4`.

Duplicate pseudo-variant logic was moved from preset-mini and preset-wind4 into shared utilities in rule-utils. Both presets had nearly identical files that only differed in type annotations and import paths.

- New: - `packages/presets/rule-utils/src/pseudo.ts`
- Shared pseudo variant utilities
- Refactored the files `preset-mini/src/_variants/pseudo.ts` and `preset-wind4/src/_variants/pseudo.ts`.
- `rule-utils/test/pseudo.test.ts`
- Tests in 'preset-mini', 'preset-wind4' and 'rule-utils' are passing.

Both presets continue to work exactly as before, but now share the underlying pseudo variant implementation while maintaining their specific type annotations and configurations.


onmax:
Please, let me know if this is the right approach of if instead I should move code to the preset-mini and import mini from wind 4. 

Any other approach/comment is also welcomed!

zyyv:
> Please, let me know if this is the right approach of if instead I should move code to the preset-mini and import mini from wind 4.
> 
> Any other approach/comment is also welcomed!

This is the next step I will take. I am going to extract the same logic in preset-mini and preset-wind4 into a new package and share it with preset-mini and preset-wind4, but I hope it is a private package, so I don't plan to put it in `rule-utils`.

I think I will temporarily pass this PR, because in the future, similar to `pseudo` files, I will move them to the correct package.

Thanks for your contribution.

zyyv (approved):

onmax:
If there are more tasks that you feel you need help, I am more than happy to help :)

PR #4736, opened 2025-06-12
chore: use tsgo run typecheck

zyyv:


package.json
```diff
@@ -60,6 +60,7 @@
     "@types/splitpanes": "catalog:",
     "@types/vscode": "^1.84.0",
     "@types/ws": "catalog:",
+    "@typescript/native-preview": "catalog:",
```
hannoeru, line 62: I think you also need to add this to pnpm's onlyBuiltDependencies.

.github/workflows/ci.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         run: pnpm test
 
       - name: Typecheck
+        if: matrix.node_version != '18.20.7'
```
zyyv, line 56: Error log: https://github.com/unocss/unocss/actions/runs/15613165647/job/43979000089#step:9:12

I will temporarily abandon support for lower versions of typecheck, hoping it will be fixed in the future.

PR #4758, opened 2025-06-21
feat(svelte-scoped): handle clsx like classes

### LINKED ISSUE #3631: Svelte Scoped doesn't work with CLSX
### UnoCSS version

0.58.6

### Describe the bug

using clsx breaks the classes

### Reproduction

https://stackblitz.com/edit/unocss-unocss-nfuydr?file=src%2Froutes%2F%2Bpage.svelte

### System Info

_No response_

### Validations

fehnomenal:
fix #3631

### DIFF
--- a/docs/integrations/svelte-scoped.md
+++ b/docs/integrations/svelte-scoped.md
@@ -42,14 +42,15 @@ A regular UnoCSS/Tailwind CSS setup places utility styles in a global CSS file w
 
 Because Svelte Scoped rewrites your utility class names, you are limited in where you can write them:
 
-| Supported Syntax          | Example                          |
-| ------------------------- | -------------------------------- |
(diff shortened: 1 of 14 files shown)


henrikvilhelmberglund (changes requested):
LGTM but I would like the [example index page](https://github.com/unocss/unocss/blob/main/examples/sveltekit-scoped/src/routes/%2Bpage.svelte) to be updated with an example (basically something like the test Svelte file) and the feature mentioned in the [docs](https://github.com/unocss/unocss/blob/main/docs/integrations/svelte-scoped.md) so it's more obvious that this works now.

fehnomenal:
Thanks for your suggestion. Please have another look.

henrikvilhelmberglund (approved):
LGTM

PR #4762, opened 2025-06-23
fix(core): respect `preprocess` type definitions

zyyv:


packages-engine/core/src/generator.ts
```diff
@@ -128,9 +128,7 @@ class UnoGeneratorInternal<Theme extends object = object> {
     if (this.cache.has(cacheKey))
       return this.cache.get(cacheKey)
 
-    let current = raw
-    for (const p of this.config.preprocess)
-      current = p(raw)!
```
zyyv, line 133: We can't roughly assert that process has a value, because by definition, it doesn't always need to return a value.

PR #4784, opened 2025-06-30
feat: Update config.ts

1411430556:
Click on the copyright information at the bottom to jump to Anthony Fu's GitHub homepage.


antfu:
Thank you, but I don't think it's necessary.

PR #4788, opened 2025-07-01
fix(vscode): fix custom theme colors display

### LINKED ISSUE #4786: vscode unocss extension not show correct color by wind4
### UnoCSS version

66.3.2

### Describe the bug

In the `uno.config.js`：
```ts
const UnoColors = { sb2: '#CDE8FF'  }

{
  theme: {
      colors: UnoColors
  }
}
```

in code，always show black point

Image

### Reproduction

this topics

### System Info

_No response_

### Validations

lzl0304:
Fixes #4786  Related #4653

### DIFF
--- a/packages-integrations/vscode/src/utils.ts
+++ b/packages-integrations/vscode/src/utils.ts
@@ -6,7 +6,7 @@ import prettier from 'prettier/standalone'
 
 const remUnitRE = /(-?[\d.]+)rem(\s+!important)?;/
 const matchCssVarNameRE = /var\((?<cssVarName>--[^,|)]+)(?:,(?<fallback>[^)]+))?\)/g
-const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:(?:rgb|hsl)a?|oklch)?\(.*\)/g
+const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:rgb|rgba|hsl|hsla|oklch)\(.*\)/g
 const varFnRE = /var\((--[^,|)]+)(?:,([^)]+))?\)/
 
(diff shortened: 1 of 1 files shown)


JunyWuuuu91:
Hello, could you expedite the review process? The issue is affecting my ability to determine whether the problem lies with my theme color variables not taking effect or some other error after coding. @zyyv

packages-integrations/vscode/src/utils.ts
```diff
@@ -6,7 +6,7 @@ import prettier from 'prettier/standalone'
 
 const remUnitRE = /(-?[\d.]+)rem(\s+!important)?;/
 const matchCssVarNameRE = /var\((?<cssVarName>--[^,|)]+)(?:,(?<fallback>[^)]+))?\)/g
-const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:(?:rgb|hsl)a?|oklch)?\(.*\)/g
+const cssColorRE = /(?:#|0x)(?:[a-f0-9]{3}|[a-f0-9]{6})\b|(?:rgb|rgba|hsl|hsla|oklch)\(.*\)/g
```
zyyv, line 9: Is there any difference between this rewrite and the original one? It seems to be the same behavior

lzl0304: The rewritten regex differs from the original. In the original, the `?` in `(?:rgb|hsl)a?|oklch)?` makes the prefix optional, allowing it to match any parenthetical content (e.g., `(foo)` or `()`). The rewritten version requires an explicit prefix (`rgb`, `rgba`, `hsl`,` hsla`, or `oklch`), making it stricter for CSS colors.

zyyv (approved):
Thank you. There is an extra question mark at the end of the regular expression.

PR #4793, opened 2025-07-02
chore(svelte-scoped): resolve deprecation of preprocessing from vite plugins

fehnomenal:
The `vite-plugin-svelte` has deprecated passing preprocessors to itself: https://github.com/sveltejs/vite-plugin-svelte/pull/1145
This will be released in the next major and results in this warning:

> [!WARNING]
> [vite-plugin-svelte] The following plugins use the deprecated 'plugin.api.sveltePreprocess' field. Please contact their maintainers and ask them to use a vite plugin transform instead: unocss:svelte-scoped:pass-preprocess

This PR should probably be postponed after that major release and have the peer dep version adjusted.

### DIFF
--- a/packages-integrations/svelte-scoped/package.json
+++ b/packages-integrations/svelte-scoped/package.json
@@ -57,7 +57,13 @@
     "test:attw": "attw --pack --config-path ../../.attw-esm-only.json"
   },
   "peerDependencies": {
+    "@sveltejs/vite-plugin-svelte": "catalog:",
     "svelte": "catalog:"
+  },
+  "peerDependenciesMeta": {
+    "@sveltejs/vite-plugin-svelte": {
+      "optional": true
+    }
   },
   "dependencies": {
     "@unocss/config": "workspace:*",
--- a/packages-integrations/svelte-scoped/package.json
+++ b/packages-integrations/svelte-scoped/package.json
@@ -56,6 +56,9 @@
     "test:integration": "pnpm -r --filter=\"./test/fixtures/*\" test",
     "test:attw": "attw --pack --config-path ../../.attw-esm-only.json"
   },
+  "peerDependencies": {
+    "svelte": "catalog:"
+  },
   "dependencies": {
     "@unocss/config": "workspace:*",
     "@unocss/preset-uno": "workspace:*",
--- a/packages-integrations/svelte-scoped/src/_vite/passPreprocessToSveltePlugin.ts
+++ b/packages-integrations/svelte-scoped/src/_vite/passPreprocessToSveltePlugin.ts
@@ -1,24 +1,44 @@
 import type { Plugin } from 'vite'
 import type { SvelteScopedContext } from '../preprocess'
 import type { UnocssSvelteScopedViteOptions } from './types'
+import { preprocess } from 'svelte/compiler'
 import UnocssSveltePreprocess from '../preprocess'
 
 export function PassPreprocessToSveltePlugin(context: SvelteScopedContext, options: UnocssSvelteScopedViteOptions = {}): Plugin {
-  let commandIsBuild = true
-  const isBuild = () => commandIsBuild
+  let preprocessor: ReturnType<typeof UnocssSveltePreprocess>
 
-  return {
+  const plugin = {
     name: 'unocss:svelte-scoped:pass-preprocess',
     enforce: 'pre',
 
-    configResolved({ command }) {
-      commandIsBuild = command === 'build'
+    configResolved({ command, plugins }) {
+      preprocessor = UnocssSveltePreprocess(options, context, () => command === 'build')
+
+      // use the exact same id filter as vite-plugin-svelte itself
+      const svelteIdFilter = plugins.find(p => p.name === 'vite-plugin-svelte:config')?.api.idFilter
+      plugin.transform.filter = svelteIdFilter
     },
 
-    api: {
-      sveltePreprocess: UnocssSveltePreprocess(options, context, isBuild),
+    transform: {
+      filter: {
+        // filled above
+      },
+
+      // after base preprocess, but before compile
+      order: 'pre',
+
+      async handler(source, id) {
+        const { code, map } = await preprocess(source, preprocessor, { filename: id })
+
+        if (typeof map === 'string')
+          return { code, map }
+
+        return { code }
+      },
     },
-  }
+  } satisfies Plugin
+
+  return plugin
 }
 
 // reference: https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#how-do-i-add-a-svelte-preprocessor-from-a-vite-plugin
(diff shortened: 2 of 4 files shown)


dominikg:
wow, this was quick, thank you!

you might be able to refactor the transform as you are no longer bound to svelte's preprocessor api with the new transform hook.

This change should also have an optional? peer dependency on vite-plugin-svelte@6, otherwise it would break for users who still use vite-plugin-svelte@5.

fehnomenal:
> This change should also have an optional? peer dependency on vite-plugin-svelte@6, otherwise it would break for users who still use vite-plugin-svelte@5.

Actually not. I also thought this would be the case but the sveltekit-scoped example uses
```yaml
      '@sveltejs/vite-plugin-svelte':
        specifier: ^5.0.3
        version: 5.1.0(svelte@5.34.7)(vite@6.3.5)
```
and it works just fine.

fehnomenal:
> you might be able to refactor the transform as you are no longer bound to svelte's preprocessor api with the new transform hook.

I hoped to do this but I'm not having the capacity at the moment. Also, instead of manually parsing the SFC into markup and style the preprocess API already gives us this. And I think it is better to keep having a single way of transforming the components to reduce code duplication. In fact I already fixed one discrepancy: #4794

dominikg:
it only works when the uno plugin is in front of the sveltekit plugin in the vite config and the files being processed by it are already valid svelte syntax so other markup preprocessors like mdsvex could cause an issue because then the unocss svelte scoped vite plugin would get to read markdown instead of svelte syntax.

dominikg:
> > you might be able to refactor the transform as you are no longer bound to svelte's preprocessor api with the new transform hook.
> 
> I hoped to do this but I'm not having the capacity at the moment. Also, instead of manually parsing the SFC into markup and style the preprocess API already gives us this. And I think it is better to keep having a single way of transforming the components to reduce code duplication. In fact I already fixed one discrepancy: #4794

this is indeed a limitation, vite plugin transforms are bascially always markup preprocessors and would have to parse it if they are interested in script or style blocks specifically.

We are working on improvements for that too. For vite-plugin-svelte@6 you can also opt to keep using the plugin.api.sveltePreprocess way, happy to adjust or remove the runtime logged warning if that annoys uno users too much. I super glad you are putting this through its paces  :hearts:

fehnomenal:
> it only works when the uno plugin is in front of the sveltekit plugin in the vite config and the files being processed by it are already valid svelte syntax so other markup preprocessors like mdsvex could cause an issue because then the unocss svelte scoped vite plugin would get to read markdown instead of svelte syntax.

I guess my experiments below were just pure luck. The `vite-plugin-svelte@5` does not expose `api.idFilter` and thus the transform is not filtered at all and runs on all files and in the examples this does not give any problems. In a real project the scoped plugin ran on [this file](https://github.com/melt-ui/melt-ui/blob/develop/src/lib/builders/calendar/create.ts) among others and interpreted the comments as real components and added the `<style>` block to the end which in turn resulted in obviously invalid JS.

So thanks for your concerns, I updated the peerDependency.

Old:
~OK, so I'm absolutely not sure I setup everything correctly. But using mdsvex and importing a `.md` file with `<p class="c-red">` worked regardless of the vite plugin (`sveltekit` and `unocss`) and preprocessor (`vitePreprocess` and `mdsvex`) order and `vite-plugin-svelte` version (5.1.0 and 6.0.0-next.2).~

fehnomenal:
> > > you might be able to refactor the transform as you are no longer bound to svelte's preprocessor api with the new transform hook.
> > 
> > 
> > I hoped to do this but I'm not having the capacity at the moment. Also, instead of manually parsing the SFC into markup and style the preprocess API already gives us this. And I think it is better to keep having a single way of transforming the components to reduce code duplication. In fact I already fixed one discrepancy: #4794
> 
> this is indeed a limitation, vite plugin transforms are bascially always markup preprocessors and would have to parse it if they are interested in script or style blocks specifically.
> 
> We are working on improvements for that too. For vite-plugin-svelte@6 you can also opt to keep using the plugin.api.sveltePreprocess way, happy to adjust or remove the runtime logged warning if that annoys uno users too much. I super glad you are putting this through its paces ♥️

I'm not sure what is the best approach. Probably let the maintainers decide @jacob-8

fehnomenal:
I marked this as a draft because I'm thinking about to separate the vite plugin and the svelte preprocessor. This would allow us to get rid of the complex handling of global classes by importing a virtual css file (prompted by disussion at https://github.com/sveltejs/vite-plugin-svelte/issues/1157).

But this would only work for the vite plugin as the preprocessor has no idea about any bundler or such a thing and thus cannot simply import anything that it does not know for sure exists (or at least can be somehow imported).

I did a bit of research and found 31 imports of the preprocessor (https://github.com/search?q=%40unocss%2Fsvelte-scoped%2Fpreprocess+path%3Asvelte.config&type=code) but:
- some are false positives (unused, example in this repo or commented)
- some **ADDITIONALLY** use the vite plugin (`@unocss/svelte-scoped/vite`)
	- https://github.com/henrikvilhelmberglund/fancy-caption
	- https://github.com/TESD-Tech/ps-pennsylvania
	- https://github.com/RedstoneWizard08/New-LPS-Website
	- https://github.com/Tricked-dev/knucklebones
	- https://github.com/mawoka-myblock/FediPrint
	- https://github.com/zapal-tech/zapal
	- https://github.com/henrikvilhelmberglund/Tutorials
	- https://github.com/tyom/tyom.semonov.com
	- https://github.com/Ikziriv/sveltekit-web-starter
	- https://github.com/Create-Technomancy/create-technomancy.github.io
	- https://github.com/murl-digital/scalar-rs
	- https://github.com/mate-h/typescale
- some **ADDITIONALLY** use the main vite plugin (`unocss/vite`)
	- https://github.com/LynithDev/meme-maker/tree/main/apps/frontend
	- https://github.com/lyha23/filio-svelte
	- https://github.com/lyha23/comac-web
	- https://github.com/partei-bergstrasse/partei-bergstrasse.de
- legit usages but also use vite and thus could migrate:
	- https://github.com/diamondburned/e2clicker
	- https://github.com/berlintech-ai/kommo-chrome-extension
	- https://github.com/jacob-8/kitbook/blob/main/packages/kitbook
	- https://github.com/boereh/antsvelte
	- https://github.com/jacob-8/svelte-pieces
	- https://github.com/Ladvace/Basic-sveltekit-boilerplate
	- https://github.com/henrikvilhelmberglund/vscode-hotkey-viewer
	- https://github.com/fivaz3/uno-css-preprocessor-bug-report (is probably inactive guessing from the repo name)
	- https://github.com/henrikvilhelmberglund/unocss-unocss-y3ienz
	- https://github.com/chrrs/freddit
- legit usages:
	- https://github.com/7rs/map (uses astro and applies `unocss/vite` and `unocss/astro`. Here I'm definitely not sure)

Not trying to call anybody out but it seems there is a bit of a confusion between the vite plugin and the preprocessor. And the only potentially real usage might be the last project all other (at least public on GitHub) could switch to `@unocss/svelte-scoped/vite` if we were to deprecate the preprocessor.

@jacob-8 Two projects are from you. Is there a reason you chose the preprocessor over the vite plugin?

PR #4801, opened 2025-07-04
feat(eslint): support sort in function calls like clsx

magicdawn:
the `unoFunctions` name take from [prettier-plugin-tailwindcss](https://www.npmjs.com/package/prettier-plugin-tailwindcss#:~:text=//%20.prettierrc%0A%7B%0A%20%20%22tailwindFunctions%22%3A%20%5B%22clsx%22%5D%0A%7D) `tailwindFunctions`

- no css included
- `<brand>Functions`

### DIFF
--- a/packages-integrations/eslint-plugin/src/rules/order.ts
+++ b/packages-integrations/eslint-plugin/src/rules/order.ts
@@ -16,10 +16,29 @@ export default createRule({
     messages: {
       'invalid-order': 'UnoCSS utilities are not ordered',
     },
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          unoFunctions: {
+            type: 'array',
+            items: {
+              type: 'string',
+            },
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
   },
-  defaultOptions: [],
+  defaultOptions: [
+    {
+      unoFunctions: ['clsx', 'classnames'],
+    },
+  ],
   create(context) {
+    const { unoFunctions = ['clsx', 'classnames'] } = context.options[0] || {}
+
     function checkLiteral(node: TSESTree.Literal | SvelteLiteral, addSpace?: 'before' | 'after' | undefined) {
       if (typeof node.value !== 'string' || !node.value.trim())
         return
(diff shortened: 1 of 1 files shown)


magicdawn:
PTAL

if the `unoVariables` part too much, default option can be set to empty array.

antfu:
Overall I am happy to have this feature. I'll leave it to @zyyv to check the implementation (Thanks!)

danielwaltz:
I love this and am very excited to see it land! Maybe not suitable for this chunk of work, but I would also love to see this working inside of config files `uno.config.ts` as well, specifically for shortcuts. Would be happy to make a separate issue for it though.

The amount of time I've copied classes there to an element in a component in my project to get consistent ordering is too much. 😅

magicdawn:
What I think: to sort uno.config.ts (or any other file) config.shortcuts can use 

1. split shortcuts to a object, and the name matches specific pattern, e.g `^cls`
	```ts
	const clsShortcuts = {
	  'flex-center': 'flex items-center justify-center'
	}
	export default defineConfig({
	  shortcuts: clsShortcuts.
	})
	```
2. OR use a identity function to mark for sort 
  ```tsx
  import {identity} from 'es-toolkit'
  const defineUnoClassName = identity
  export default defineConfig({
    shortcuts: {
      'flex-center': defineUnoClassName('flex items-center justify-center'), // add defineUnoClassName to rule option.unoFunctions
      'flex-center2': clsx('flex items-center justify-center'), // or use clsx
    }
  })
  ```

-----

U can try like [this](https://github.com/magicdawn/Bilibili-Gate/blob/v0.33.4/pnpm-workspace.yaml#L3)
```yaml
overrides:
  '@unocss/eslint-plugin': npm:@magicdawn/unocss__eslint-plugin@dev
```

I'm adding template literal support, even has variable interpolation like 
```tsx
clsx(`pr1 pl1 ${more} none-uno-class mr-1 ml-1 ${more} none-uno-class2`)

expect(output).toMatchInlineSnapshot(
  '   "clsx(`pl1 pr1 ${more} none-uno-class ml-1 mr-1 ${more} none-uno-class2`)"',
),
```
which is pretty **aggressive** and may not be accepted. 😂

magicdawn:
Thanks, I will file a documents PR reflect this change later.

Issue #4384, opened 2025-01-14
transformer-attributify-jsx blocks Astro MDX file rendering


TheJiahao:
### UnoCSS version

65.4.0

### Describe the bug

Astro build becomes extremely slow when
- a MDX file contains:
  - heading
  - TypeScript, HTML or JavaScript codeblock with bracketed content `<content>`
  - footnote
- `transformerAttributifyJsx` is loaded

The issue starts from v0.60.1, previous versions does't have the issue.

### Reproduction

1. Add MDX and UnoCSS integrations in `astro.config.ts`
2. Load `transformerAttributifyJsx` in `unocss.config.ts`
3. Add a MDX file to `src/pages` with, for example (remove "\\"):
  ```mdx
  # Heading

  \```html
  <anything>
  \```
  
  [^footnote]
  
  [^footnote]: Content
  ```
4. Run `astro build` and it won't finish

Minimal reproduction in StackBlitz, https://stackblitz.com/edit/withastro-astro-pgoyxnq9?file=src%2Fpages%2Findex.mdx
Run `pnpm install` and `pnpm astro build` to build.

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

henrikvilhelmberglund:
Likely caused by https://github.com/unocss/unocss/pull/3791.

Simon-He95:
> Likely caused by [#3791](https://github.com/unocss/unocss/pull/3791).

No, this is because the regular expression of `elementRE` is blocked in an infinite loop.

paschun:
Still appears broken with the new babel plugin. I can make a full bug report if needed

zyyv:
> Still appears broken with the new babel plugin. I can make a full bug report if needed

Please create new issue to report the problem.

TheJiahao:
The problem seems fixed in UnoCSS v66.4.2.

PR #4807, opened 2025-07-06
refactor(transformer-attributify-jsx)!: use babel plugin instead of regex plugin

zhiyuanzmj:
## Description

transformer-attributify-jsx using regex to transform by default. but regex  has many bugs and performance issue.

#4384 #4195

### Performance
#3088

Since in unocss's vscode plugin, the regex version of `transformer-attributify-jsx` would be running many times when user is typing. So i use `process.env.VSCODE_CWD` to prevent it running in vscode plugin. it's only useful in runtime plugin.

After resolve the performance issue, we can safely use the babel plugin.

### Babel plugin

Before the babel plugin was walked ast twice. so i use `@babel/traverse` walk the ast, make it faster.


zyyv:
Thanks @zhiyuanzmj , it was too difficult to maintain using regular expressions before🥹.

paschun:
This change appears to have broken integration with @astrojs/mdx

PR #4810, opened 2025-07-08
refactor(svelte-scoped): no longer use svelte preprocessing in vite plugin

### LINKED ISSUE #5141: `svelte-scoped` broken with `vite-plugin-svelte` removal of `plugin.api.sveltePreprocess` in v.7
### UnoCSS version

all

### Describe the bug

`svelte-scoped` broken with `vite-plugin-svelte` removal of `plugin.api.sveltePreprocess` in v.7 as mentioned in https://github.com/sveltejs/vite-plugin-svelte/blob/main/packages/vite-plugin-svelte/CHANGELOG.md

I need to use https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/advanced-usage.md#transform-svelte-files-with-vite-plugins to fix things. Just putting a bookmarker for now.

### Reproduction

Issue is obvious

### System Info

_No response_

### Validations

fehnomenal:
Alternative to and closes #4793
fixes #5141 

Now the transformation done by the vite plugin does not use svelte preprocessing at all. Styles are no longer inserted into the components (with `:global` modifiers) but imported from individual css files.

It works with sveltekit (which bundles styles for each entry point).
To further verify the functionality I added a new vite example that uses the plugin and this also does work without problems.


fehnomenal:
I rebased onto the latest master. This is now required to work with the wind4 preset after the release of @sveltejs/vite-plugin-svelte version 7.

henrikvilhelmberglund:
Related: https://github.com/unocss/unocss/issues/5141

examples/vite-svelte-scoped/src/App.svelte
```diff
@@ -0,0 +1,74 @@
+<script lang="ts">
```
henrikvilhelmberglund, line 1: I think this example should be updated to use runes and the new class features (https://svelte.dev/docs/svelte/class#Attributes-Objects-and-arrays)

fehnomenal: I updated to the latest svelte and migrated to runes.

Additionally, I updated to v-p-s 7.0.0 to demonstrate the changes actually work. The svelte-scoped integration now also works with the wind4 preset 🎉

henrikvilhelmberglund (approved):
LGTM, sorry for the wait.

Issue #4825, opened 2025-07-21
[presetWind4] Missing CSS variables in vue-scoped mode


keyoftime:
### UnoCSS version

66.3.3

### Describe the bug

When using in `mode: 'vue-scoped` along with `presetWind4`, the style utilities like `font-bold`, `text-red`, etc. Those variables are not defined, resulting in broken styles.

However, this issue does not occur when switching back to `presetWind3`.

### Reproduction

1.  Set up a Vue 3 + Vite Project
2. set Unocss with `mode: 'vue-scoped``
3. Use `presetWind4()`
4. Add class like `text-red` in a component's template
5. Check the generated style

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4829, opened 2025-07-24
fix(vite): styles missing or duplicated when mode is `vue-scoped`

Jungzl:
fix: #4825

### DIFF
--- a/packages-integrations/vite/src/modes/vue-scoped.ts
+++ b/packages-integrations/vite/src/modes/vue-scoped.ts
@@ -1,24 +1,17 @@
 import type { GenerateResult, UnocssPluginContext } from '@unocss/core'
 import type { Plugin, ResolvedConfig, Rollup, Update, ViteDevServer } from 'vite'
-import type { VitePluginConfig } from '../types'
 import { LAYER_MARK_ALL } from '#integration/constants'
 import { defaultPipelineExclude } from '#integration/defaults'
 import { getHash } from '#integration/hash'
 import { resolveId, resolveLayer } from '#integration/layers'
 import { getPath } from '#integration/utils'
 import { notNull } from '@unocss/core'
-import MagicString from 'magic-string'
 import { createFilter } from 'unplugin-utils'
 
-const regexScriptTags
-  // eslint-disable-next-line regexp/no-dupe-disjunctions
-  = /(?<!<!--\s*)<script(?:\s+[^=>'"/\s]+=(?:"[^"]*"|'[^']*'|[^>\s]+)|\s+[^=>'"/\s]+)*\s*(?:\/>|>([\s\S]*?)<\/script>)(?!\s*-->)/g
 const virtualModuleCSS = 'uno.css'
-const importStmt = `import "${virtualModuleCSS}";`
-const WS_EVENT_PREFIX = 'unocss:hmr'
 const HASH_LENGTH = 6
 
-export function VueScopedPlugin(ctx: UnocssPluginContext): Plugin[] {
+export function VueScopedPlugin(ctx: UnocssPluginContext): Plugin {
   let filter = createFilter([/\.vue$/], defaultPipelineExclude)
   let globalLayers: string[] = []
   let viteConfig: ResolvedConfig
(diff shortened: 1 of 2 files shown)


packages-integrations/vite/src/modes/vue-scoped.ts
```diff
@@ -5,20 +5,35 @@ import { createFilter } from 'unplugin-utils'
...
   }
 
   return {
     name: 'unocss:vue-scoped',
     enforce: 'pre',
     async configResolved() {
       const { config } = await ctx.ready
+
+      const presets = ctx.uno.config.presets
+      globalLayers = presets.some(p => p.name === '@unocss/preset-wind4')
```
antfu: I don't like this being coupled with the built-in presets, where this module should be considered agnostic.

/cc @zyyv do you have any better ideas to make this work?

packages-integrations/vite/src/modes/vue-scoped.ts
```diff
@@ -5,20 +5,29 @@ import { createFilter } from 'unplugin-utils'
...
+    ].join('\n')
   }
 
   return {
     name: 'unocss:vue-scoped',
     enforce: 'pre',
     async configResolved() {
       const { config } = await ctx.ready
+
+      globalLayers = ctx.uno.config.preflights.map(p => p.layer ?? '')
```
zyyv: I don’t recommend judging whether a layer is a global layer by layername.🧐

packages-integrations/vite/src/modes/vue-scoped.ts
zyyv: ok, I'll look into it.

packages-integrations/vite/src/modes/vue-scoped.ts
Jungzl: I think preflight layers shouldn't be scoped at least, they are themes, css variables, reset styles that take effect across the app

packages-integrations/vite/src/modes/vue-scoped.ts
```diff
@@ -1,24 +1,33 @@
...
     await ctx.ready
-    const { css } = await ctx.uno.generate(code)
+    const { css, getLayers } = await ctx.uno.generate(code)
     if (!css)
       return null
-    return `${code}\n<style scoped>${css}</style>`
+
+    return [
+      code,
+      `<style>${getLayers(globalLayers)}</style>`,
```
Jungzl: After further testing, I think it's not ideal to directly put preflight layers' css into every vue file which will generate duplicate styles. And things getting worse when `outputToCssLayers` is enabled.

For example, I duplicated `App.vue` to `CompA.vue`, `CompB.vue`, `CompC.vue` in `examples/vite-vue3` and imported them in `App.vue` and enabled `outputToCssLayers`, the build result was:

Image

which is unacceptable

packages-integrations/vite/src/modes/vue-scoped.ts
zyyv: Do you have any good ideas?
```ts
globalLayers = ctx.uno.config.preflights.map(p => p.layer ?? '')
```

I understand that some users are customizing preflight to do some hacks, which may improve the behavior of preflight and may cause some unexpected problems.

zyyv: > Do you have any good ideas?
> 
> ```ts
> globalLayers = ctx.uno.config.preflights.map(p => p.layer ?? '')
> ```
> 
> I understand that some users are customizing preflight to do some hacks, which may improve the behavior of preflight and may cause some unexpected problems.

I added an extra option to normalize this behavior, but it also needs to resolve the issue you raised below.

packages-integrations/vite/src/modes/vue-scoped.ts
zyyv: I think we can add a cache to record which global layers have been added to non-scoped <style>.

Jungzl:
does layer order matters here?

<details><summary>global mode</summary>image</details>
<details><summary>vue-scoped mode with build.cssMinify: "lightningcss" to merge layers</summary>image</details>

zyyv:
> does layer order matters here?

It should be consistent with our `layerorder` configuration.

zyyv:
@Jungzl Sorry for pushing the commit directly. I will continue to follow up on the weekend. Happy weekend!

Jungzl:
HMR DX is broken BTW☹️

PR #4833, opened 2025-07-28
feat(nuxt): introduce `wind3`,`wind4`, delete `uno` and `wind`

9romise:
related: #4413 

May need to merge #4834 first


zyyv (approved):

antfu:
We should also do `wind4`? Or maybe `wind: 3 | 4 | false`?

9romise:
> We should also do `wind4`? Or maybe `wind: 3 | 4 | false`?

Yeah, I just want to create a new PR to introduce it.

PR #4834, opened 2025-07-28
fix: run prepare

9romise:


zyyv:
We have a script in CI that tries to update automatically, but this repo restricts the robot's permissions and forbid it from automatically pushing code, so the icon collection will not be updated in time. I will consider new ways to update automatically, such as automatically sending updated pr.

/cc @ntnyq about #4364

zyyv (approved):

PR #4835, opened 2025-07-29
feat(nuxt): introduce `wind4`

9romise:
related: https://github.com/unocss/unocss/pull/4833#issuecomment-3126669823

Since the `wind` already exists, I think it would be more appropriate to add a `wind4`.


zyyv:
Focus on #4833 completion

PR #1416, opened 2022-08-15
perf(transformer-compile-class): generate same class for different sequence of same utilities

sabereen:
before:
```js
console.log(':uno: flex justify-center bg-red') // uno-x03015
console.log(':uno: flex bg-red justify-center') // uno-6kgt2n

```

now:
```js
console.log(':uno: flex justify-center bg-red') // uno-x8m0ub
console.log(':uno: flex bg-red justify-center') // uno-x8m0ub
```

PR #4845, opened 2025-08-01
fix(transformer-compile-class)!: remove not rigorous sort

### LINKED ISSUE #4840: Incorrect class priority when using transformerCompileClass
### UnoCSS version

v66.3.3

### Describe the bug

  For example btn shortcut has bg slate and If i add a new bg class, this happens. this behavior is useful for components where you'd want to get styles of original shortcuts overridden.

Image

But when I used compileclass `:uno:` in that class then the behavior got reversed 

Image

### Reproduction

[playground](https://unocss.dev/play/#html=DwEwlgbgBAtgtAZgAxQGYBsCmAPKBzAQwAcBWAPgCgooqpgAjAVwBdmB7AOygGN0CBnfgF4A5PWZd6eOACdMIEZWoc2MmAXS1gAeiatOSmtQYt2XXgOEiAXIxXWo4ydLkLDPNjCJgsUAMJ8glq6pga0FDrgEGRAA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAHU8aACyy2AQ1x0s0EBihFwwADaFTmujKQBfOFigQQcAORdcECdddIkGAA9IWHRsTS5TeEwcfCISYFIACmRUKAj6AC44AG0qVGzXCS46GGcAWjTzV3JEOClTaCzXWjRXOFsAXVkuqjoAC2gYIpg6LJTUNw4YXFcmmECYMoB3PuB5uA5SMrpLdbAAJmqqWw0obV19Q1HsrR09KAMjZzAzCys6RIBKHuF5RlGcnk5KIlCpVF9ZL8QSw2JweHx%2BBCgSIFJJpMkgag6BJNOYsgBGAB0%2B0hEwkaFwTT6rDAowA9HT6CBCf06UcJu1Pt1KLZPkggA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4IgLgTghgdgzgMwPYQLYAkyoDYgFwJTZwCmAvkA)

Start typing anything and the btn will turn grey!

### System Info

_No response_

### Validations

zyyv:
close #4840, related #1416

This PR removes non-strict sorting. Using only `sort()` to sort can cause class priority confusion. Since classes have a clear order of precedence, incorrect sorting can lead to unexpected compilation results.

e.g.

Before:
```html
<div class="bg-red bg-blue" /> 
```

Will be transformed

```html
<div class="bg-blue bg-red" /> 
```

For example, if we consider that there are shortcuts, since we don’t know how the shortcuts are named by users, I think the sorting method is extremely complicated.

Before:
```ts
{
  shourtcuts: {
    btn: 'bg-teal'
  }
}
```

```html
<div class="btn bg-blue" /> 
```

Will be transformed

```html
<div class="bg-blue btn" /> 
```

To sum up, I decided to temporarily give up the support for deduplication after sorting.

### DIFF
--- a/packages-presets/transformer-compile-class/src/index.ts
+++ b/packages-presets/transformer-compile-class/src/index.ts
@@ -107,7 +107,6 @@ export default function transformerCompileClass(options: CompileClassOptions = {
         }
 
         if (body) {
-          body = body.split(/\s+/).sort().join(' ')
           let hash: string
           let explicitName = false
 
--- a/test/transformer-compile-class.test.ts
+++ b/test/transformer-compile-class.test.ts
@@ -46,32 +46,24 @@ describe('transformer-compile-class', () => {
 </div>
     `.trim())
     expect(result.code.trim()).toMatchInlineSnapshot(`
-      "<div class="uno-pe1esh">
+      "<div class="uno-qe05lz">
       <div class="foo bar">
 
-      <div class="uno-cbgd7b foo">
-        <div class="uno-s9yxer"/>
+      <div class="uno-qlmcrp foo">
+        <div class="uno-0qw2gr"/>
       </div>"
     `)
     expect(result.css).toMatchInlineSnapshot(`
       "/* layer: shortcuts */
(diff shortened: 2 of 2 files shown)


zyyv:
I think this package should not be used for order sorting. It is only used to compile classes. It is more appropriate to leave the order sorting to `eslint-plugin` package.

Issue #4876, opened 2025-08-12
[presetWebFonts] italic: true without weights misses the italic variant on Google Fonts


yellowsink:
### UnoCSS version

66.4.2

### Describe the bug

When using the `presetWebFonts` preset, with an entry such as `{ name: "IBM Plex Serif", italic: true }`, despite this following the docs and types to create a google fonts import of the form `IBM+Plex+Serif:ital@0;1`, it actually just generates `IBM+Plex+Serif` as if `italic: true` was not specified.

If weights are specified, then the italic variant will be correctly added to the import URL.

This behaviour seems unintentional given that `weights` is an optional key, and the default weight non-italic font will be fetched without it specified.

### Reproduction

Minimal repro at https://github.com/yellowsink/uno-web-font-noital-repro

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #4878, opened 2025-08-13
fix(preset-web-fonts): support width axis and single axis config for google provider

Jungzl:
fix: #4876

### DIFF
--- a/packages-presets/preset-web-fonts/src/types.ts
+++ b/packages-presets/preset-web-fonts/src/types.ts
@@ -21,6 +21,11 @@ export interface WebFontMeta {
   italic?: boolean // ital axis
 
   /**
+   * Font width axis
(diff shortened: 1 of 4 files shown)


packages-presets/preset-web-fonts/src/types.ts
```diff
@@ -20,6 +20,11 @@ export interface WebFontMeta {
    */
   italic?: boolean // ital axis
 
+  /**
+   * Font width axis
+   */
+  width?: string | number // wdth axis
```
zyyv:
```suggestion
  width?: (string | number)[] // wdth axis
```
Is its type an array?

Jungzl: yes, array is better.

PR #4889, opened 2025-08-19
feat(core): enhance raw css utils parse with apply variants

zyyv:
UnoCSS will marked it as `RawUtil` in parsing, and it missed variants applied and its raw selectors.
e.g. `hover:foo` -> `custom css body`

```ts
rules: [
  [/^foo$/, ()=> ['custom css body']]
]
```

I wanted to enhance it in some way without affecting the original method, since it's already being used in the `keyframe`. Therefore, I chose to define a new key to customize our CSS, marking it as a normal CSSEntry for the next round of parsing.


Jungzl:
the [document](https://unocss.dev/config/rules#special-symbols) needs to be updated.

zyyv:
> the [document](https://unocss.dev/config/rules#special-symbols) needs to be updated.

Yes, I found that our docs have not updated some practical advanced usages in a timely manner. I will update the documentation after I complete the next PR.

PR #4891, opened 2025-08-20
fix(autocomplete): respect variant autocomplete template without `mutipass`

zyyv:
I think autocomplete should fully parse the existing template, regardless of the internal logic.

### DIFF
--- a/packages-engine/autocomplete/test/autocomplete.test.ts
+++ b/packages-engine/autocomplete/test/autocomplete.test.ts
@@ -152,11 +152,6 @@ describe('autocomplete', async () => {
       .toMatchSnapshot()
   })
 
-  it('should skip single-pass variants', async () => {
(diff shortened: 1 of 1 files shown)


packages-engine/autocomplete/test/autocomplete.test.ts
```diff
@@ -151,11 +152,6 @@ describe('autocomplete', async () => {
       .toMatchSnapshot()
   })
 
-  it('should skip single-pass variants', async () => {
-    expect(await ac.suggest('dark:dar')).not.toContain('dark:')
```
zyyv, line 155: I think there is no need to skip not `mutipass` variants. Such errors should be returned to the user themselves. We should provide them, and users can choose not to select them. Otherwise, all non-mutipass variant templates will become invalid.

PR #4892, opened 2025-08-20
perf(core): cache rule index to reduce query index

lghuahua:
Modify rulesDynamic and rulesStaticMap structures to cache rule indexes and reduce query indexes


zyyv:
@lghuahua  I think we can add an internal property, such as `_index`, to the `rulemeta` of the rule in the config analysis to mark the index of this `rule`.

zyyv (approved):

Issue #4886, opened 2025-08-17
`_DownloadFont` does not use `customFetch`


Indekkusu545:
### UnoCSS version

main

### Describe the bug

As the code below, it uses `fetch` from `ofetch` but not `customFetch`:

https://github.com/unocss/unocss/blob/1cd3bf61f3f077bebfc6bd4ef88e001743f51450/packages-presets/preset-web-fonts/src/local.ts#L53

This prevents the font from downloading when using a proxy.

### Reproduction

```js
{
  customFetch: (url: string) => axios.get(url, { httpsAgent: new ProxyAgent() }).then(it => it.data),
}
```

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

runyasak:
Here is my PR for implementing customFetch to resolve this issue.
Please check it out: #4894

PR #4894, opened 2025-08-20
feat: add customFetch support to LocalFontProcessor for proxy scenarios

runyasak:
fix: #4886

## Overview
This PR resolves issue #4886 where `_downloadFont` in the LocalFontProcessor was hardcoded to use the default `fetch` instead of respecting the `customFetch` option.
This prevented font downloading when using proxy configurations.

## Scope of Work
- Added `customFetch` property to `LocalFontProcessorOptions` interface
- Now uses `options?.customFetch ?? fetch` to support custom fetch implementations

## Example Usage

```typescript
const processor = createLocalFontProcessor({
  customFetch: (url: string) => 
    axios.get(url, { httpsAgent: new ProxyAgent() })
      .then(response => ({
        arrayBuffer: () => response.data
      })),
  // ... other options
})
```

---
This is my first PR for Unocss. If you have any feedback or suggestions, please let me know.

### DIFF
--- a/packages-presets/preset-web-fonts/src/local.ts
+++ b/packages-presets/preset-web-fonts/src/local.ts
@@ -40,6 +40,11 @@ export interface LocalFontProcessorOptions {
    * @default '/assets/fonts'
    */
   fontServeBaseUrl?: string
+
+  /**
+   * Custom fetch function to provide the font data.
+   */
+  customFetch?: (url: string) => Promise<any>
 }
 
 export function createLocalFontProcessor(options?: LocalFontProcessorOptions): WebFontProcessor {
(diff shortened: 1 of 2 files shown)


packages-presets/preset-web-fonts/src/local.ts
```diff
@@ -40,6 +40,11 @@ export interface LocalFontProcessorOptions {
    * @default '/assets/fonts'
    */
   fontServeBaseUrl?: string
+
+  /**
+   * Custom fetch function to provide the font data.
+   */
+  customFetch?: (url: string) => Promise<any>
```
antfu:
```suggestion
  fetch?: typeof fetch
```

PR #4895, opened 2025-08-21
fix(transformer-directives): `@apply` in grouping selectors

Jungzl:


packages-presets/transformer-directives/src/apply.ts
```diff
@@ -128,7 +131,6 @@ export async function parseApply({ code, uno, applyVariable }: TransformerDirect
         code.appendRight(childNode!.loc!.end.offset + semicolonOffset, body)
     }
   }
-  // todo: after transformered remove empty css like `selector{}`
```
Jungzl, line 131: #4626

PR #4916, opened 2025-09-09
feat(vscode): add advanced intellisense for css variable-based preset…

### LINKED ISSUE #4905: 🚀 VS Code: Compute and display final CSS values from presetWind4's default theme in hover previews

Hey team! Loving the speed and power of UnoCSS, especially with `presetWind4` :)

I've run into a significant DX ssue when using the **VS Code extension**. When I use utilities from the default `presetWind4`, the hover previews only show the underlying CSS variables and `calc()` functions, not the final computed values.

For example:
*   Hovering `p-4` shows `padding: calc(var(--spacing) * 4);`.
*   Hovering `font-bold` shows `font-weight: var(--fontWeight-bold);`.

This makes it really hard to know what styles I'm actually applying without breaking my flow. I find myself constantly trying to remember things like "what is --spacing again?" and then doing the math for calc(...) in my head. It would be a game-changer if the extension could just do this work for me.

The goal is to make the VS Code extension resolve these default theme variables and calculations on the fly, providing immediate and useful feedback.

### Suggested solution

I think the developer experience here could be massively improved if the language server resolved these values for the tooltip. It already has the resolved theme in memory, so it seems achievable.

My ideal workflow would be:
1.  The server sees a utility like `p-4` and its CSS `calc(var(--spacing) * 4)`.
2.  It looks up `var(--spacing)` in the theme and gets its value (e.g., `1rem`).
3.  It then computes the `calc()` expression, resulting in `1rem`.
4.  As a bonus, if a `rem-to-px` processor is active, it could also calculate the pixel value (`16px`).

The perfect tooltip would then show both the final value and how it was derived, which is incredibly helpful for debugging and understanding the design system. Something like this:

For `p-4`:
```css
padding-block: 1rem /* 16px */;

.py-4 {
  padding-block: calc(var(--spacing) * 4) /* 1rem = 16px */;
}
```

This would make the whole theme system feel transparent and accessible right from the editor.

### Alternative

The only alternative is the current manual process: see a `var()` in the tooltip, then either remember the value by heart, inspect it in DevTools, or dig through config files/source code and do the math. It's a real workflow killer.

### Additional context

For inspiration, the official Tailwind CSS IntelliSense extension handles this perfectly.  It would be fantastic if the UnoCSS extension could achieve a similar level of clarity.

**Screenshot 1: Spacing utility (`py-6`) from Tailwind CSS IntelliSense**
Image
 _Example from Tailwind CSS IntelliSense for a spacing utility. It shows the final value, the pixel equivalent, and the source calculation._

**Screenshot 2: Font weight utility (`font-bold`) from Tailwind CSS IntelliSense**
Image
 _Example from Tailwind CSS IntelliSense for a font-weight utility. It resolves the CSS variable to its final value._

Thanks for considering this! 

### Validations

vtestah:
This PR implements the advanced IntelliSense discussed in #4905.

It resolves CSS variables and `calc()` expressions for presets like `@unocss/preset-wind4`, showing final computed values in tooltips instead of raw `var()` functions.

**Core features:**

-   **Value Resolution:** Computes CSS variables from theme, rules, and `@property` defaults.
-   **Enhanced Tooltips:** Hover shows commented rules (with rem-to-px, color hexes). Autocomplete shows final values or shortcut definitions.
-   **Opt-in:** The feature is enabled by a `meta: { cssVarsIntelliSense: true }` flag in a preset, ensuring no impact on existing presets.

The new logic is encapsulated in `src/css-vars-intellisense/` for maintainability.

Closes #4905


ixycej12:
How long will it take to implement this preview feature?

PR #4938, opened 2025-09-29
fix(playground): add reset styles for non-preset-wind4 configurations

Jungzl:
Not inject reset styles by default


playground/src/composables/uno.ts
```diff
@@ -73,6 +74,13 @@ debouncedWatch(
           layer: customCSSLayerName,
           getCSS: () => cleanOutput(transformedCSS.value || ''),
         })
+        // @ts-expect-error ignore
+        if (!result.presets?.some(preset => preset.name === '@unocss/preset-wind4')) {
```
zyyv, line 78: I think there is a certain coupling to the exact lookup for `presetwind4`
I will merge the PR first, and I will consider improving it later, thank you

Issue #4965, opened 2025-10-28
tailwind v4: spacing not parsed right


zhihengGet:
### UnoCSS version

latest

### Describe the bug

we have an class like this  `class="bg-background group/calendar p-3 [--cell-size:--spacing(8)] [[data-slot=card-content]_&]:bg-transparent [[data-slot=popover-content]_&]:bg-transparent rounded-md border shadow-sm"`

tailiwndcss gives [--cell-size:--spacing(8)]  =  calc(var(--spacing)*8)

but unocss gives  var(--spacing\(8\))

### Reproduction

https://unocss.dev/play/#html=DwEwlgbgBAxgNgQwM5ILwCIDaBabMCmcc2SYAXvgFy5IAOCMYAdgOYAUAHAJQC66AfFGAB6cBH5A&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAHU8aACyUAvnCxQIIOAHIuuCBLp0DSJBgAekWOmwBDLgBt4mHPiIlgpABTIqFBu9ABccADaVKiRBhJcdDB6ALQhrhgG5IhwUq7QEQa0aAZwGgC6spVUdAAW0DAJMHQRQaiGCUmpdQ1NBoUwtjAprqRwgzbD0E64pBhw9QBuGFBhE8ODTq5ZVBqyIgotUTFyokoqqv4AlLLC8owsbJw8fPzXt6cKktKBJ6h0Ei2GAiAEYAHQAJg%2BqAkaFwhVqrDALQA9Cj6CAwXUUTt2mUblRqhorkggA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4XyA

```
.\[--cell-size\:--spacing\(8\)\] {
  --cell-size: calc(var(--spacing) * 8);
}

```

### Validations

[...]

PR #4968, opened 2025-10-29
fix(preset-wind4): add spacing property support and related CSS variables

Simon-He95:
close: #4965

### DIFF
--- a/packages-presets/preset-wind4/src/rules/variables.ts
+++ b/packages-presets/preset-wind4/src/rules/variables.ts
@@ -38,13 +38,22 @@ export const cssVariables: Rule<Theme>[] = [
   }],
 ]
 
+// #4965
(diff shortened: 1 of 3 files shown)


zyyv:
@Simon-He95  This isn't a good solution. I'll design a new syntax support based on this.

Simon-He95:
> @Simon-He95 This isn't a good solution. I'll design a new syntax support based on this.这不是一个好的解决方案。我将基于此设计新的语法支持。

--spacing is a special case, used in the tailwind function.
There's no need to redo such a complex project for this specific case.
https://play.tailwindcss.com/9qzXzRenie

zyyv:
https://play.tailwindcss.com/qLmkzRCx0p

I'm not sure if `spacing` is the only special key, but I want to support multiple theme keys as a function.

And it should appear not only in arbitrary value rule, but also within all bracket syntax.

PR #4981, opened 2025-11-04
fix(vite): resolve path comparison issue for config HMR

ncwhh:
Attempts to resolve the issue #4982.

From what I can observe, it appears that changes to `.js` config files aren't being detected automatically. I suspect that the issue occurs because the Vite file watcher reports file paths in different formats (relative vs absolute), but the config change detection uses direct string comparison instead of normalized path comparison. For example: `"./uno.config.js" !== "/project/uno.config.js"` causes the change to be ignored.

> Disclaimer: This bugfix was created with the help of GitHub Copilot. I’m not an expert in JavaScript, Vite, or UnoCSS, but I have a solid understanding of JavaScript and have reviewed the changes carefully. They make sense to me, but I’d really appreciate a detailed review from someone more experienced to ensure everything looks correct.

### DIFF
--- a/packages-integrations/vite/src/config-hmr.ts
+++ b/packages-integrations/vite/src/config-hmr.ts
@@ -1,5 +1,6 @@
 import type { UnocssPluginContext } from '@unocss/core'
 import type { Plugin } from 'vite'
+import { resolve } from 'node:path'
 
 export function ConfigHMRPlugin(ctx: UnocssPluginContext): Plugin | undefined {
   const { ready } = ctx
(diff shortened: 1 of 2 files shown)


packages-integrations/vite/src/config-hmr.ts
```diff
@@ -17,7 +18,8 @@ export function ConfigHMRPlugin(ctx: UnocssPluginContext): Plugin | undefined {
 
       server.watcher.add(sources)
       server.watcher.on('change', async (p) => {
-        if (!sources.includes(p))
+        const resolvedChangedPath = resolve(p)
+        if (!sources.some(source => resolve(source) === resolvedChangedPath))
```
zyyv, line 22: I'm certain it's not a `path` issue.
The problem is that the correct CSS isn't being updated to the page after `reloadConfig`. I'm currently investigating this.

ncwhh: Okay, thanks for investigating the issue! Should I then close this PR?

PR #4988, opened 2025-11-08
update filter_id  from .vue to .vue||.uvue||.nvue

tggdddd:
To enable the use of unocss in .uvue and .nvue suffix files, I modified filter_id  from .vue to .vue||.uvue||.nvue


antfu:
I am sorry we can't just add arbitrary non-standard extensions hard-coded. Instead, we could try to make them configurable/extensible so users could configure on their own. I am closing this PR for now but feel free to bring up another one. Thanks.

PR #4991, opened 2025-11-12
chore: enable trust policy

iiio2:
Since we are using pnpm 10.21, it has a very good security feature for enabling trust policy. i think maybe we can use it here.
https://pnpm.io/settings#trustpolicy


zyyv (approved):

iiio2:
Thanks @zyyv .

zyyv:
@iiio2 I think I need to restore this configuration; it's mistakenly identifying many dependencies as untrusted, when I think they're fine.

iiio2:
@zyyv , i think we can exclude some deps as well. (https://pnpm.io/settings#trustpolicyexclude)

PR #4995, opened 2025-11-17
feat: add first-class support for Marko

LuLaValva:
Marko 6 is out, and interest has been growing! Uno works great with Marko already, but it would be awesome if examples didn't need a custom config for searching through `.marko` files. This PR adds that feature alongside some Marko documentation and examples.

## Changes

- Add `.marko` to the list of files for searching, everywhere I could find
- Add a Marko example (using Marko Run, which is built on Vite)
- Include Marko examples in a few docs pages

## Limitations

- Eventually, we _will_ need a Marko extractor if we want support for the [class shorthand](https://markojs.com/docs/reference/language#shorthand-class-and-id)
  - This almost isn't necessary if the project includes Attributify, but it is definitely nice to have for completion
- The regex in `packages-integrations/vscode/src/utils.ts` doesn't support [concise mode](https://markojs.com/docs/reference/concise-syntax). I added support for optional ending tag name (`<style></>`), but concise mode is definitely a bigger deal 😅


antfu (approved):

zyyv (approved):

LuLaValva:
Thanks for the super fast merge @antfu and @zyyv! This PR wasn't just documentation, it also changed the regex in the core library to include `.marko` files. Are there any steps I missed that would have triggered a patch or minor release?

zyyv:
@LuLaValva Released in v66.5.7

PR #5010, opened 2025-12-02
feat(core): allow postprocessors to append or remove rules

Menci:
Currently the `postprocess` hook only allow to modify the existing rule object, not append more rules or remove the existing rule. However, sometimes we need to transform a rule to multiple rules. For example, color fallback:

```css
/* Before post-processing: an OKLCH color only */
.text-shadow-color-blue-300 { --un-text-shadow-color: oklch(0.8 0.11 256); }

/* After post-processing: add RGB fallback */
.text-shadow-color-blue-300 { --un-text-shadow-color: oklch(0.8 0.11 256); }
@supports not (color: oklch(0 0 0)) { .text-shadow-color-blue-300 { --un-text-shadow-color: #8ec1ff; } }
```

Since PostCSS doesn't work well with UnoCSS (`--at-apply` is handled AFTER PostCSS!), built-in post-process hook is important for such transformations.

This PR allows the `config.postprocess` hook functions to return:

* `false`: to remove the rule
* Other falsy values: to keep the rule object (maybe mutated by the function, the original behavior)
* An object: to replace the rule with a new rule object
* An array: to replace the rule with an array of rules

### DIFF
--- a/packages-engine/core/src/generator.ts
+++ b/packages-engine/core/src/generator.ts
@@ -546,13 +546,19 @@ class UnoGeneratorInternal<Theme extends object = object> {
     }
 
     return this.config.postprocess.reduce<UtilObject[]>(
-      (utilList, p) => utilList
-        .flatMap((util): UtilObject[] => {
-          const result = p(util)
-          if (!result) // Backward compatibility
-            return [util]
-          return !Array.isArray(result) ? [result] : result.filter(notNull)
-        }),
+      (utilities, p) => {
+        const result: UtilObject[] = []
+        for (const util of utilities) {
+          const processed = p(util)
+          if (Array.isArray(processed)) {
+            result.push(...processed.filter(notNull))
+          }
+          else {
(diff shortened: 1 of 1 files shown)


packages-engine/core/test/postprocess.test.ts
```diff
@@ -42,3 +42,41 @@ it('postprocess', async () => {
...
+      ['foo', { color: 'red' }],
+      ['bar', { color: 'blue' }],
+    ],
+    postprocess: [
+      (util) => {
+        if (util.selector.includes('.foo'))
+          return false
+      },
+    ],
+  })
```
zyyv: I disagree with return a `false`. If you want to disable this utility, you can use `blocklist` or use `//unocss-skip-start/end` to prevent extraction.

Menci: Removed false

zyyv (approved):
LGTM

PR #5014, opened 2025-12-07
Fix/eslint flat config prefix

### LINKED ISSUE #4412: Flat config rules seem prefixed incorrectly
### UnoCSS version

0.63.6

### Describe the bug

The rules in the flat config seem to be prefixed with `unocss/` instead of `@unocss/`. This looks like it was done deliberately since the rules are actually listed out that way in https://github.com/unocss/unocss/blob/main/packages-integrations/eslint-plugin/src/configs/flat.ts but I do not think the actual rule names should be changing when people switch over to flat config. Seems very unexpected.

At the very least if this is intentional for some reason, the docs should mention that the rule names differ by the config that is used.

### Reproduction

N/A

### System Info

_No response_

### Validations

Jack-sh1:
This PR fixes two issues:

### 1. MDX Parsing Crash (`transformer-attributify-jsx`)
- Excluded `.mdx` files from the default include list because the Babel parser cannot handle MDX syntax (e.g. frontmatter) which was causing a `SyntaxError`.
- Added a `try-catch` block around the parser to gracefully handle invalid syntax in other files without crashing the build.
- Fixes #4807

### 2. ESLint Flat Config Prefix (`eslint-plugin`)
- Renamed the plugin registration from `unocss` to `@unocss` in [flat.ts](cci:7://file:///Users/zhetaowang/Desktop/Jack/unocss/packages-integrations/eslint-plugin/src/configs/flat.ts:0:0-0:0) to be consistent with the package name.
- Updated rule prefixes to match, e.g., `@unocss/order` and `@unocss/order-attributify`.
- Fixes #4412

### DIFF
--- a/packages-integrations/eslint-plugin/src/configs/flat.ts
+++ b/packages-integrations/eslint-plugin/src/configs/flat.ts
@@ -2,10 +2,10 @@ import { plugin } from '../plugin'
 
 export default {
   plugins: {
-    unocss: plugin,
+    '@unocss': plugin,
   },
   rules: {
-    'unocss/order': 'warn',
-    'unocss/order-attributify': 'warn',
+    '@unocss/order': 'warn',
+    '@unocss/order-attributify': 'warn',
   } as const,
 }


antfu:
It was on purpose

elio-Wang:
Got it. I reverted the changes to the include option since MDX support is intentional.

Instead, I implemented a try-catch block with a regex fallback inside the transform function. This keeps MDX support active but prevents Babel from crashing on non-standard syntax.

PR #5016, opened 2025-12-08
refactor: migrate to tsdown

zyyv:
- Replace `unbuild` to `tsdown`
- Replace `tsup` to `tsdown` (`runtime` package is currently being retained, pending future refactoring into esm.)


zyyv:
I will use the attw check that comes with `tsdown` to deprecate the attw check command.

PR #5023, opened 2025-12-25
fix(inspector): prevent blinking after toggling dark

liangmiQwQ:
Fixed the blinking after toggling dark like the video below

https://github.com/user-attachments/assets/9d48eb5a-e4ed-4393-b23f-8814145eec09


userquin (approved):
Thx ❤️

Issue #4006, opened 2024-07-23
`@unocss/cli` writes cli log to stdout which should be stderr instead


nomyfan:
### UnoCSS version

0.61.5

### Describe the bug

I tried to use the CLI to generate CSS content and captured the output from stdout in my script, but I found the CLI log is also written to the stdout, which makes the output look like this
```
unocss v0.61.5
◐ UnoCSS for production...
.bg-black{--un-bg-opacity:1;background-color:rgb(0 0 0 / var(--un-bg-opacity)) /* #000 */;}
```

### Reproduction

1. File `index.html`
```html index.html
<div class="bg-black"></div>
```
2. Run the command `pnpm unocss index.html --stdout > styles.css` and inspect the CSS file

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

henrikvilhelmberglund:
Do you really need `--stdout`? By default (without `--stdout`) it outputs a `uno.css` in the same folder. You can configure the path with `-o, --out-file <file>`

If you really want to use `--stdout` you can do something like `pnpm unocss index.html --stdout | tail -n + 3 | tee styles.css`

nomyfan:
Redirecting to a file is just a help for inspecting the stdout since we cannot tell the difference between stderr and stdout in the terminal.

Of course we can do some transform(as what you do) to the stdout but it's not stable IMO, since any change to the logging might break it.

I don't know the purpose of designing the `--stdout` option. If it functions like the `-o` option, putting the CSS somewhere, I think it would be better to remove any log from it.

henrikvilhelmberglund:
I don't really use the CLI so I don't know when you would want to use `--stdout` but to me it sounds like you use it when you want to capture everything output in the terminal, basically when you want to log what is happening.

So that's why I asked why you want to use it because it seems if you don't want the extra logs you could just use the normal output mode which already doesn't have logs.

[...]

PR #5033, opened 2025-12-30
fix(cli): use stderr for cli logs for cleaner stdout output for piping

henrikvilhelmberglund:
Apparently logging inside `stdout` is annoying when using piping. This PR adds a new consola logger when using `--stdout` which should give cleaner output. It still gives you an angry error when trying to use `--stdout` and `-o` together.

To log errors/CLI logs you can do `unocss index.html 2> errors.log`.

Feel free to make changes.

fixes https://github.com/unocss/unocss/issues/4006


zyyv:
Since I've refactored some of the code, I'll be making adjustments based on the https://github.com/unocss/unocss/issues/4006#issuecomment-3698353528, and I might create a new pull request. Thank you.

PR #5041, opened 2026-01-06
feat(vscode): support spacing px comment

zhangpaopao0609:
## Summary

   - Add px value comments for `calc(var(--spacing) * N)` expressions in VSCode hover tooltips
   - When using `preset-wind4`, CSS like `width: calc(var(--spacing) * 16)` will now show helpful comments like `/* 1px * 16 = 16px */`
   - Reads `theme.spacing.DEFAULT` from UnoCSS config to calculate the actual px value
   - Supports both `px` and `rem` units for spacing values (rem values are converted using `remToPxRatio`)

## Background

   `preset-wind4` generates CSS using `calc(var(--spacing) * N)` instead of direct `rem` values. The existing `addRemToPxComment` function only handles `Xrem` format, so users couldn't see the calculated px values when hovering over utility classes.

## Example

   With `theme.spacing.DEFAULT: '1px'`:
   ```css
   .w-16 {
     width: calc(var(--spacing) * 16) /* 1px * 16 = 16px */;
   }
   ```

   With `theme.spacing.DEFAULT: '0.25rem'` (preset-wind4 default):
   ```css
   .w-16 {
     width: calc(var(--spacing) * 16) /* 0.25rem * 16 = 64px */;
   }
   ```
   
### before

image

### after
image

## Known Limitations

   - **Type casting**: Currently using `(ctx.uno.config.theme as any)?.spacing?.DEFAULT` because the core `Theme` type is generic (`Theme extends
   object = object`) and doesn't know about preset-specific properties like `spacing`. Left `// todo: types?` comments for future improvement. A better approach would be to define a shared interface or improve core type definitions.

## Test plan

   - [x] Unit tests added for `parseSpacingValue` and `addSpacingToPxComment`
   - [x] Manual testing with preset-wind4 and custom spacing values
       - [x] preset-wind4 
       image

      - [x] custom spacing values
       image

   - [x] Verify hover tooltips show correct px calculations

### DIFF
--- a/packages-integrations/vscode/src/annotation.ts
+++ b/packages-integrations/vscode/src/annotation.ts
@@ -109,14 +109,15 @@ export async function registerAnnotations(
       const remToPxRatio = docConfig.remToPxPreview
         ? docConfig.remToPxRatio
         : -1
-
       const positions = await getMatchedPositionsFromDoc(ctx.uno, doc)
+      // todo: types?
+      const spacingValue = (ctx.uno.config.theme as any)?.spacing?.DEFAULT as string | undefined
       const isAttributify = ctx.uno.config.presets.some(i => i.name === '@unocss/preset-attributify')
 
       const ranges: DecorationOptions[] = (
         await Promise.all(positions.map(async (i): Promise<DecorationOptions> => {
           try {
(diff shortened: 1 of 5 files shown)


zyyv:
To be honest, I'm really looking forward to merging #4916, where all variable values ​​will be automatically calculated.

I also really want to see more standardized code, instead of just adding features and appending parameters to function, which will be difficult to maintain later.

PR #5053, opened 2026-01-09
feat(vue-scoped): custom include & exclude patterns for sfc

### LINKED ISSUE #5051: (vue-scoped) add an options to also include none *.vue files

#3661 limits `vue-scoped` injection to `*.vue` files, but sometimes files with other extensions (such as `.md`) are also transformed into Vue SFCs ealier in the pipeline, and we hope these files be treated as vue files too.

### Suggested solution

Add a `vueInclude: string | RegExp | (string | RegExp)[] | undefined` field in the config, similar to vue's `include`, and test it against the `id` param in `transform`. A possible version might be:

```typescript
UnoCSS({
  mode: 'vue-scoped',
  vueInclude: [/\.vue$/, /\.md$/]
})
```

The default value for this field might just be `/\.vue$/`

### Alternative

Or we could do this automatically, seems like the `vue` plugin exposes an api for the `include` field

```typescript
    configResolved(config) {
      console.log(config.plugins.find((p) => p.name === 'vite:vue')?.api.include)
    },
```

### Additional context

_No response_

### Validations

- [x] Already used the [Interactive Docs](https://unocss.dev/interactive/).

illusionaries:
resolves #5051 

- Introduced `sfc` option in `ConfigBase` to control whether to treat one file as Vue SFCs
- Introduced a second filter derived from this option to control whether we should inject styles in `vue-scoped` mode

also, hmr seems to be broken in this mode...


illusionaries:
Sorry to bother, but it has been quite a while since I opened this PR, and I have been patching unocss since 66.5.9. Is there any feedback on this? 🥺

PR #5057, opened 2026-01-12
feat(core): add `allLayers` option to output all layer names

### LINKED ISSUE #5042: Ordering custom CSS layers
### UnoCSS version

66.1.2

### Describe the bug

I have some custom css in a specific layer (called `my-layer` in repro below).
I need these rules to be in a precise position amongst other UnoCSS layers.

The `layers` config key serves this purpose, but only layers that are "used" by UnoCSS classes are included so it does not work in this situation.

### Reproduction

Here's [a small UnoCSS playground](https://unocss.dev/play/#version=66.1.2&html=DwCwjABAxgNghgZwQXgERQK4IC4HsC2qAfABICmMMuAhMAPThEBQQA&config=JYWwDg9gTgLgBAbzgEwKYDNgDtUGEJaYDmANHGFKgM6owDq2yALHAL5zpQQhwDkArlggBjKlV4AoCagAekWCgwBDfgBt4aTDnyFgRABQIJccpRowqALjgBtCtVoMszfQEoAuiWNxVSgJ6oUFZGJia8IH4AtL4BULzWkQBMAAzJ3qxeJhD8MGA5ACoQuGIAMv6BVnAwUPyoEqyuEkA&css=AIGwhgngpgTgBAWwgWnNeBvAUHOB6AKjgHUYBLAFyjgGMBXAZwoHsE4BhAZU7gAtYoAGjhgAdgBM4FGGIYAzZjASw4DOgAd1iigDo4AMUVwoADzAJ1IKAC44BPDjg76TVnGy5cC0RWQB3KDIAc14KWwBWAAZIgG5HXGAwTRAIOHUARilTXwAKP15KahMQAEo4zzgAIzAaAGsgmGY6CWQaZhBFWwp%2BZRyAcjaOmAYdGChxHQAWaL6yxwBfLEWgA&options=N4XyA) that demonstrate the issue.

Notice: 
- In configuration a layer `"my-layer": 200` is declared.
- In custom CSS pane, `@layer my-layer` is used. The content is processed by UnoCSS as demonstrated by `@apply` directive.
- In output CSS, we can see that `@layer` at top of file does not list `my-layer`, so its ordering is not what I want. At the end of file you can see the custom CSS layer is actually included.

### System Info

Check out UnoCSS playground

### Validations

zyyv:
close #5042

### DIFF
--- a/docs/config/index.md
+++ b/docs/config/index.md
@@ -88,6 +88,18 @@ Specifies the name of the CSS layer the internal layer should be output to (can 
 If `undefined` is return, the internal layer name wil be used as the CSS layer name.
 If `null` is return, the internal layer will not be output to a CSS layer.
 
+#### allLayers
+
+- **Type:** `boolean`
+
+Force output all css layers, even if they are not used.
+
+For example:
+
+```css
+@layer theme, preflights, [unused-layer], default;
+```
+
 ### sortLayers
 
 - **Type:** `(layers: string[]) => string[]`
--- a/docs/config/layers.md
+++ b/docs/config/layers.md
@@ -101,6 +101,26 @@ outputToCssLayers: {
 }
 ```
 
+## Output All CSS Layers
+
+UnoCSS outputs all used CSS layers by default. If you want to force output all defined CSS layers, you can set the `allLayers` option:
+
+<!-- eslint-skip -->
+
+```ts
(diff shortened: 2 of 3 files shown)


zyyv:
@copilot What do you think of the name of this new option? Do you have any suggestions?

PR #5062, opened 2026-01-12
feat(preset-typography):  destructuring nested CSS for better compatibility

### LINKED ISSUE #4971: CSS Nesting syntax introduced in unocss@66.5 preset-typography breaks compatibility with older browsers
### UnoCSS version

66.5

### Describe the bug

After upgrading from `unocss@66.4` to `unocss@66.5`, the CSS code generated by `preset-typography` includes CSS Nesting syntax, which cannot be parsed by older browser versions, resulting in the prose styles not taking effect.  
However, some browsers still do not support [CSS Nesting](https://caniuse.com/?search=CSS+Nesting).  

CSS generated in `unocss@66.5`:

```css
/* layer: typography */
:is(.prose) {
  color: var(--un-prose-body);
  max-width: 65ch;
  :where(p):not(:where([class~='not-prose'], [class~='not-prose'] *)) {
    margin-top: 1.25em;
    margin-bottom: 1.25em;
  }
  :where([class~='lead']):not(:where([class~='not-prose'], [class~='not-prose'] *)) {
    color: var(--un-prose-lead);
    font-size: 1.25em;
    line-height: 1.6;
    margin-top: 1.2em;
    margin-bottom: 1.2em;
  }
  /* ... */
}
```

Before:

Image

Now:

Image

## Temporary solution:  

I previously imported presetTypography from unocss. By installing `@unocss/preset-typography@66.4` and importing `presetTypography` from it, the issue can be temporarily resolved. But I hope for a more long-term solution.  

Before:  
```
import { presetTypography } from 'unocss' // 66.5
```  

Now:  
```
import { presetTypography } from '@unocss/preset-typography' // 66.4
```

### Reproduction

https://stackblitz.com/edit/unocss-unocss-b3ruyphi?file=uno.config.ts

### System Info

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36

### Validations

zyyv:
close #4971


packages-presets/preset-typography/package.json
```diff
@@ -32,11 +32,14 @@
...
-    "build": "tsdown",
-    "dev": "tsdown --watch"
+    "build": "tsdown --config-loader unrun",
+    "dev": "tsdown --config-loader unrun --watch"
   },
   "dependencies": {
     "@unocss/core": "workspace:*",
     "@unocss/rule-utils": "workspace:*"
+  },
+  "optionalDependencies": {
```
antfu, line 42: I guess it should be peerDeps with peerDepsMeta optional: true

optionalDependencies was for cross platform IIRC

packages-presets/preset-typography/src/resolve.ts
```diff
@@ -60,3 +61,35 @@ export function getElements(modifier: string) {
...
+    const { transform } = await import('lightningcss')
+    const result = transform({
+      code: Buffer.from(code),
+      filename: `${selector}.css`,
+      minify: false,
+      sourceMap: false,
+      targets: {
+        safari: (16 << 16),
+      },
+    })
```
antfu, line 86: This is creative haha. I wonder if we could have an opt-in lightningcss integration at the unocss level to downgrade the entire generated css in one-go. (which might also be more efficient?)

zyyv: I'll give it a try. I want to keep the core package clean and avoid integrating LightningCSS internally. I'd like to expose the API externally. 😄

PR #5069, opened 2026-01-14
build(vscode): switch to `copy` for `babel.cjs` instead of `onSuccess` hook

9romise:


zyyv:
I think they're the same. An extra log output in onsuccess is a good thing.

PR #5105, opened 2026-02-11
fix(preset-wind4): remove leading space from space utility values for negative variant

### LINKED ISSUE #4996: `presetWind4` not support `-space-x-2`
### UnoCSS version

66.5.7

### Describe the bug

The writing of a negative sign at the beginning is not supported, but `space-x--2` is fine.

### Reproduction

https://unocss.dev/play/#html=DwEwlgbgBAxgNgQwM5ILwCIBmcCmAPKAWiQAcEYdC9CAmdAPgCgopgxZEUMxCYEAnAEYB7AHaE4wgObDCUsABcAFgFdBDYAHowTFmw7I06HnyFiJ02fOVqN23a3bxD3XgJHjJMuYtXr6WjqMWuAQ9EA&config=JYWwDg9gTgLgBAbwFBzgEwKYDNgDsMDCEuOA5gDQpxhQYDOGMAgjDFMAEYCuMwWAnpVQ16jAJIBjYnSHVaDGAHU8aACyUAvnCxQIIOAHIuuCBLp0DSJBgAekWOmwBDLgBt4mHPiIlgpABTIqFBu9ABccADaVKiRBhJcdDB6ALQhrhgG5IhwUq7QEQa0aAZwGgC6spVUdAAW0DAJMHQRQaiGCUmpdQ1NBoUwtjAprqRwgzbD0E64pBhw9QBuGFBhE8ODTq5ZVBqyIgotUTFyokoqqv4AlLLC8owsbJw8fPzXt6cKktKBJ6h0Ei2GAiAEYAHQAJg%2BqAkaFwhVqrDALQA9Cj6CAwXUUTt2mUblRqhorkggA&css=PQKgBA6gTglgLgUzAYwK4Gc4HsC2YDCAyoWABYJQIA0YAhgHYAmYcUD6AZllDhWOqgAOg7nAB0YAGLcwCAB60cggDYIAXGBDAAUKDBi0mXGADe2sGC704AWgDuCGAHNScDQFYADJ4Dc5sAACtMLKAJ5gggCMLPK2ABR2pPBIcsoAlH4WAEa0yADWTlBYqEw2yFjK3Bpw5LxxAOTllVDoYpSMYgAs3vUZ2gC%2BmsBAA&options=N4XyA

### System Info

_No response_

### Validations

kaigritun:
## Summary

Fixes #4996 - `presetWind4` not supporting `-space-x-2`

## Problem

The space utility handler was outputting margin values with a leading space:
```
margin-inline-start: calc(calc(var(--spacing) * 2) * var(--un-space-x-reverse))
```

The negative variant uses `cssMathFnRE = /^(calc|clamp|min|max)\s*\((.+)\)(.*)/` to match and negate calc expressions, but the regex requires the value to **start with** the function name. The leading space caused the regex to fail, resulting in no CSS output for negative space utilities.

## Fix

Remove the leading space from the margin value template string in `handlerSpace`:

```diff
- const value = ` calc(${v} * ...)`
+ const value = `calc(${v} * ...)`
```

## Before/After

**Before:** `-space-x-2` produces no CSS in the default layer

**After:** `-space-x-2` correctly produces:
```css
.-space-x-2{
  :where(&>:not(:last-child)){
    --un-space-x-reverse:0;
    margin-inline-start:calc(calc(calc(var(--spacing) * 2) * var(--un-space-x-reverse)) * -1);
    margin-inline-end:calc(calc(calc(var(--spacing) * 2) * calc(1 - var(--un-space-x-reverse))) * -1);
  }
}
```

## Testing

- [x] All preset-wind4 tests pass
- [x] Snapshot updated to include the now-working `-space-x-4` test case

### DIFF
--- a/packages-presets/preset-wind4/src/rules/spacing.ts
+++ b/packages-presets/preset-wind4/src/rules/spacing.ts
@@ -56,7 +56,7 @@ function* handlerSpace([m, d, s]: string[], { theme, symbols }: RuleContext<Them
   if (v != null) {
     const results = directionMap[d === 'x' ? 'inline' : 'block'].map((item, index): [string, string] => {
       const key = `margin${item}`
-      const value = ` calc(${v} * ${index === 0 ? `var(--un-space-${d}-reverse)` : `calc(1 - var(--un-space-${d}-reverse))`})`
+      const value = `calc(${v} * ${index === 0 ? `var(--un-space-${d}-reverse)` : `calc(1 - var(--un-space-${d}-reverse))`})`
       return [key, value]
     })
 
--- a/test/assets/output/preset-wind4-targets.css
+++ b/test/assets/output/preset-wind4-targets.css
@@ -1739,6 +1739,9 @@ unocss .scope-\[unocss\]\:block{display:block;}
 .view-transition-none{view-transition-name:none;}
 .field-sizing-fixed{field-sizing:fixed;}
 .field-sizing-content{field-sizing:content;}
+.-space-x-4{
+:where(&>:not(:last-child)){--un-space-x-reverse:0;margin-inline-start:calc(calc(calc(var(--spacing) * 4) * var(--un-space-x-reverse)) * -1);margin-inline-end:calc(calc(calc(var(--spacing) * 4) * calc(1 - var(--un-space-x-reverse))) * -1);}
+}
 .divide-\$variable{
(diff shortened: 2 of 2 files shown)


zyyv:
This is not the correct way to solve this problem. I will try to find a solution, but for now, please use `space-x--[num]` instead.

PR #5132, opened 2026-03-04
fix(config): enhance loadConfig to track user custom config presence

zyyv:
related #5107


kevinmarrec:
@zyyv Is there any issue regarding [**66.6.5**](https://github.com/unocss/unocss/releases/tag/v66.6.5) release ?

It does not seem that **@unocss/config** has been released on NPM as **66.6.5**, while other packages did ?

I wanted to try this fix because I got this earlier :

image

and seem related :)

Jungzl:
~~not only `@unocss/config`~~(npmx has some cache issues), but [`@unocss/config`](https://npmx.dev/package/@unocss/config) is still 66.6.4.

Issue #4989, opened 2025-11-09
presetWind4 cannot rewrite variable prefix


ocavue:
### UnoCSS version

66.5.5

### Describe the bug

To prevent CSS conflicts, I’d like to add a prefix to the generated class names and variable names.

For example, I want to change the class name `text-red-500` to `bar-text-red-500`, and variable name `--colors-red-500` to `--foo-colors-red-500`. 

It appears that there are three places where I can adjust the prefix in `presetWind4()`: `variablePrefix`, `prefix`, and `preflights.theme.process`. I’ve tried all three of them. There is the final output CSS:

```css
/* layer: properties */
@supports ((-webkit-hyphens: none) and (not (margin-trim: inline))) or
  ((-moz-orient: inline) and (not (color: rgb(from red r g b)))) {
  *,
  ::before,
  ::after,
  ::backdrop {
    --un-text-opacity: 100%;
  }
}
@property --un-text-opacity {
  syntax: '<percentage>';
  inherits: false;
  initial-value: 100%;
}
/* layer: theme */
:root,
:host {
  --baz-colors-red-500: oklch(63.7% 0.237 25.331);
}
/* layer: default */
.bar-text-red-500 {
  color: color-mix(
    in srgb,
    var(--colors-red-500) var(--foo-text-opacity),
    transparent
  );
}
@supports (color: color-mix(in lab, red, red)) {
  .bar-text-red-500 {
    color: color-mix(
      in oklab,
      var(--colors-red-500) var(--foo-text-opacity),
      transparent
    );
  }
}
```

Notice that I can still see unmodified variable names `--un-text-opacity` and `--colors-red-500` in the output CSS.

### Reproduction

[...]

PR #5142, opened 2026-03-10
fix(preset-wind4): fix option variable-prefix not working

micaiguai:
close https://github.com/unocss/unocss/issues/4989


zyyv:
@micaiguai Could you move the test file to `preset-wind4.test.ts`?

micaiguai:
> @micaiguai Could you move the test file to `preset-wind4.test.ts`?

Thank you for your correction, the test file moved.

zyyv:
cc @ocavue Modifying variables within the theme itself is somewhat tricky; however, it is still possible to implement a workaround via certain `postprocess` steps.
In my view, the current approach constitutes a relatively robust fix.

Issue #4919, opened 2025-09-12
Unocss Eslint Config will Report an Error when using Eslint `defineConfig`


DEAN-Cherry:
### UnoCSS version

66.5.1

### Describe the bug

import `unocss` and then put it on `defineConfig` area

<img width="3500" height="1317" alt="Image" src="https://github.com/user-attachments/assets/67e4ad43-42eb-441c-b8e3-db4ea0acb274" />

<img width="1344" height="2006" alt="Image" src="https://github.com/user-attachments/assets/ebbeaaf5-98f4-409e-a655-7fd12d042c0b" />

But apart from the error here, the functionality itself is normal.

### Reproduction

start a now project, install unocss and eslint, config the `@unocss/eslint-config`

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

SamuNatsu:
Still not fixed in v66.6.6, have to suppress it with `@ts-expect-error`

PR #5147, opened 2026-03-13
fix(eslint-plugin): update types for Linter compatibility

Jungzl:
close: #4919


packages-integrations/eslint-plugin/fixtures/eslint.config.ts
```diff
@@ -0,0 +1,27 @@
...
+//     rules: {
+//       'unocss/blocklist': 'error',
+//     },
+//   },
+// ])
+
+import antfu from '@antfu/eslint-config'
+import unocss from '../src'
+
+export default antfu(
```
zyyv, line 16: Why has it changed to exporting antfu's configurations?

Jungzl: I simply renamed the `eslint.config.js` to `eslint.config.ts` to ensure type correctness.

Additionally, the configuration noted in the comment above shows the type compatibility issues that users encounter when using this plugin.

zyyv (approved):

PR #5161, opened 2026-03-23
fix(preset-wind4): support nested spacing theme functions

Jack-lea:


packages-presets/preset-wind4/src/utils/handlers/handlers.ts
```diff
@@ -196,6 +208,9 @@ function bracketWithType(str: string, requiredType?: string, theme?: Theme) {
     if (base === '=""')
       return
 
+    if (theme)
+      base = replaceThemeFunctions(base, theme)
```
zyyv, line 212: Sorry, I don't think this is a very good solution.

packages-presets/preset-wind4/src/utils/handlers/handlers.ts
zyyv, line 242: My previous assessment was too narrow; I shouldn't have relied on a simple `startsWith` check. I think we should rewrite this logic to replace all theme variables enclosed within `[]`.

PR #5177, opened 2026-04-10
fix(variant-group): support data-attributes in variant group prefix

elio-Wang:
The bracket pattern in the variant group prefix regex only allowed `[&?>?:?...]` after the opening bracket, which excluded data-attribute selectors like `data-[color=red]`.

## What this PR does

- Broadens the bracket pattern in the prefix regex from `\[&?>?:?\S*\]` to `\[[^\]]*\]` to accept any characters inside brackets
- This allows variant group syntax to work with data-attributes: `data-[color=red]:(text-red font-bold)`
- Extracts inline regex to module scope to satisfy `e18e/prefer-static-regex` lint rule

## Before

```html
<!-- does NOT work -->

<p data-color="red" class="data-[color=red]:(text-red font-bold)">text</p> 
```

## After 

```html
<!-- works correctly, expands to: data-[color=red]:text-red data-[color=red]:font-bold -->

<p data-color="red" class="data-[color=red]:(text-red font-bold)">text</p>
```

### DIFF
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -2,7 +2,6 @@ import type MagicString from 'magic-string'
 import type { HighlightAnnotation } from '../types'
 import { notNull } from '../utils'
 
-/* eslint-disable e18e/prefer-static-regex */
 const regexCache: Record<string, RegExp> = {}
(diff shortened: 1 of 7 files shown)


zyyv:
Why do your pull requests always have a messy commit history? You should checkout a new branch from the latest main branch to commit code.

PR #5190, opened 2026-04-30
fix(preset-web-fonts): ensure local font files are copied to build output on initial build

### LINKED ISSUE #4146: `createLocalFontProcessor` does not store the actual web font files in `dist` during the initial build.
### UnoCSS version

0.62.4

### Describe the bug

`createLocalFontProcessor` does not store the actual web font files in `dist` during the initial build.
The download is available in the specified `public` directory and the node_modules cache, but not in `dist`.
In subsequent builds, the font files are stored in `dist`, but this means that the font is not stored in the CI environment.
This is a critical bug because I want to download Google Fonts and host them on my own server.

### Reproduction

https://stackblitz.com/edit/stackblitz-starters-hbdbqr?file=uno.config.ts

### System Info

  System:
    OS: macOS 15.0
    CPU: (11) arm64 Apple M3 Pro
    Memory: 84.73 MB / 36.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Browsers:
    Chrome: 128.0.6613.138
    Safari: 18.0

### Validations

elio-Wang:
## Summary

Fixes #4146

Frameworks like Astro determine `public/` directory contents before UnoCSS downloads font files during CSS generation. This causes font files to be missing from `dist/` on fresh/CI builds.

### Changes

- Track all downloaded font file paths via a shared `Map` registry
- Export a new `localFontsPlugin()` Vite plugin that copies font assets to the output directory during the `writeBundle` hook
- Move inline regexes to module scope to satisfy `e18e/prefer-static-regex`

### Usage

```ts
// vite.config.ts

import { localFontsPlugin } from '@unocss/preset-web-fonts/local'

export default defineConfig({

  plugins: [

    UnoCSS(),

    localFontsPlugin(),

  ],

})

### DIFF
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -2,7 +2,6 @@ import type MagicString from 'magic-string'
 import type { HighlightAnnotation } from '../types'
 import { notNull } from '../utils'
 
-/* eslint-disable e18e/prefer-static-regex */
 const regexCache: Record<string, RegExp> = {}
 const REGEX_NON_WHITESPACE = /\S+/g
 
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -2,12 +2,19 @@ import type MagicString from 'magic-string'
 import type { HighlightAnnotation } from '../types'
 import { notNull } from '../utils'
 
+/* eslint-disable e18e/prefer-static-regex */
 const regexCache: Record<string, RegExp> = {}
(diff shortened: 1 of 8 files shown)


Jack-sh1:
The failing tests are pre-existing on `main` and unrelated to this PR:

- `virtual-shared/integration/test/sort-rules.test.ts` — class sorting order
- `packages-integrations/eslint-plugin/src/rules/order.test.ts` — ESLint error counts
- `src/_preprocess/transformClasses/index.test.ts` — SVG snapshot mismatch

Verified locally: these tests also fail on a clean `main` checkout.

PR #5199, opened 2026-05-18
feat(preset-web-fonts): add onDownload callback to createLocalFontProcessor

elio-Wang:
## Description

This PR adds an `onDownload` callback option to `createLocalFontProcessor`. 

### Why?

Currently, `createLocalFontProcessor` saves downloaded fonts to the `public` directory. However, if the `vite build` command is run before the fonts are downloaded (e.g., in a fresh CI environment), Vite copies the `public` directory to `dist` **before** the fonts exist. This results in a production site missing the font files (#4146).

### How it helps

The new `onDownload` callback allows users to capture the downloaded font buffer immediately. Users can then use a custom Vite plugin to emit these fonts directly to the Rollup build output, ensuring they always appear in `dist` regardless of when the download occurs.

### Usage Example

```ts
// vite.config.ts

import UnoCSS, { createLocalFontProcessor } from '@unocss/vite' // or your package

// ...

const emittedFonts = new Map()

const fontProcessor = createLocalFontProcessor({

  onDownload(filename, buf) {

    emittedFonts.set(filename, buf)

  }

})

// Add a custom plugin to emit them

const emitFontsPlugin = {

  name: 'emit-fonts',

  generateBundle() {

    for (const [filename, source] of emittedFonts) {

      this.emitFile({ type: 'asset', fileName: `assets/fonts/${filename}`, source })

    }

  }

}

export default defineConfig({

  plugins: [

    // ...

    emitFontsPlugin,

  ],

  // ... uno css config

})


zyyv (commented):
I understand what you mean, and I think this approach is viable. It is impossible to anticipate the exact timing of a plugin's execution within a preset—this falls outside the scope of a preset. Therefore, exposing an interface to allow users to handle this themselves is a perfectly reasonable solution.

I also think we could further refine the documentation by incorporating your example and linking it to the corresponding issue.

elio-Wang:
> I understand what you mean, and I think this approach is viable. It is impossible to anticipate the exact timing of a plugin's execution within a preset—this falls outside the scope of a preset. Therefore, exposing an interface to allow users to handle this themselves is a perfectly reasonable solution.
> 
> I also think we could further refine the documentation by incorporating your example and linking it to the corresponding issue.

Thanks for the review and the fix! I'm glad the approach aligns with your vision.

Regarding the documentation, I'd be happy to add an example for onDownload and link it to #4146. Just let me know if you'd like me to update the docs in this PR as well.

zyyv (approved):

PR #5201, opened 2026-05-19
fix(eslint-plugin): resolve false positive in order-attributify rule

### LINKED ISSUE #2780: [eslint-plugin] order rule false positive

elio-Wang:
## Description

This PR fixes a false positive bug in the `@unocss/order-attributify` rule (Fixes #2780).

### Problem
The rule was incorrectly reporting errors for multi-line Vue attributes or boolean attributes, triggering `UnoCSS attributes are not ordered` even when the order was correct. This was due to how the Vue parser handles empty values and how whitespace was compared in multi-line strings.

### Fix
- **Improved attribute filtering**: Updated the logic to strictly identify valueless attributes, ensuring boolean attributes with parser-specific empty values are handled correctly.
- **Normalized whitespace**: Added `.trim()` to the sorted output to prevent comparison failures caused by formatting differences (e.g., newlines or indentation).
- **Added safety checks**: Ensured the sort action result is valid before proceeding with comparison.


Jungzl:
Sorry, I can't reproduce that issue. Perhaps it no longer needs to be fixed.
image

elio-Wang:
> Sorry, I can't reproduce that issue. Perhaps it no longer needs to be fixed. image

Hi @Jungzl, thanks for checking!

I noticed your screenshot shows `vue/html-self-closing` warning instead of 
the `@unocss/order-attributify` false positive. 

Could you confirm whether this issue might have already been fixed in main 
via another commit? The original bug (#2780) was:

- Boolean attributes like `permanent` / `rail` got parsed with empty values
  by Vue's parser
- The old filter only checked `i.value == null`, so these were excluded
- Multi-line whitespace caused the sorted comparison to fail

If the issue no longer reproduces on main, perhaps I can rebase this PR 
against the latest main and add a regression test case instead?

Happy to close this PR if the fix is already in.

Jungzl:
Yes, there's an ESLint warning, but it has nothing to do with this issue.

I simply copied the code from that issue and create a `repro.vue` under `packages-integrations/eslint-plugin/fixtures/src` to see if the issue still exists in main branch.

I don't know which commit fixed it or not.

However, since you’ve already started working on fixing this issue, I assume you’ve been able to reproduce it. Could you show me the error message? Maybe there’s something wrong with my reproduction steps.

elio-Wang:
> Yes, there's an ESLint warning, but it has nothing to do with this issue.
> 
> I simply copied the code from that issue and create a `repro.vue` under `packages-integrations/eslint-plugin/fixtures/src` to see if the issue still exists in main branch.
> 
> I don't know which commit fixed it or not.
> 
> However, since you’ve already started working on fixing this issue, I assume you’ve been able to reproduce it. Could you show me the error message? Maybe there’s something wrong with my reproduction steps.

Hi @Jungzl,

I just ran a test on the latest `main` branch, and I can confirm that the issue **no longer reproduces** on my side either.

Here is my output:
```text
test-bug.vue

  11:3   warning  Require self-closing on HTML elements (<a>)    vue/html-self-closing

  12:12  error    Newline required at end of file but not found  style/eol-last
```  
  It seems the false positive was indeed fixed in `main` by another commit (perhaps related to whitespace or parser updates).
  **However, I believe this PR is still valuable for the following reasons** :
1. Defensive Fix: My PR explicitly handles edge cases where the Vue parser returns empty objects or `true` for boolean attributes. This prevents future regressions if the parser behavior changes again.

2. Strict Normalization: Added `.trim()` to the sorted output ensures robust string comparison, guarding against potential whitespace inconsistencies from the `syncAction`.

Since I've already implemented these improvements, would you be open to merging this as a hardening measure? Or should I close the PR?

Thanks for your time!

Jungzl:
Thanks for your contribution, since there are currently no tests for this rule, could you please add the appropriate tests by referring to the other rules?

elio-Wang:
> Thanks for your contribution, since there are currently no tests for this rule, could you please add the appropriate tests by referring to the other rules?

Hi @Jungzl, I've added tests for the `order-attributify`  rule as requested. The test file covers:

Valid cases: already-sorted attributes, `class`/ `style` ignored, boolean attributes, multi-line attributes (the original false positive scenario from #2780)
Invalid cases: unsorted single-line, unsorted multi-line (regression test), boolean attributes with unsorted utilities
All 8 tests pass locally. Please review when you have a chance, thanks!

packages-integrations/eslint-plugin/src/rules/order-attributify.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    {
+      description: 'unsorted attributify attributes',
+      code: html`
+        <template>
+          <div mx1 m1 mr-1></div>
+        </template>
+      `,
+      output: output => expect(output).toMatchInlineSnapshot(`
+         "<template>
+           <div    m1 mx1 mr-1 ></div>
```
Jungzl: the extra whitespace are not desired. But perhaps this behavior has existed for a long time, and it just wasn't discovered because it hadn't been tested before.

packages-integrations/eslint-plugin/src/rules/order-attributify.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    unocss: {
+      configPath: fileURLToPath(new URL('./uno.config.ts', import.meta.url)),
+    },
+  },
+
+  valid: [
+    // Already sorted
+    html`
+      <template>
+        <div m1 mx1 mr-1></div>
```
Jungzl: add tests for empty attribute like `foo=""`

Jungzl (changes requested):

packages-integrations/eslint-plugin/src/rules/order-attributify.test.ts
elio-Wang: > add tests for empty attribute like `foo=""`

Hi @Jungzl, thanks for the review!

1. Added a test case for empty attribute values like `foo=""` as requested.
2. Also fixed the trailing whitespace issue — it was caused by a space in the `s.overwrite` replacement string (` ${sorted} ` → ` ${sorted}`). Since the tests exposed this pre-existing behavior, I corrected it and updated all snapshots.

Please let me know if anything else needs adjusting. Thanks!

PR #5202, opened 2026-05-20
fix(autocomplete): implement natural sorting for suggestions

### LINKED ISSUE #3412: VSCode Extension - Sort IntelliSense Variants

elio-Wang:
## Description

This PR improves the sorting logic for autocomplete suggestions to fix unintuitive ordering of items containing numbers (Fixes #3412).

### Problem
Previously, suggestions containing numbers were sorted strictly alphabetically (lexicographically). This caused `p-10` to appear before `p-2`, and similar issues with responsive variants like `sm:p-y-12` appearing before `sm:p-y-2`.

### Solution
Replaced the custom sorting implementation with native `String.prototype.localeCompare` using the `{ numeric: true }` option.

### Changes
- Updates `packages-engine/autocomplete/src/create.ts`.
- `processSuggestions` now performs a "natural sort" that correctly interprets embedded numbers within strings.


packages-engine/autocomplete/src/create.ts
```diff
@@ -220,20 +220,7 @@ export function createAutocomplete(uno: UnoGenerator, options: AutocompleteOptio
...
-          return -1
-
-        const numA = +(a.match(/\d+$/)?.[0] || Number.NaN)
-        const numB = +(b.match(/\d+$/)?.[0] || Number.NaN)
-        if (!Number.isNaN(numA) && !Number.isNaN(numB))
-          return numA - numB
-
-        return a.localeCompare(b)
-      })
+      .sort((a, b) => a.localeCompare(b, undefined, { numeric: true }))
```
Jungzl: Could you compare it with `Intl.Collator.prototype.compare` and see which is better?

elio-Wang: Good point. `Intl.Collator`  avoids the overhead of creating a new collator instance for every comparison, making it more performant for sorting large suggestion lists.

I've updated the PR to use a shared `Intl.Collator`  instance instead of  `localeCompare` . Thanks for the suggestion!

zyyv:
@Jack-sh1  Run `test:update` script

elio-Wang:
> @Jack-sh1 Run `test:update` script

Done. Updated snapshots and pushed.

packages-engine/autocomplete/test/__snapshots__/autocomplete.test.ts.snap
```diff
@@ -513,11 +513,11 @@ exports[`autocomplete with attributify prefix > should provide autocomplete 1`]
...
+  "u-transition-": "u-transition-all u-transition-colors u-transition-delay-75 u-transition-delay-100 u-transition-delay-150 u-transition-delay-200 u-transition-delay-300 u-transition-delay-500 u-transition-delay-700 u-transition-delay-1000",
   "u-translate-": "u-translate-0 u-translate-10 u-translate-20 u-translate-30 u-translate-40 u-translate-50 u-translate-60 u-translate-70 u-translate-80 u-translate-90",
   "u-translate-x-": "u-translate-x-0 u-translate-x-10 u-translate-x-20 u-translate-x-30 u-translate-x-40 u-translate-x-50 u-translate-x-60 u-translate-x-70 u-translate-x-80 u-translate-x-90",
-  "u-v-": "u-v-base u-v-baseline u-v-bottom u-v-btm u-v-end u-v-inherit u-v-initial u-v-mid u-v-middle u-v-revert",
-  "u-w-": "u-w-auto u-w-full u-w-lg u-w-md u-w-none u-w-prose u-w-screen u-w-screen-lg u-w-screen-md u-w-screen-sm",
+  "u-v-": "u-v-10% u-v-20% u-v-30% u-v-40% u-v-50% u-v-60% u-v-70% u-v-80% u-v-90% u-v-100%",
+  "u-w-": "u-w-2xl u-w-3xl u-w-4xl u-w-5xl u-w-6xl u-w-7xl u-w-auto u-w-full u-w-lg u-w-md",
```
zyyv: I think the utility items should be listed before the numbers; please change that.

zyyv (changes requested):

packages-engine/autocomplete/test/__snapshots__/autocomplete.test.ts.snap
elio-Wang: Done. I've updated the sorting logic to prioritize non-numeric utility items (like `auto`, `full`) before numeric/scale items (like `2xl`, `3xl`). The snapshot has been updated accordingly.

zyyv (approved):

PR #5203, opened 2026-05-21
fix(preset-wind4): correctly parse border-[color:*] arbitrary values

### LINKED ISSUE #5188: preset-wind4: typed arbitrary values for border-[color:*] are mapped to border-width instead of border-color
### UnoCSS version

66.6.8

### Describe the bug

When using Tailwind v4's typed arbitrary value syntax `border-[color:<value>]`, `preset-wind4` ignores the `color:` type hint and generates `border-width` instead of `border-color`. The type hint is effectively a no-op for `border-*`.

---

**Additional observations:**
- The type hint is fully ignored — all `border-[*:<value>]` variants unconditionally map to `border-width`
- `border-[width:2px]` produces an invalid CSS value (`border-width: width:2px`) because the hint prefix is concatenated into the value
- This is the syntax that Tailwind v4 Play CDN correctly handles (e.g. `border-[color:var(--x)]` → `border-color: var(--x)`)
- Workaround: use the arbitrary property syntax `[border-color:red]` instead

### Reproduction

**Minimal reproduction:**

```js
import { createGenerator } from "@unocss/core";
import { presetWind4 } from "@unocss/preset-wind4";

const uno = await createGenerator({ presets: [presetWind4()] });

const classes = [
  "border-[color:red]",
  "border-[color:#000]",
  "border-[color:var(--my-color)]",
  "border-[length:2px]",
  "border-[width:2px]",
];

for (const cls of classes) {
  const { css } = await uno.generate(new Set([cls]), { preflights: false });
  console.log(cls, "=>", css.replace(/\n/g, " ").trim());
}
```

**Actual output:**
```
border-[color:red]             => .border-\[color\:red\]          { border-width: red; }
border-[color:#000]            => .border-\[color\:\#000\]        { border-width: #000; }
border-[color:var(--my-color)] => .border-\[color\:var\(--my-color\)\] { border-width: var(--my-color); }
border-[length:2px]            => .border-\[length\:2px\]         { border-width: 2px; }
border-[width:2px]             => .border-\[width\:2px\]          { border-width: width:2px; }
```

**Expected output:**
```
border-[color:red]             => .border-\[color\:red\]          { border-color: red; }
border-[color:#000]            => .border-\[color\:\#000\]        { border-color: #000; }
border-[color:var(--my-color)] => .border-\[color\:var\(--my-color\)\] { border-color: var(--my-color); }
border-[length:2px]            => .border-\[length\:2px\]         { border-width: 2px; }
border-[width:2px]             => .border-\[width\:2px\]          { border-width: 2px; }
```

Note: `border-[width:2px]` produces the malformed value `border-width: width:2px` (the type hint leaks into the value string).

---

### System Info

node v22.22.2

### Validations

elio-Wang:
## Description

This PR fixes an issue where explicit color arbitrary values in borders (e.g., `border-[color:red]`) were incorrectly matched by the border width handler instead of the color handler.

### The Bug
When using typed arbitrary values like `border-[color:red]`, UnoCSS mistakenly treated `[color:red]` as a size value because `handlerBorderSize` did not explicitly ignore color-typed inputs.

### The Fix
Added a check in `handlerBorderSize` to return early if the input value starts with `[color:` or `[c:`. This ensures that these values are correctly passed to and handled by `handlerBorderColorOrSize`.

Fixes #5188

### DIFF
--- a/packages-presets/preset-wind4/src/rules/border.ts
+++ b/packages-presets/preset-wind4/src/rules/border.ts
@@ -70,6 +70,10 @@ function borderColorResolver(direction: string) {
 }
 
 function handlerBorderSize([, a = '', b = '1']: string[]): CSSEntries | undefined {
+  // Explicitly typed color arbitrary values (e.g., [color:red]) should be handled by the color handler
+  if (b.startsWith('[color:') || b.startsWith('[c:'))
+    return
+
   const v = h.bracket.cssvar.global.px(b)
   if (a in directionMap && v != null)
     return directionMap[a].map(i => [`border${i}-width`, v])


zyyv:
This is not the correct way to resolve this. @Jack-sh1 

By the way, you should branch off from the latest `main` branch in the remote repository to carry out your development; this helps prevent your code from falling too far behind and causing conflicts.

image

PR #5102, opened 2026-02-10
feat(language-server): add `@unocss/language-server`

Jungzl:
Introduces an UnoCSS Language Server and refactor VSCode integration.

Then Zed integration will be much easier.

[unocss-66.6.0-lsp-preview.zip](https://github.com/user-attachments/files/25205388/unocss-66.6.0-lsp-preview.zip)

PR #5227, opened 2026-06-19
feat(zed): Add Zed extension

oddcelot:
## Description

Adds first-class [Zed](https://zed.dev) support for UnoCSS via a new editor extension, backed by the **official** `@unocss/language-server`. To make underlining of matched utilities work in editors that don't use VSCode-style decorations, this also adds a semantic-tokens capability to the language server (see https://github.com/bajrangCoder/zed-unocss/issues/6).

### Why

There is an existing community extension (`bajrangCoder/zed-unocss`), but it relies on a fork of the language server and only attaches to `TSX`. This extension uses the official server and additionally attaches to plain `TypeScript` and `CSS`, so Lit components (`class="..."` inside `` html`...` `` in `.ts` files) and stylesheets get completion, hover, and color previews too.

As currently i work heavily with Zed as main editor I (and I assume a few other people) would appreciate an official extension with official LSP support 😄

>[!NOTE]
> The Zed registry `unocss` ID is currently owned by the community extension. Publishing to the registry will require coordinating an ID takeover with the Zed maintainers — this PR only lands the extension in the monorepo.
We could also do `unocss-official` as ID, but i would hope for just replacing the other one. 

## What's in this PR

**1. `@unocss/language-server` — semantic tokens provider** (`packages-integrations/language-server`)
- New `semanticTokens` capability emitting every matched utility as a single custom `unocss` token type, letting any LSP client style them (an editor-agnostic equivalent of the VSCode `underline` decoration).
More Info on that here 👉 https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens and https://zed.dev/docs/semantic-tokens
- **Off by default** (`unocss.semanticTokens: false`) — VSCode renders its own underline, so clients like Zed opt in.
- Re-requests tokens on settings/config change (guarded for clients without refresh support).
- Encoding logic (single-line filter, sort, overlap removal for variant matches) is extracted into a pure `buildSemanticTokensData()` with unit tests.

**2. New Zed extension** (`packages-integrations/zed`)
- Rust/wasm extension (`zed_extension_api` 0.7.0) that fetches `@unocss/language-server` from npm on first use (does not vendor it), with an offline fallback and a `binary` override for local builds.
- Forwards Zed settings under the `unocss` namespace; maps Zed language names to the server's LSP `languageId`s.
- Attaches to HTML, CSS, JavaScript, TypeScript, TSX, Vue, Svelte, Astro, Markdown, PHP, ERB.
- Ships a `README` and example `.zed/settings.json` documenting the opt-in underline setup.

**3. Repo config**
- ESLint ignores the Cargo-managed files (`Cargo.toml`, `target/`) — first Rust crate in the monorepo.

## Testing

- `pnpm test` — new `semanticTokens.test.ts` (6 cases) + existing language-server suite pass.
- `pnpm typecheck`, `pnpm lint`, `pnpm build` — all green.
- Zed extension builds clean (`cargo build` / `cargo clippy`, `wasm32-wasip2`).

>[!TIP]
> You can test this with Zed directly by cloning my fork and this branch where I've set up a small demo, instructions are here:  https://github.com/oddcelot/unocss/blob/demo/zed-extension/packages-integrations/zed/DEMO.md
Branch would be `demo/zed-extension`
Should look like this: 
>image

## Notes for reviewers

- The Rust crate is **not** wired into `pnpm build`/CI; it's built by Zed's own toolchain. `Cargo.lock` is committed (extension convention); `target/` and `*.wasm` are gitignored.
- `extension.toml` `authors` is currently the original contributor — happy to adjust to the org/maintainers' preference.

---

I specifically did not split this into two PRs with one being the `semanticTokens` because this is a feature that would currently be only used by Zed. Possibly VSCode could use this too, but out of scope for now.


oddcelot:
I know that the section with the current Zed extension needs adjusting, but I would wait adjusting this until this would be also available in the Zed extension repo 😅
https://unocss.dev/integrations/zed

edit: maybe a preview of this wouldn't hurt though, as it still can dropped

packages-integrations/vscode/package.json
```diff
@@ -145,7 +145,7 @@
     }
   },
   "scripts": {
-    "build": "tsdown",
+    "build": "nr update && tsdown",
```
oddcelot, line 148: Noticed that the Netlify preview was failing on `@unocss/vscode`'s build (`Could not resolve './generated/meta'`) 

I think this might happen when pnpm install is fully cached and skips the `prepare` script that generates `src/generated/meta.ts`.

Did this happen to other PRs or is something off here? 🤔

Jungzl:
Since #5102, I had the thought that we may need a separate repo to maintain zed extension according to https://zed.dev/docs/extensions/developing-extensions#directory-structure-of-a-zed-extension.

Could you push your related codes to [the new repo](https://github.com/unocss/zed)?

oddcelot:
> Could you push your related codes to [the new repo](https://github.com/unocss/zed)?

Sure thing :)
Just have to move a few things around and check again if the refs resolve correctly then.
Probably a new PR with just the changes for the semantic tokens would be good here, so I'll close this one when ready.

oddcelot:
closing this in favor of https://github.com/unocss/zed/pull/1 and https://github.com/unocss/unocss/pull/5229 :)

PR #5229, opened 2026-06-23
feat(language-server): semantic tokens for matched utilities

oddcelot:
Hey! 👋 This is the language-server side of the new [UnoCSS Zed extension](https://github.com/unocss/zed/pull/1). It started life in #5227 alongside the extension
Now the extension has its own repo, and this PR brings the `@unocss/language-server` changes it relies on.

### What's in here

- **Semantic tokens** — the server now emits each matched utility as a custom `unocss` token, so any LSP client can style them (an editor-agnostic take on VSCode's underline). The encoding lives in a pure, unit-tested `buildSemanticTokensData()` that handles the single-line / non-overlapping rules LSP expects, and tokens refresh on config/settings changes.
- **On by default** — `semanticTokens` defaults to `true` (set `unocss.semanticTokens: false` to opt out). This flips the off-by-default from #5227; emitting by default is the nicer baseline, and VSCode is unaffected since it draws its own underline.
- **Graceful config fallback** — if a config file exists but fails to load (e.g. `vite.config.ts` imports a plugin you haven't installed), the server falls back to a default context so basic utilities still work, instead of going dark.
- **Color completion fix** — color docs now show comma-form `rgba(r, g, b, a)` for consistent previews.
- **Docs** — the Zed integration page now points at the standalone extension and covers the opt-in underline setup.

### Testing

See the the counterpart PR in the Zed extension repo 😄 

2026-06-23 um 19 16 16
grafik

### DIFF
--- a/docs/integrations/zed.md
+++ b/docs/integrations/zed.md
@@ -4,7 +4,7 @@ title: Zed Extension for UnoCSS
 
 # Zed Extension
 
-The official [UnoCSS extension for Zed](https://github.com/unocss/unocss/tree/main/packages-integrations/zed), maintained in the UnoCSS monorepo. It runs the official [`@unocss/language-server`](https://www.npmjs.com/package/@unocss/language-server) for completion, hover, color previews and matched-utility underlining.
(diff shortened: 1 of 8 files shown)


antfu:
It seems you are using agents to create this PR - I would happy to have meaningful contributions but please do your review and don't let LLM speak to you. Thanks.

PR #5231, opened 2026-06-24
perf: batch token parsing

TheAlexLichter:
This PR limits UnoCSS token parsing to batches of 4,096, reducing microtask pressure for large token sets while keeping things as-is for smaller amounts of tokens.

## Bench

In the npmx client build, 11 runs (alternating pairs) improved the median from 8.840s to 8.510s, with a mean improvement of **4.39%** and the candidate winning every pair.

Running the given bench script on a Cloudflare Worker and testing improved from 2.636s to 1.885s median across nine runs, resulting in a **31.3%** mean improvement (for the token parsing operation).


packages-presets/preset-wind4/src/preflights/property.ts
```diff
@@ -26,6 +26,7 @@ export function property(options: PresetWind4Options): Preflight<Theme> | undefi
         return
 
       const css = Array.from(trackedProperties.entries())
+        .sort(([a], [b]) => a.localeCompare(b))
```
TheAlexLichter, line 29: This is necessary to guarantee byte-by-byte equivalence. Due to the token batching it wouldnt be the case otherwise

zyyv (approved):

antfu (approved):

PR #5232, opened 2026-06-24
perf(core): prefilter dynamic rules

TheAlexLichter:
This PR builds a (conservative) regex prefilter for dynamic rules, allowing tokens that cannot match any compatible rule to skip the full dynamic-rule scan while safely falling back for prefixes, backreferences, lookarounds, and stateful regexes. 

On the real npmx corpus, median UnoCSS generation improved from 782.3ms to 536.3ms (31.5% faster), while the full Nuxt client build improved from 6.124s to 5.908s on average (3.53% faster).

### DIFF
--- a/packages-engine/core/src/config.ts
+++ b/packages-engine/core/src/config.ts
@@ -11,6 +11,39 @@ export function resolveShortcuts<Theme extends object = object>(shortcuts: UserS
   })
 }
 
+function createDynamicRuleFilters<Theme extends object>(rules: readonly DynamicRule<Theme>[]) {
+  const filterSources = new Map<string, string[]>()
+  const fallback: DynamicRule<Theme>[] = []
+
+  for (const rule of rules) {
+    const [matcher, , meta] = rule
+    // These patterns cannot be safely embedded into a larger alternation.
+    if (meta?.prefix || matcher.global || matcher.sticky || /\\(?:[1-9]|k<)|\(\?</.test(matcher.source)) {
+      fallback.push(rule)
+      continue
+    }
+    const sources = filterSources.get(matcher.flags) ?? []
+    sources.push(`(?:${matcher.source})`)
+    filterSources.set(matcher.flags, sources)
+  }
+
+  try {
+    return {
+      fallback,
+      filters: Array.from(
+        filterSources,
+        ([flags, sources]) => new RegExp(sources.join('|'), flags),
+      ),
+    }
+  }
+  catch {
+    return {
+      fallback: [...rules],
+      filters: [],
+    }
+  }
+}
+
 const __RESOLVED = '_uno_resolved'
 
 /**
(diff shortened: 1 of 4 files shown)


packages-engine/core/src/config.ts
```diff
@@ -11,6 +11,39 @@ export function resolveShortcuts<Theme extends object = object>(shortcuts: UserS
...
+    sources.push(`(?:${matcher.source})`)
+    filterSources.set(matcher.flags, sources)
+  }
+
+  try {
+    return {
+      fallback,
+      filters: Array.from(
+        filterSources,
+        ([flags, sources]) => new RegExp(sources.join('|'), flags),
```
zyyv, line 35: Is it really worthwhile to merge multiple regular expressions into one super-large regular expression?

packages-engine/core/src/config.ts
```diff
@@ -11,6 +11,39 @@ export function resolveShortcuts<Theme extends object = object>(shortcuts: UserS
...
+      ),
+    }
+  }
+  catch {
+    return {
+      fallback: [...rules],
+      filters: [],
+    }
+  }
```
zyyv, line 44: Excessive regular expressions cannot guarantee that this method will pre-filter and return under ideal conditions. If it returns to the initial starting point, then I think this pre-filtering is meaningless.

packages-engine/core/src/config.ts
TheAlexLichter: Yes it is! 

The reason is that most apps are "reject-heavy" because Uno’s extractor sees lots of normal code text, not just CSS classes. 

That means that testing each dynamic rule separately would still repeat the expensive part: Trying all the dynamic regexes with around O(number of dynamic rules) tests. One combined regex can cheaply answer “could any dynamic rule match this?” and skip the full rule loop for obvious non-classes.

TheAlexLichter: One example from the npmx.dev codebase:

files scanned: 563
unique extracted tokens: 44,395
dynamic rules: 258
static rule hits: 159
prefilter groups: 1
fallback rules: 1
combined regex size: ~8.9 KB

non-static variant candidates: 44,238
prefilter rejected: 37,033
prefilter admitted: 7,205

Matcher-loop:

(current) baseline dynamic regex attempts: 10,946,487
(with this PR) prefilter dynamic regex attempts: 1,429,006
combined-filter tests: 44,238

baseline median: ~414ms
prefilter median: ~83ms

packages-engine/core/src/config.ts
TheAlexLichter: That is correct but fallbacks are there as an escape hatch.

If a rule cannot be safely included, or if building the combined regex fails, we keep correctness and fall back to the old behavior.

In normal presets this should not be the common path at all.
For my tests with npmx.dev, only the prefixed icon rule ended up in fallback mode, while ~37k extracted non-class tokens were still rejected by the prefilter. 

So the optimization is not meaningless unless most rules would fall into fallback.

Jungzl (approved):
I'm OK with this PR, the perf improvements look promising

PR #5238, opened 2026-06-28
fix(transformer-directives): track @apply tokens in pseudo selectors

### LINKED ISSUE #5076: Directives transformer does not generate classes for @apply inside a state pseudo-class
### UnoCSS version

66.6.0

### Describe the bug

If the classes in `@apply` are only added to a pseudo-class selector (e.g. hover, focus) then Uno will not generate the needed properties for the CSS to work. The classes need to be set outside the psuedo-class for the properties to be generated.

This is made evident with ring color. If the ring and ring color are not set outside the pseudo-class selector, then they will not show up.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-6wtnutck?file=src%2Fstyle.css

### Steps to reproduce

1. Add the transformer directive to your uno config
2. Add `@apply ring ring-red-500` to the hover state of an element in the document
3. Hover over the element

### Expected result

A red ring shows up on hover.

### Actual result

No CSS is shown. Chrome devtools says `--un-ring-color is not defined`.

### System Info

System: MacOS
Browser: Chrome
Framework: Tanstack Start

### Validations

LevDomasnih:
### What changed

Register utilities used by `@apply` with the UnoCSS token set when the directives transformer runs through the plugin context.

### Why

When `@apply` appears inside a pseudo-class selector, for example `.btn:hover { @apply ring ring-red-500; }`, the transformer expands the declarations but the generator may not see the referenced utilities as scanned tokens. As a result, generated utility CSS such as ring color variables can be missing when those classes only exist inside `@apply`.

### Validation

- `pnpm vitest run test/transformer-directives.test.ts`
- `pnpm exec eslint packages-presets/transformer-directives/src/apply.ts packages-presets/transformer-directives/src/index.ts packages-presets/transformer-directives/src/transform.ts packages-presets/transformer-directives/src/types.ts test/transformer-directives.test.ts`

Closes #5076

### DIFF
--- a/packages-presets/transformer-directives/src/apply.ts
+++ b/packages-presets/transformer-directives/src/apply.ts
@@ -13,13 +13,13 @@ export async function handleApply(ctx: TransformerDirectivesContext, node: Rule)
   await Promise.all(
     node.block.children.map(async (childNode) => {
       if (childNode.type === 'Raw')
-        return transformDirectives(code, uno, options, filename, childNode.value, childNode.loc!.start.offset)
+        return transformDirectives(code, uno, options, filename, childNode.value, childNode.loc!.start.offset, ctx.tokens)
       await parseApply(ctx, node, childNode)
     }).toArray(),
   )
 }
 
-export async function parseApply({ code, uno, applyVariable }: TransformerDirectivesContext, node: Rule, childNode: CssNode) {
+export async function parseApply({ code, uno, applyVariable, tokens }: TransformerDirectivesContext, node: Rule, childNode: CssNode) {
   const original = code.original
 
   let body: string | undefined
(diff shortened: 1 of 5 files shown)


zyyv:
No, when using the UnoCSS context within a transformer, token extraction happens automatically; there is no need for you to manually add tokens. 
https://github.com/unocss/unocss/blob/b48cca818e616a5a16f9e0d88380f11f1b8454de/virtual-shared/integration/src/context.ts#L44-L67
see line:62 

If you believe there is an issue, please provide a test PR that demonstrates the error so we can verify it.

zyyv:
This is not the correct solution. The problem arises because the `@property` layer is resolved within its scope, not at the top level, causing the CSS variables to fail to take effect.

PR #5239, opened 2026-06-28
fix(transformer-compile-class): update explicit shortcuts

### LINKED ISSUE #4926: [transformer-compile-class] Updating class names during dev causes duplicate class name error
### UnoCSS version

v66.5.1

### Describe the bug

When using a pre-named compiled class via `:uno-{name}:` during dev mode, updating the class will cause a duplicated class name error.

> [plugin:unocss:transformers:pre] Duplicated compile class name "uno-foo". One is "font-bold text-green text-lg" and the other is "font-bold text-lg". Please choose different class name or set 'alwaysHash' to 'true'.

Refreshing the page does not fix it and requires a full restart of the dev server.

### Reproduction

https://stackblitz.com/edit/unocss-unocss-94wngyll?file=src%2FApp.vue

### System Info

System: Windows 11/WSL
Browsers: Firefox
Framework: Nuxt/Vue

### Validations

LevDomasnih:
### What changed

Allow explicitly named compile classes to update across separate transformer runs.

### Why

In dev mode, changing a pre-named compiled class such as `:uno-foo:` reused the same generated class name while the stored shortcut body changed. The transformer treated that as a duplicate-name conflict and required a dev server restart.

This keeps the duplicate-name error for conflicting explicit names within the same transform pass, but updates the stored shortcut and clears the cached generator entry when the same explicit class changes across transforms.

### Validation

- `pnpm vitest run test/transformer-compile-class.test.ts`
- `pnpm exec eslint packages-presets/transformer-compile-class/src/index.ts test/transformer-compile-class.test.ts`

Closes #4926

### DIFF
--- a/packages-presets/transformer-compile-class/src/index.ts
+++ b/packages-presets/transformer-compile-class/src/index.ts
@@ -73,7 +73,7 @@ export default function transformerCompileClass(options: CompileClassOptions = {
     alwaysHash = false,
   } = options
   // #2866
-  const compiledClass = new Set()
+  const compiledClass = new Map<string, string>()
 
   // Provides backwards compatibility. We either accept a trigger string which
   // gets turned into a regexp (like previously) or a regex literal directly.
@@ -89,7 +89,8 @@ export default function transformerCompileClass(options: CompileClassOptions = {
       if (!matches.length)
         return
 
-      const size = compiledClass.size
+      let hasChanges = false
+      const currentClasses = new Map<string, string>()
       for (const match of matches) {
         let body = ((match.length === 4 && match.groups)
           ? expandVariantGroup(match[3].trim())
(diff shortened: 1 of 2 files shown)


packages-presets/transformer-compile-class/src/index.ts
```diff
@@ -121,18 +122,26 @@ export default function transformerCompileClass(options: CompileClassOptions = {
...
+          if (compiledClass.get(className) !== body) {
+            compiledClass.set(className, body)
+            uno.cache.delete(className)
+            uno.blocked.delete(className)
+            hasChanges = true
+          }
           replacements.unshift(className)
-          if (options.layer)
-            uno.config.shortcuts.push([className, body, { layer: options.layer }])
+          const shortcut = options.layer ? [className, body, { layer: options.layer }] : [className, body]
```
zyyv, line 139:
```suggestion
          const shortcut: Shortcut = [className, body, options.layer ? { layer: options.layer } : undefined]
```

packages-presets/transformer-compile-class/src/index.ts
zyyv, line 1:
```suggestion
import type { Shortcut, SourceCodeTransformer } from '@unocss/core'
```

zyyv (changes requested):

zyyv:
@LevDomasnih Please take a look.

PR #5242, opened 2026-07-03
ci: add CVE Lite dependency audit workflow

sonukapoor:
This adds a CVE Lite CLI dependency audit workflow to scan unocss for vulnerable dependencies on every push, pull request, and weekly on Mondays.

CVE Lite CLI is an OWASP Lab Project that reads the pnpm lockfile directly without installing packages, so it runs fast and requires no setup beyond checkout. It queries the OSV database for known vulnerabilities and classifies each finding as a direct or transitive dependency, with copy-and-run fix commands where a safe version is available.

A scan of the current lockfile found 17 findings: 1 critical, 7 high, 5 medium, and 4 low. The workflow is configured with `fail-on: high`, which means CI will block on high and critical findings until they are resolved. SARIF output is enabled and uploaded to GitHub Code Scanning so findings appear in the Security tab alongside other code scanning results.

More details on the tool and its output at https://owasp.org/cve-lite-cli.

### DIFF
--- a/.github/workflows/cve-lite.yml
+++ b/.github/workflows/cve-lite.yml
@@ -1,0 +1,36 @@
+name: CVE Lite dependency audit
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 6 * * 1'
+
+permissions:
+  contents: read
+  security-events: write
+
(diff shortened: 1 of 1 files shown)


zyyv:
Sorry, we don't need it temporarily.

Issue #5007, opened 2025-11-30
Example of SvelteKit cause ReferenceError in Safari when importing uno.css in +layout.svelte


utakotoba:
### UnoCSS version

66.5.9

### Describe the bug

When using Svelte 5 (Runes) with SvelteKit, following the official examples (`examples/sveltekit`) to `import 'uno.css'` directly inside `<script>` in `+layout.svelte` causes a runtime Error on Safari (macOS/iOS), breaking the runtime reactive system of Svelte.

Safari console shows: ReferenceError: Cannot access uninitialized variable. Almost the same bug like #618.

<img width="1243" height="1056" alt="Image" src="https://github.com/user-attachments/assets/d0dda262-5572-4bdd-ba5e-68c9ce4055e7" />

In that case, the runes system is broken since the runtime of Svelte throws. So, as in the image, the counter button won't work anymore.

This problem only happens in the `sveltekit` example; for others like `sveltekit-preprocesser`, `sveltekit-scoped` they are fine since they do not use that `import 'uno.css'` in the root layout component.

Importing `uno.css` in client-side hooks may solve this problem in my local testing; check [SvelteKit Hooks](https://svelte.dev/docs/kit/hooks) for details. But it also causes a flicker in the first load since it does not load early enough.

### Reproduction

Just run the `sveltekit` example, then open Safari.

### System Info

System: macOS 15.7.2
Browser: Safari 26.1 (20622.2.11.119.1)

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/unocss/unocss/blob/main/.github/CONTRIBUTING.md).
- [x] Check that there isn't [already an issue](https://github.com/unocss/unocss/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/unocss/unocss/discussions) or join our [Discord Chat Server](https://chat.antfu.me).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

caelyreth:
Only occurs in the dev server (via `vite dev`); for the production build, everything works fine on Safari.

henrikvilhelmberglund:
Safari doesn't support top-level await properly and it's unclear when it will (since https://github.com/WebKit/WebKit/pull/48409 was closed). Probably a good idea to just use Svelte Scoped instead.

dmvjs:
[...]

PR #5257, opened 2026-07-27
fix(examples): disable hmrTopLevelAwait in the sveltekit example

Nic-Polumeyv:
Fixes the sveltekit example on WebKit engines that haven't shipped the 242740 fix. Mechanism in https://github.com/unocss/unocss/issues/5007#issuecomment-5095849100. Verified the served `/__uno.css` no longer contains a top-level await.


henrikvilhelmberglund (approved):
LGTM (although should probably be removed when fixed browser is shipped)

PR #5265, opened 2026-08-17
feat(reset, preset-wind4): port new tailwind changes

kidonng:
Port two changes from Tailwind to the `tailwind-v4` reset:

https://github.com/tailwindlabs/tailwindcss/pull/20292
https://github.com/tailwindlabs/tailwindcss/pull/20318

Note the latter changes the default font stack to fix inappropriate use of `system-ui`, which could change the appearance on existing sites depending on their configuration.


zyyv:
https://github.com/unocss/unocss/blob/main/packages-presets/preset-wind4/src/preflights/reset.ts

Could you please help update `preset-wind4` as well?

zyyv (approved):

PR #5267, opened 2026-08-17
fix(svelte-scoped): use SvelteKit assets path for the global stylesheet link

### LINKED ISSUE #5264: svelte-scoped broken with base path
### UnoCSS version

66.6.8 (but doesn't matter)

### Describe the bug

The global styles integration of the svelte-scoped plugin uses `base` from the vite config to generate an absolute path to the global stylesheet. It looks like `<link href="/_app/immutable/assets/unocss-svelte-scoped-global.C9ljiVS4.css" rel="stylesheet" />`.

Since https://github.com/sveltejs/kit/pull/15936 (released as v2.65.0) the vite base is overwritten to `./` if the `kit.paths.relative` option is omitted or set to `true`. This result in the vite config having `/` as base and thus the generated path as above does not contain the base path.

### Reproduction

This is reproducible in the `examples/sveltekit-scoped`: Run `pnpm up` to get an effected sveltekit version, `pnpm build`, `pnpm preview`. Observe in the dev tools the 404 to the global stylesheet.

### System Info

_No response_

### Validations

lazerg:
The link to the global stylesheet was built from the vite base, which is `/` in the SvelteKit server build regardless of `kit.paths`. Setting `kit.paths.base` therefore produced a root-absolute href and a 404.

It now concatenates SvelteKit's `assets` template variable instead, the same one `%sveltekit.assets%` compiles to, so the href follows whatever paths config the app uses.

Verified against `examples/sveltekit-scoped` with `paths.base: '/sub'`; output at the default root config is unchanged.

closes #5264


fehnomenal (approved):
I like the simplicity but absolutely don't love how dependent on the generated code this is. 

I tried injecting a css import into the client entry, so vite would handle the css on its own but couldn't get a FOUC away...

henrikvilhelmberglund:
Does this work with Sveltekit 3 RC as well? Otherwise I would prefer to wait and fix it everywhere instead.

lazerg:
Yes, it should. SvelteKit's server codegen still uses the same `assets` variable substitution for `%sveltekit.assets%` in the current `main` branch (source of the 3.0.0-next.x releases), same as in v2. I checked `packages/kit/src/core/sync/write_server.js` in sveltejs/kit and the pattern is unchanged. So this fix should keep working on SvelteKit 3 RC too.

fehnomenal:
I'm one of the original authors of the svelte-scoped integration and so partly responsible for the current mess of putting placeholders in place and having different flows in the plugin for dev/build with sveltekit or plain svelte.

I'm thinking about getting rid of all of this and simply instruct the user to `import 'uno.css'` (or `uno-global.css`?) into their root layout. Then we could get rid of the hoops and reuse the flow from the basic vite integration and build the virtual file from just the reset and other global styles.

This is probably for another PR, but what do you think @henrikvilhelmberglund @jacob-8 ?

henrikvilhelmberglund:
I'm fine with `import 'uno.css'`,

henrikvilhelmberglund (approved):
This is fine but should probably be improved later, but can merge this now as it solves an issue.

PR #5270, opened 2026-08-17
perf(core): index selector merge candidates

hannoeru:
## Summary

Make CSS selector merging scale linearly for selector-heavy rule groups.

## Changes

- Replace forward candidate scans with an index keyed by declaration body and selector shape.
- Preserve sorted emission order, `noMerge` isolation, raw rules, parent rules, scoping, and minified output.
- Add a 250-selector regression case that covers scoped ordering and merge isolation.

## Performance impact

The previous forward candidate search can become `O(n²)` for `n` matching-body rules. The indexed merge pass is expected `O(n)` per sorted rule group, with constant-time candidate lookup.


hannoeru:
Benchmark result: https://github.com/unocss/unocss/actions/runs/32045091355/attempts/1#summary-95431100906

zyyv (approved):

Issue #5164, opened 2026-03-26
Unrelated assets in Webpack are being loaded and transformed improperly since 66.5.10


grant-progress:
### UnoCSS version

66.5.10

### Describe the bug

When using the UnoCSS Webpack plugin in a project which processes non-JS or binary assets, assets are either breaking the build with `Module parse failed` or (in case of binaries) being silently corrupted by re-encoding.

I have looked over the `66.5.9`->`66.5.10` changelog and suspect it's due to the removal of the `transformInclude` filter: https://github.com/unocss/unocss/compare/v66.5.9...v66.5.10#diff-e7e726a717257e779ade4c95121e72326df7cff1fb26bbe98b86b4eef6618299L50

In our case the failure was particularly pernicious - it rewrote a WOFF2 font file into a corrupted form (I assume by reading and re-encoding the file) which was not caught until fonts broke (the build succeeded).

I have spent a few days trying to hone in on the problem and am confident it's directly related to this change in UnoCSS' Webpack plugin version. It's reproducible on the latest version at time of writing (66.6.7).

It's possible I've misconfigured my Webpack config here, but this same general configuration worked fine prior to 66.5.10 and I don't see any details in docs which suggest an error on my part.

### Reproduction

https://github.com/grant-progress/unocss-webpack-repro

Included in this repro is a WOFF2 font file being imported from CSS as an asset module. But the build actually breaks first on the HTML file processed by HTMLWebpackPlugin.

Run `npm run build`.

```
ERROR in   Error: Child compilation failed:
  Module parse failed: Unexpected token (1:0)
  File was processed with these loaders:
   * ./node_modules/html-webpack-plugin/lib/loader.js
   * ./node_modules/unplugin/dist/webpack/loaders/transform.js
   * ./node_modules/unplugin/dist/webpack/loaders/load.js
  You may need an additional loader to handle the result of these loaders.
  > <!doctype html>
  | <html lang="en">
  |   <head>
  ModuleParseError: Module parse failed: Unexpected token (1:0)
  File was processed with these loaders:
   * ./node_modules/html-webpack-plugin/lib/loader.js
   * ./node_modules/unplugin/dist/webpack/loaders/transform.js
   * ./node_modules/unplugin/dist/webpack/loaders/load.js
  You may need an additional loader to handle the result of these loaders.
  > <!doctype html>
  | <html lang="en">
  |   <head>
[...]

PR #5275, opened 2026-08-22
fix(webpack): filter load/transform hooks away from binary assets

### LINKED ISSUE #5274: fix(webpack): load hook without filter still corrupts imported binary assets (PNG)
## Description

Since `@unocss/webpack` 66.5.10+, importing local binary assets (PNG/JPG/fonts) via webpack can silently corrupt them when the UnoCSS webpack plugin is enabled.

This is the remaining part of #5164 after #5204: #5204 added `transform.filter` to exclude `.html`, but **`load` is still an unfiltered hook**, so unplugin registers a global webpack `load` loader (`type: javascript/auto`) that also runs on PNG and other binary modules. **`transform` still matches binary extensions** (only `.html` was excluded).

When the UnoCSS hook handler returns early (not a virtual CSS module), unplugin passes the original source back as a string. Without `raw = true`, webpack treats binary as text and the PNG magic bytes `0x89 0x50 0x4E 0x47` become UTF-8 replacement chars (`ef bf bd 50 4e 47`).

## Expected

Imported PNG files in the webpack output should be byte-identical to the source files.

## Actual

Output PNG headers are corrupted; images fail to render (e.g. in Taro weapp builds).

## Reproduction

1. Use `@unocss/webpack` 66.7.x+ with webpack 5
2. Enable UnoCSS webpack plugin
3. `import logo from './logo.png'` in an entry module
4. Build and inspect `dist/logo.png` with `xxd` — header is corrupted

## Root cause

- unplugin expects plugin authors to declare `load.filter` / `transform.filter` so loaders only run on intended modules: [unjs/unplugin#524](https://github.com/unjs/unplugin/issues/524)
- #5204 scoped `transform` to exclude `.html` only; **`load` was never scoped**, and **binary assets still hit `transform`**
- Global `raw = true` on unplugin loaders ([unjs/unplugin#560](https://github.com/unjs/unplugin/pull/560)) was rejected by maintainers; the correct fix is **narrow filters**, not global `raw`

## Proposed fix

1. **`load.filter`** — only virtual UnoCSS CSS modules (`__uno*.css`)
2. **`transform.filter`** — also exclude common binary extensions (png/jpg/gif/webp/svg/ico/fonts/…)

PR: #5275

## References

- #5164 (closed; transform-only fix)
- #5204 (merged)
- unjs/unplugin#524

ozyl:
Fixes #5274

## Problem

Since 66.5.10, `@unocss/webpack` registers unplugin webpack `load` / `transform` loaders without sufficiently narrow filters. When those loaders touch PNG/fonts/etc., unplugin passes binary through as strings (no `raw`), corrupting file headers — see #5164 and unjs/unplugin#524.

#5204 only excluded `.html` from `transform`; **`load` was never filtered**, and **binary extensions were still matched by `transform`**.

## Fix

1. **`load.filter`** — only virtual UnoCSS CSS modules (`__uno*.css`)
2. **`transform.filter`** — also exclude common binary asset extensions (png/jpg/gif/webp/svg/ico/fonts/…)

Restores scoped `load` handling for virtual CSS modules (remaining gap after #5204), using unplugin's filter API.

## Test

Added `test/webpack-png.test.ts` + minimal webpack fixture: import a PNG with `@unocss/webpack` enabled, assert emitted file is byte-identical to source (`0x8950…`).

## Notes

- Does **not** use global `raw = true` on unplugin loaders (rejected in unjs/unplugin#560)
- After merge, downstream projects (e.g. Taro + `@unocss/webpack`) can remove webpack-chain workarounds like `unocss-scope-load` / `restore-assets`


packages-integrations/webpack/src/unplugin.ts
```diff
@@ -88,13 +90,22 @@ export function unplugin<Theme extends object>(configOrPath?: WebpackPluginOptio
...
-      async load(id) {
-        const layer = await getLayer(ctx, id)
-        if (!layer)
-          return
+      // #5164: unfiltered load hook registers unplugin's global load loader on all modules,
+      // corrupting binary assets (unjs/unplugin#524). Only match virtual CSS module ids.
+      load: {
+        filter: {
+          id: {
+            include: [/[/\\]__uno(?:_.*?)?\.css(\?.*)?$/],
```
zyyv: https://github.com/unocss/unocss/blob/a441ef4d8b14a20c0b3551383ae1b1e96940c0d2/virtual-shared/integration/src/context.ts#L118

UnoCSS support custom the virtual file prefix.

ozyl: Good catch — updated in f931fe0.

`load.filter` no longer hardcodes `__uno`. After config is ready it uses `RESOLVED_ID_WITH_QUERY_RE` from `getVMPRegexes()`, so a custom `virtualModulePrefix` still matches the virtual CSS modules. Also added a webpack fixture case with `virtualModulePrefix: 'custom_uno'`.

zyyv (approved):

PR #5278, opened 2026-08-26
fix(presets): fix utilities misidentifying arbitrary length values as colors

s97712:
This PR fixes an issue where multiple utility rules (including border, outline, decoration, stroke, and text) misidentified arbitrary length/size values with units (e.g. outline-[3px], stroke-[5px], text-[7px]) as colors instead of sizes, causing them to fail to generate the correct CSS width/size declarations.

### DIFF
--- a/test/assets/output/preset-wind4-targets.css
+++ b/test/assets/output/preset-wind4-targets.css
@@ -618,7 +618,7 @@
 .border-s-width-\$variable{border-inline-start-width:var(--variable);}
 .border-t-width-3{border-top-width:3px;}
 .b-block-size-\$variable{border-block-start-width:var(--variable);border-block-end-width:var(--variable);}
-.border-\[\#124\]{border-width:#124;}
+.border-\[\#124\]{border-color:color-mix(in oklab, #124 var(--un-border-opacity), transparent) /* #124 */;}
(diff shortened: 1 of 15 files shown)


test/assets/output/preset-wind4-targets.css
```diff
@@ -618,14 +618,14 @@
 .border-s-width-\$variable{border-inline-start-width:var(--variable);}
 .border-t-width-3{border-top-width:3px;}
 .b-block-size-\$variable{border-block-start-width:var(--variable);border-block-end-width:var(--variable);}
-.border-\[\#124\]{border-color:color-mix(in oklab, #124 var(--un-border-opacity), transparent) /* #124 */;}
+.border-\[\#124\]{border-width:#124;}
```
zyyv: This generated incorrect css.

zyyv:
> This PR fixes an issue where multiple utility rules (including border, outline, decoration, stroke, and text) misidentified arbitrary length/size values with units (e.g. outline-[3px], stroke-[5px], text-[7px]) as colors instead of sizes, causing them to fail to generate the correct CSS width/size declarations.

Based on the example you described, I tested it but didn't find it being parsed as color css. Could you provide a playground link?

PR #5285, opened 2026-08-29
fix(extractor-arbitrary-variants): extract quoted attribute selectors behind class and pseudo-class prefixes

### LINKED ISSUE #4026: [🐞BUG]: `[&_.xxx_[stroke="#ccc"]]` doesn't work in real project but does work in interactive doc
### UnoCSS version

latest

### Describe the bug

---
Update:

Below will work:
`[&_[stroke="#ccc"]]:whatever`
`[&_tag[stroke="#ccc"]]:whatever`

Below won't work
`[&_.class[stroke="#ccc"]]:whatever`
`[&_.class_[stroke="#ccc"]]:whatever`
`[&_.class_tag[stroke="#ccc"]]:whatever`

---

I have a selector: `[&_.recharts-cartesian-grid_line[stroke="#ccc"]]:hidden`
It's working well in the interactive doc, but in the real project it won't generate the CSS rule.
`[&_.recharts-cartesian-grid_line[stroke=#ccc]]:hidden` without `"` will generate the rule however it's not matching the element.

The real usage is not like that, but just for minimal prod purposes. As long as I can match that `line` element with `[stroke="#ccc"]`, I can do whatever I want then.

image
image

### Reproduction

https://stackblitz.com/edit/vitejs-vite-noeatr?file=src%2FApp.tsx

### System Info

_No response_

### Validations

### LINKED ISSUE #4638: [Wind4] Difficult selectors not working
### UnoCSS version

last

### Describe the bug

For example selector `[&_svg:not([class*='size-'])]:size-4` works on tw4, but not with preset wind4

### Reproduction

UnoCSS variant:

https://stackblitz.com/edit/nuxt-starter-99z2c3jb?file=components%2Fui%2Fbutton%2Findex.ts

Tailwind variant:

https://v4.shadcn-vue.com/#carousel

### System Info

------------------------------
- Operating System: Windows_NT
- Node Version:     v22.14.0
- Nuxt Version:     3.17.1
- CLI Version:      3.25.0
- Nitro Version:    2.11.11
- Package Manager:  pnpm@10.10.0
- Builder:          -
- User Config:      compatibilityDate, future, devtools, css, features, app, modules, eslint, shadcn, colorMode
- Runtime Modules:  @nuxt/eslint@1.3.0, @unocss/nuxt@66.1.0, template-copyright@1.0.1, shadcn-nuxt@2.1.0, @nuxtjs/color-mode@3.5.2
- Build Modules:    -
------------------------------

### Validations

giaBaoJS:
Closes #4026
Closes #4638

## Reproduction

```html
<div class='[&_.recharts-cartesian-grid_line[stroke="#ccc"]]:hidden'></div>
<button class="[&_svg:not([class*='size-'])]:size-4"></button>
```

Neither produces a rule. The second one is verbatim from current shadcn/ui `button.tsx`, which is why #4638 hits every shadcn user on preset-wind4.

The reported pattern in #4026 is oddly specific, and it is worth restating because it points straight at the cause:

| token | before |
|---|---|
| `[&_[stroke="#ccc"]]:hidden` | works |
| `[&_tag[stroke="#ccc"]]:hidden` | works |
| `[&_.class[stroke="#ccc"]]:hidden` | no rule |
| `[&_svg:not([class*='size-'])]:size-4` | no rule |

## This is an extraction bug, not a parsing one

The engine already handles all four. Calling `parseToken` directly on the failing tokens returns correct CSS today, on `main`, with no changes:

```
[&_.recharts-cartesian-grid_line[stroke="#ccc"]]:hidden
  -> .\[\&_\.recharts-cartesian-grid_line\[stroke\=\"\#ccc\"\]\]\:hidden .recharts-cartesian-grid line[stroke="#ccc"] { display:none }

[&_svg:not([class*='size-'])]:size-4
  -> .\[\&_svg\:not\(\[class\*\=\'size-\'\]\)\]\:size-4 svg:not([class*='size-']) { width:...; height:... }
```

So the token never reaches the engine. `defaultSplitRE` splits on quotes, so any candidate containing a quoted attribute selector is shredded, and `quotedArbitraryValuesRE` in `extractor-arbitrary-variants` is the only thing that rescues it. That regex cannot describe these two shapes:

- the leading character class `[\w&:[\]-]` has no `.`, so matching restarts after it. `[&_.class[stroke="#ccc"]]:hidden` is salvaged only as `class[stroke="#ccc"]]:hidden`, which matches no rule.
- there is no `(`, and the trailing `\]?` cannot consume the `)]` that closes `:not(...)`, so the shadcn token is not matched at all.

## The change

```diff
-  = /(?:[\w&:[\]-]|\[\S{1,64}=\S{1,64}\]){1,64}\[\\?['"]?\S{1,64}?['"]\]\]?[\w:-]{0,64}/g
+  = /(?:[\w&:[\].(-]|\[\S{1,64}=\S{1,64}\]){1,64}\[\\?['"]?\S{1,64}?['"]\](?:\)?\])?[\w:-]{0,64}/g
```

Two additions: `.` and `(` in the leading class, and an optional `)` before the closing `]` in the tail. All `{1,64}` / `{0,64}` bounds are unchanged.

The tail is `(?:\)?\])?` rather than something like `[)\]]{0,4}` on purpose. A greedy version also eats the trailing `)` of unrelated code such as `foo(bar['baz'])`, which changes 7 tokens the current regex already extracts. Requiring the `)` to be followed by `]` keeps the change additive for every one of them.

## What I did not add, and why

I measured each character separately over this repo's own sources (tracked `.ts`/`.tsx`/`.vue`/`.js`/`.svelte`/`.html` files, excluding `dist/`, the 1 MB ReDoS asset, and the test file this PR edits, so the counts are not inflated by my own fixtures):

| leading class | fixes | raw matches | distinct tokens |
|---|---|---|---|
| `[\w&:[\]-]` (current) | neither issue | 208 | 154 |
| `+ .` | #4026 only | 208 | 154 |
| `+ . (` (this PR) | both | 586 | 187 |
| `+ . ( )` | both, plus `:nth-child(2)` | 1050 | 549 |

Worth being explicit about the shape of that:

- `.` is free. It fixes #4026 completely and does not add a single match on this corpus.
- `(` is where the entire cost sits. It is required for #4638, since `:not(` cannot otherwise be crossed.
- `)` in the leading class is **not** included. It would only buy selectors whose parentheses close before the attribute bracket, like `[&_li:nth-child(2)[data-x="y"]]:mt-2`, which neither issue reports, and it nearly triples distinct candidate tokens (187 to 549). That did not look like a trade worth making, but it is a one character change if you disagree.

If the `(` cost is unwelcome, dropping it still fixes #4026 at zero cost and leaves #4638 open. I would rather you make that call than have it hidden in the diff.

## Cost of the extra matches

The 33 extra distinct candidates are things like `:global([type="A"]):not` from `svelte-scoped` CSS test fixtures. They cost extraction time and nothing else. I checked every one of them against the generator:

**0 of the 55 added tokens produce any CSS.** So this widens the candidate set, and cannot emit a spurious rule.

The other consumer of this regex, `getMatchedPositions` in `match-positions.ts`, gates on `plain.has(match[0])`, so unmatched extras are dropped there too. It gains correct editor highlighting for these tokens.

### ReDoS

`test/assets/regex-dos.ts` (1,010,775 bytes), 9 runs each, same machine and process:

| | min | max | median |
|---|---|---|---|
| before | 166 ms | 189 ms | 175 ms |
| after | 157 ms | 178 ms | 171 ms |

No measurable change, which is expected since every quantifier bound is preserved.

## Tests

Two added to `extractor-arbitrary-variants.test.ts`, both against `splitCodeWithArbitraryVariants` rather than the regex, so they assert extraction behaviour:

- one covering both issue tokens. Reverting only the regex fails it with the token shredded into `"[&_.recharts-cartesian-grid_line[stroke="`, `"#ccc"`, `"]]:hidden"`.
- one covering `[&_[stroke="#ccc"]]:hidden` and `[&_tag[stroke="#ccc"]]:hidden`, which already worked. It passes before and after this change, and fails independently if the leading class is otherwise damaged (removing `&` from it fails this test and not the first one).

Full suite before: 97 files, 678 passed, 3 skipped. After: 97 files, 680 passed, 3 skipped.

## Known limitation

Parentheses that close before the quoted attribute bracket are still unmatched, as noted above. One level of `:not(...)`/`:is(...)`/`:where(...)` wrapping the bracket, which is the shadcn shape, is handled.


zyyv (approved):
Thank you for your thorough investigation. Based on your description, I think adding a period `.` `(` is more reasonable.

PR #5289, opened 2026-08-31
fix(core): accept @ and * in variant group bodies, and stop bracket prefixes over-matching

rajanpanth:
Two defects in `makeRegexClassGroup`, both caused by a character class. Found by fuzzing `expandVariantGroup`, not from a report.

### 1. `@` and `*` variants inside a group body kill the whole group

```js
expandVariantGroup('hover:(@sm:text-red bg-blue)')
// -> 'hover:(@sm:text-red bg-blue)'   (unchanged)
expandVariantGroup('hover:(*:text-red bg-blue)')
// -> 'hover:(*:text-red bg-blue)'     (unchanged)
```

The prefix class is `[!@*<~\w+:_-]` and accepts `@` and `*`, but the body class is `[~!<>\w\s:/\,%#.$?-]` and does not. A single unmatchable character makes the entire group fail to match, so nothing expands and the raw text is emitted.

The consequence is worse than a missed expansion. Running `hover:(@sm:text-red bg-blue)` through `presetWind4` produces **0 bytes of CSS**, so `bg-blue` is lost too, not just the container-query utility. `@sm:(text-red bg-blue)` generates correct `@container (min-width: 24rem)` output, and the un-grouped `hover:@sm:text-red hover:bg-blue` is fine, so this only bites inside a group body.

Both are first-class variants (`preset-wind4/src/variants/container.ts` registers `@`, `variants/children.ts` registers `*` and `**`) and both already work as a group *prefix*, so accepting them as a prefix but not inside a body looks unintended rather than deliberate.

### 2. A group nested inside an arbitrary variant captures the wrong prefix

```js
expandVariantGroup('[&>a]:([&>b]:(p-1 p-2))')
// -> '[&>a]:([&>b]:p-1 [&>a]:([&>b]:p-2'
```

Note the unbalanced `(` and the duplicated `[&>a]:(`. The bracket alternative in the prefix is `\[&?>?:?\S*\]`, and `\S*` runs straight past the closing `]`, so on this input the prefix captures `[&>a]:([&>b]` instead of `[&>b]`. Bounding it to non-bracket characters fixes it.

This needs an arbitrary variant on *both* the outer and inner group. `[&>a]:(p-1 p-2)`, `[&>a]:(hover:(p-1 p-2))` and `hover:([&>b]:(p-1 p-2))` are all correct today and stay correct.

### Tests

Two cases added to `packages-engine/core/test/variant-group.test.ts`, covering `@`, `*`, `**` in a body and two-level arbitrary-variant nesting.

Mutation tested: reverting only `variant-group.ts` while keeping the tests fails exactly those two and leaves the other 14 passing.

Full run of `test/`, `packages-engine/core/test/` and `virtual-shared/`: 78 files, 512 passed, 3 skipped, no regressions.

### Note on overlap

This touches the same file as #5287 but a different part of it, the regex rather than the rebuild loop, so they should not conflict. Happy to rebase whichever lands second.


zyyv:
Thank you for fixing these details!

zyyv (approved):

PR #5305, opened 2026-09-04
fix(transformer-directives): honor the deprecated varStyle option

rajanpanth:
### Description

`resolveApplyVariables` computes `applyVariable` from `varStyle` and then overwrites it unconditionally on the very next line:

```ts
if (applyVariable === undefined) {
  if (varStyle !== undefined)
    applyVariable = varStyle ? [`${varStyle}apply`] : []
  applyVariable = ['--at-apply', '--uno-apply', '--uno']   // always wins
}
```

So the `varStyle` option has no effect at all. It is deprecated but still public and typed `false | string`:

```
varStyle: '--my-'  ->  ["--at-apply","--uno-apply","--uno"]   expected ["--my-apply"]
varStyle: false    ->  ["--at-apply","--uno-apply","--uno"]   expected []
```

`varStyle: false` is documented as the way to turn off custom-property `@apply` entirely, and it currently leaves all three defaults active.

### Change

The default assignment moves into an `else`, so it only applies when `varStyle` is not provided. `applyVariable` still takes precedence over `varStyle`, which is the documented order.

### Tests

A new `apply variables` block covering the defaults, `varStyle` as a string, `varStyle: false`, and `applyVariable` winning over `varStyle`. Reverting only the source and keeping the tests fails with `expected [ Array(3) ] to deeply equal [ '--my-apply' ]`; restored, the transformer-directives suite passes 78/78 and eslint is clean.

### DIFF
--- a/packages-presets/transformer-directives/src/transform.ts
+++ b/packages-presets/transformer-directives/src/transform.ts
@@ -15,7 +15,8 @@ export function resolveApplyVariables(options: TransformerDirectivesOptions) {
   if (applyVariable === undefined) {
     if (varStyle !== undefined)
       applyVariable = varStyle ? [`${varStyle}apply`] : []
-    applyVariable = ['--at-apply', '--uno-apply', '--uno']
+    else
+      applyVariable = ['--at-apply', '--uno-apply', '--uno']
   }
   return toArray(applyVariable || [])
 }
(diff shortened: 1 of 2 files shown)


zyyv:
This deprecated flag has been around for a long time. I plan to remove it in the next version.

PR #5312, opened 2026-09-06
fix(scripts): close default-export bypass in dist verification

hannoeru:
## Problem

The postbuild dist checker (`scripts/dist-verify.ts`) could report success while a forbidden import was present. Its forbidden-`require()` rule was skipped for any file containing `as default`, so a declaration like:

```ts
import uno = require("@unocss/core")
export { uno as default }
```

passed validation. The script also carried obsolete commented-out checks, misleading `cjsFiles` naming, and had no test coverage.

## Changes

- Forbidden `require()` of `@unocss/*` in CJS declaration output (`.d.ts` / `.d.cts`) now always fails the build — the `as default` exemption is removed.
- Scope intentionally limited to **declaration outputs**: runtime `.cjs` may `require()` ESM-only packages because Node ≥ 22.12 (the minimum engine) supports `require(esm)`. The nuxt and webpack integrations rely on this; an earlier draft that scanned runtime output produced false positives in CI.
- Removed obsolete commented-out implementation; renamed `cjsFiles` → `cjsDeclarationFiles` so logs match what is scanned.
- Direct-execution guard now compares via `pathToFileURL()`, fixing a silent no-op on Windows (`argv[1]` is `C:\...`, never equal to a `/C:/` URL pathname).
- Added `test/dist-verify.test.ts`: spawns the real checker against temporary fixtures and asserts pass/fail exit codes, including the default-export case.

## Validation

- `pnpm exec vitest run test/dist-verify.test.ts` — 3/3 passed
- Full checker run against a complete local build: `Dist files verify passed`
- CI green across ubuntu/macos/windows matrices

### DIFF
--- a/scripts/dist-verify.ts
+++ b/scripts/dist-verify.ts
@@ -1,42 +1,31 @@
 import fs from 'node:fs/promises'
+import { join } from 'node:path'
 import process from 'node:process'
 import { glob } from 'tinyglobby'
 
-export async function verifyDist() {
-  const cjsFiles = await glob([
-    'packages-*/*/dist/**/*.d.ts',
+// UnoCSS packages are ESM-only, so CJS output must not require them
+const forbiddenCjsImport = /\brequire\(\s*['"]@?unocss(?:\/core)?['"]\s*\)/
+
+export async function verifyDist(root = process.cwd()) {
+  const cjsOutputFiles = await glob([
+    'packages-*/*/dist/**/*.cjs',
     'packages-*/*/dist/**/*.d.cts',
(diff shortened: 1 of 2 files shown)


hannoeru:
The first CI run failed because the rewritten checker scanned runtime `.cjs` output, where `require()` of ESM-only packages is valid on Node ≥ 22.12 (`require(esm)`) — this produced false positives for the nuxt and webpack integrations (b4f583a08 restricts the check to declaration outputs, matching the original intent while keeping the default-export fix).

PR #5327, opened 2026-09-09
fix(core): support nested arbitrary variant brackets

### LINKED ISSUE #5326: preset-wind4 emits invalid CSS for `/oklab` color modifiers inside shortcuts
### UnoCSS version

66.10.1

### Describe the bug

With `presetWind4`, using the color interpolation modifier `/oklab` inside an arbitrary selector variant nested in a shortcut makes the Vite production build emit invalid CSS.

The warning is:

```
[unocss] unmatched utility "[&[aria-selected=true]]:(bg-accent/9/oklab" in shortcut "rail-tab"
```

The generated stylesheet then fails Lightning CSS minification:

```
[vite:css-post] SyntaxError: [lightningcss minify] Unexpected token CloseParenthesis
```

The failure is reproducible in [CNSeniorious000/macaron-agent-skills#158](https://github.com/CNSeniorious000/macaron-agent-skills/pull/158), at commit [f624c01](https://github.com/CNSeniorious000/macaron-agent-skills/commit/f624c01ad440e1161ad782f5161e445d9fe9f835). The relevant configuration is [webui/uno.config.ts](https://github.com/CNSeniorious000/macaron-agent-skills/blob/f624c01ad440e1161ad782f5161e445d9fe9f835/webui/uno.config.ts#L110-L120):

```ts
presets: [presetWind4(), presetIcons()],
theme: {
  colors: {
    accent: 'light-dark(#a65e2b, #c99076)',
    'accent-strong': 'light-dark(#a64f4f, #cb7676)',
  },
},
shortcuts: {
  'rail-tab':
    '... [&[aria-selected=true]]:(bg-accent/9/oklab text-accent-strong/oklab)',
},
```

The project uses Vite 8.2.2 with the default Lightning CSS minifier. Lint, type checking, and tests pass; only the production CSS build fails.

### Expected behavior

Either the shortcut variant should be parsed and compiled to valid CSS, or it should be rejected without emitting malformed CSS. `vite build` should not fail during CSS minification.

### Reproduction

Run `deno install` and `deno task build` from the linked commit above.

Related reports: [#5251](https://github.com/unocss/unocss/issues/5251) covers a different Lightning CSS parse failure, and [#5290](https://github.com/unocss/unocss/pull/5290) fixes a separate CSS-variable fallback bug.

CNSeniorious000:
Fixes #5326.

Expand nested attribute selectors in both variant-group prefixes (`[&[open]]:(...)`) and bodies (`hover:([&[open]]:...)`). These groups previously remained unexpanded, which could send malformed tokens into CSS generation and fail Lightning CSS minification. Remove ambiguous prefix matching that caused excessive backtracking on repeated arbitrary variants.

Share bracket-aware tokenization across expansion, collapse, and rule sorting so quoted or bracketed values containing spaces stay intact. Preserve literal brackets and the existing collapse whitespace normalization. Prefix matching supports one inner bracket level; deeper nesting remains unexpanded and is covered by tests.

Regression coverage includes string and MagicString expansion, nested groups followed by siblings, collapse round-trips, sorting, repeated prefixes, and the exact `light-dark(...)` / `bg-accent/9/oklab` reproduction.

Validation:
- Built `@unocss/core` and the relevant presets and transformer.
- `vitest run packages-engine/core/test virtual-shared/integration/test/sort-rules.test.ts test/transformer-variant-group.test.ts`: 171 tests passed across 31 files.
- ESLint passed for variant-group and sort-rules source/tests.
- Shortcut tests compare Wind4 CSS against expanded utilities and verify Lightning CSS minification.

### DIFF
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -13,11 +13,11 @@ export function makeRegexClassGroup(separators = ['-', ':']) {
     // are valid inside a group, not just as its prefix. A single unmatchable
     // character makes the whole group fail to match, so omitting them silently
     // left the group unexpanded.
-    // The bracket alternative stops at the matching `]`, allowing a nested
-    // attribute bracket, rather than stopping at the first inner `]`.
-    // Without this, `[&[aria-selected=true]]:(...)` is left unexpanded and the
-    // unmatched group body is later emitted as malformed CSS.
-    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?(?:[^\\s\\[\\]]|\\[[^\\[\\]]*\\])*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
+    // Both the prefix and body must consume nested attribute brackets as a unit.
+    // Stopping at the inner `]` leaves `[&[open]]:(...)` or `hover:([&[open]]:...)`
+    // unexpanded, and the unmatched group can later produce malformed CSS.
+    // Quoted body values can contain literal brackets, as in `content-['[']`.
+    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?(?:[^\\s\\[\\]]|\\[[^\\[\\]]*\\])*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[(?:[^\\[\\]'\"]|'(?:\\\\.|[^'\\\\])*'|\"(?:\\\\.|[^\"\\\\])*\"|\\[[^\\[\\]]*\\])*\\])+?)\\)(?!\\s*?=>)`, 'gm')
   regexCache[key].lastIndex = 0
   return regexCache[key]
 }
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -4,7 +4,55 @@ import { notNull } from '../utils'
 import { escapeRegExp } from './escape'
 
 const regexCache: Record<string, RegExp> = {}
-const itemRE = /(?:[^\s'"]|'(?:\\.|[^'\\])*'|"(?:\\.|[^"\\])*")+/g
+function splitVariantGroupBody(body: string) {
+  const items: { index: number, value: string }[] = []
+  let start = -1
+  let depth = 0
+  let quote = ''
+  let escaped = false
+  const push = (end: number) => {
+    if (start >= 0)
+      items.push({ index: start, value: body.slice(start, end) })
+    start = -1
+  }
+  for (let i = 0; i < body.length; i++) {
+    const char = body[i]
+    if (quote) {
+      if (escaped)
+        escaped = false
+      else if (char === '\\')
+        escaped = true
+      else if (char === quote)
+        quote = ''
+      continue
+    }
+    if (char === '\'' || char === '"') {
+      if (start < 0)
+        start = i
+      quote = char
+      continue
+    }
+    if (char === '[') {
+      if (start < 0)
+        start = i
+      depth++
+      continue
+    }
+    if (char === ']') {
+      if (start < 0)
+        start = i
+      if (depth)
+        depth--
+      continue
+    }
+    if (/\s/.test(char) && depth === 0)
+      push(i)
+    else if (start < 0)
+      start = i
+  }
+  push(body.length)
+  return items
+}
 
 export function makeRegexClassGroup(separators = ['-', ':']) {
   const escaped = separators.map(s => escapeRegExp(s))
@@ -55,8 +103,8 @@ export function parseVariantGroup(str: string | MagicString, separators = ['-', 
         groupsByOffset.set(groupOffset, group)
 
         // Whitespace inside quoted arbitrary values belongs to the utility, not the group.
-        for (const itemMatch of [...body.matchAll(itemRE)]) {
-          const itemOffset = bodyOffset + itemMatch.index!
+        for (const itemMatch of splitVariantGroupBody(body)) {
+          const itemOffset = bodyOffset + itemMatch.index
           let innerItems = groupsByOffset.get(itemOffset)?.items
           if (innerItems) {
             // We won't need to look up this group from this offset again.
@@ -66,8 +114,8 @@ export function parseVariantGroup(str: string | MagicString, separators = ['-', 
           else {
             innerItems = [{
               offset: itemOffset,
-              length: itemMatch[0].length,
-              className: itemMatch[0],
+              length: itemMatch.value.length,
+              className: itemMatch.value,
             }]
           }
           for (const item of innerItems) {
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -68,7 +68,8 @@ export function makeRegexClassGroup(separators = ['-', ':']) {
     // Quoted body values can contain literal brackets, as in `content-['[']`.
     const nestedBracket = '\\[[^\\[\\]]*\\]'
     const quotedValue = `'(?:\\\\.|[^'\\\\])*'|"(?:\\\\.|[^"\\\\])*"`
-    const prefixBracket = `\\[&?>?:?(?:[^\\s\\[\\]]|${nestedBracket})*\\]`
+    // The character class already accepts &, > and :; optional copies create exponentially many prefix parses.
+    const prefixBracket = `\\[(?:[^\\s\\[\\]]|${nestedBracket})*\\]`
     const bodyBracket = `\\[(?:[^\\[\\]'"]|${quotedValue}|${nestedBracket})*\\]`
     regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|${prefixBracket})+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|${bodyBracket})+?)\\)(?!\\s*?=>)`, 'gm')
   }
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -8,7 +8,7 @@ const regexCache: Record<string, RegExp> = {}
 export function makeRegexClassGroup(separators = ['-', ':']) {
   const escaped = separators.map(s => escapeRegExp(s))
   const key = escaped.join('|')
-  if (!regexCache[key])
+  if (!regexCache[key]) {
     // The body accepts `@` and `*` because the container and children variants
     // are valid inside a group, not just as its prefix. A single unmatchable
     // character makes the whole group fail to match, so omitting them silently
@@ -17,7 +17,12 @@ export function makeRegexClassGroup(separators = ['-', ':']) {
     // Stopping at the inner `]` leaves `[&[open]]:(...)` or `hover:([&[open]]:...)`
     // unexpanded, and the unmatched group can later produce malformed CSS.
     // Quoted body values can contain literal brackets, as in `content-['[']`.
-    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?(?:[^\\s\\[\\]]|\\[[^\\[\\]]*\\])*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[(?:[^\\[\\]'\"]|'(?:\\\\.|[^'\\\\])*'|\"(?:\\\\.|[^\"\\\\])*\"|\\[[^\\[\\]]*\\])*\\])+?)\\)(?!\\s*?=>)`, 'gm')
+    const nestedBracket = '\\[[^\\[\\]]*\\]'
+    const quotedValue = `'(?:\\\\.|[^'\\\\])*'|"(?:\\\\.|[^"\\\\])*"`
+    const prefixBracket = `\\[&?>?:?(?:[^\\s\\[\\]]|${nestedBracket})*\\]`
+    const bodyBracket = `\\[(?:[^\\[\\]'"]|${quotedValue}|${nestedBracket})*\\]`
+    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|${prefixBracket})+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|${bodyBracket})+?)\\)(?!\\s*?=>)`, 'gm')
+  }
   regexCache[key].lastIndex = 0
   return regexCache[key]
 }
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -13,10 +13,11 @@ export function makeRegexClassGroup(separators = ['-', ':']) {
     // are valid inside a group, not just as its prefix. A single unmatchable
     // character makes the whole group fail to match, so omitting them silently
     // left the group unexpanded.
-    // The bracket alternative stops at `]` rather than running to the next
-    // whitespace, otherwise a group nested inside an arbitrary variant swallows
-    // the outer `]:(` and captures the wrong prefix.
-    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?[^\\s\\]]*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
+    // The bracket alternative stops at the matching `]`, allowing a nested
+    // attribute bracket, rather than stopping at the first inner `]`.
+    // Without this, `[&[aria-selected=true]]:(...)` is left unexpanded and the
+    // unmatched group body is later emitted as malformed CSS.
+    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?(?:[^\\s\\[\\]]|\\[[^\\[\\]]*\\])*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
   regexCache[key].lastIndex = 0
   return regexCache[key]
 }
--- a/packages-engine/core/src/utils/variant-group.ts
+++ b/packages-engine/core/src/utils/variant-group.ts
@@ -4,6 +4,7 @@ import { notNull } from '../utils'
 import { escapeRegExp } from './escape'
 
 const regexCache: Record<string, RegExp> = {}
+const itemRE = /(?:[^\s'"]|'(?:\\.|[^'\\])*'|"(?:\\.|[^"\\])*")+/g
 
 export function makeRegexClassGroup(separators = ['-', ':']) {
   const escaped = separators.map(s => escapeRegExp(s))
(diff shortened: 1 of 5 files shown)


packages-engine/core/src/utils/variant-group.ts
```diff
@@ -13,10 +13,11 @@ export function makeRegexClassGroup(separators = ['-', ':']) {
...
     // left the group unexpanded.
-    // The bracket alternative stops at `]` rather than running to the next
-    // whitespace, otherwise a group nested inside an arbitrary variant swallows
-    // the outer `]:(` and captures the wrong prefix.
-    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?[^\\s\\]]*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
+    // The bracket alternative stops at the matching `]`, allowing a nested
+    // attribute bracket, rather than stopping at the first inner `]`.
+    // Without this, `[&[aria-selected=true]]:(...)` is left unexpanded and the
+    // unmatched group body is later emitted as malformed CSS.
+    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?(?:[^\\s\\[\\]]|\\[[^\\[\\]]*\\])*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
```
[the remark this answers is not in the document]
CNSeniorious000: Fixed in ba68e2f3: the group-body branch now consumes nested attribute brackets as well. Added regression coverage for your example, both string and MagicString inputs, nested groups with a following sibling, and the resulting preset-wind4 shortcut CSS. Quoted literal brackets such as `content-['[']` are preserved too. All 144 core/variant-group tests pass, and Lightning CSS accepts both reproductions.

packages-engine/core/src/utils/variant-group.ts
```diff
@@ -13,10 +13,11 @@ export function makeRegexClassGroup(separators = ['-', ':']) {
...
     // left the group unexpanded.
-    // The bracket alternative stops at `]` rather than running to the next
-    // whitespace, otherwise a group nested inside an arbitrary variant swallows
-    // the outer `]:(` and captures the wrong prefix.
-    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?[^\\s\\]]*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
+    // Both the prefix and body must consume nested attribute brackets as a unit.
+    // Stopping at the inner `]` leaves `[&[open]]:(...)` or `hover:([&[open]]:...)`
+    // unexpanded, and the unmatched group can later produce malformed CSS.
+    // Quoted body values can contain literal brackets, as in `content-['[']`.
+    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?(?:[^\\s\\[\\]]|\\[[^\\[\\]]*\\])*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[(?:[^\\[\\]'\"]|'(?:\\\\.|[^'\\\\])*'|\"(?:\\\\.|[^\"\\\\])*\"|\\[[^\\[\\]]*\\])*\\])+?)\\)(?!\\s*?=>)`, 'gm')
```
hannoeru, line 20: suggestion: Give the two repeated patterns a name

The regex repeats the same two parts. One part matches a nested bracket. The other part matches a quoted value. Move each part into a named variable. Then use the variables in the regex. This keeps both halves equal, and it makes the long line easier to read.

I checked the code below. It builds the same regex as the current one.

```ts
const nestedBracket = '\\[[^\\[\\]]*\\]'
const quotedValue = `'(?:\\\\.|[^'\\\\])*'|"(?:\\\\.|[^"\\\\])*"`
const prefixBracket = `\\[&?>?:?(?:[^\\s\\[\\]]|${nestedBracket})*\\]`
const bodyBracket = `\\[(?:[^\\[\\]'\"]|${quotedValue}|${nestedBracket})*\\]`

regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|${prefixBracket})+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|${bodyBracket})+?)\\)(?!\\s*?=>)`, 'gm')
```

packages-engine/core/test/variant-group.test.ts
```diff
@@ -58,12 +59,27 @@ describe('variant-group', () => {
...
 
+  it.each(['string', 'MagicString'])('attribute selectors inside a group body (%s)', (inputType) => {
+    const cases = [
+      ['hover:([&[aria-selected=true]]:bg-accent text-accent)', 'hover:[&[aria-selected=true]]:bg-accent hover:text-accent'],
+      ['[&[open]]:(hover:([&[disabled]]:p-1 p-2)) focus:(m-1 m-2)', '[&[open]]:hover:[&[disabled]]:p-1 [&[open]]:hover:p-2 focus:m-1 focus:m-2'],
+      ['hover:(content-[\'[\'] p-2)', 'hover:content-[\'[\'] hover:p-2'],
+      ['hover:(content-["["] p-2)', 'hover:content-["["] hover:p-2'],
+    ]
+    for (const [input, expected] of cases) {
+      const expanded = inputType === 'string' ? expandVariantGroup(input) : expandVariantGroup(new MagicString(input)).toString()
```
hannoeru, line 78: nitpick: Move the expander function into the test table

The test body picks the expander with an if-statement. But the test name already tells you the type. Put the expander function in the table instead. Then you can delete the if-statement.

```ts
it.each([
  ['string', (s: string) => expandVariantGroup(s)],
  ['MagicString', (s: string) => expandVariantGroup(new MagicString(s)).toString()],
])('attribute selectors inside a group body (%s)', (_name, expand) => {
  for (const [input, expected] of cases)
    expect(expand(input)).toEqual(expected)
})
```

packages-engine/core/test/shortcuts.test.ts
```diff
@@ -254,4 +255,19 @@ describe('shortcuts', async () => {
       .dark .test-inline-body-with-variant{margin:3px;padding:0.5rem;}"
     `)
   })
+
+  it.each([
+    ['[&[aria-selected=true]]:(bg-red/oklab text-blue/oklab)', '[&[aria-selected=true]]:bg-red/oklab [&[aria-selected=true]]:text-blue/oklab'],
+    ['hover:([&[aria-selected=true]]:bg-red/oklab text-blue/oklab)', 'hover:[&[aria-selected=true]]:bg-red/oklab hover:text-blue/oklab'],
+  ])('attribute selector groups generate the expanded shortcut CSS: %s', async (grouped, expanded) => {
+    const uno = await createGenerator({ presets: [presetWind4()], shortcuts: { button: grouped } })
```
hannoeru, line 263: suggestion: Test the exact case from the bug report

This test uses `bg-red/oklab`. The bug report used `bg-accent/9/oklab` with a `light-dark(...)` colour. That exact case is only checked as a string, not as CSS. Also, this test checks the CSS with prettier. The bug report failed in Lightning CSS, so check the CSS with Lightning CSS too.

I ran the missing case by hand. It passes. So this is a new guard, not a bug fix.

```ts
theme: {
  colors: {
    accent: 'light-dark(#a65e2b, #c99076)',
    'accent-strong': 'light-dark(#a64f4f, #cb7676)',
  },
},
shortcuts: { button: '[&[aria-selected=true]]:(bg-accent/9/oklab text-accent-strong/oklab)' },

// then, after generating `css`:
import { transform } from 'lightningcss'
expect(() => transform({ filename: 'test.css', code: Buffer.from(css), minify: true })).not.toThrow()
```

packages-engine/core/src/utils/variant-group.ts
```diff
@@ -13,10 +13,11 @@ export function makeRegexClassGroup(separators = ['-', ':']) {
...
     // character makes the whole group fail to match, so omitting them silently
     // left the group unexpanded.
-    // The bracket alternative stops at `]` rather than running to the next
-    // whitespace, otherwise a group nested inside an arbitrary variant swallows
-    // the outer `]:(` and captures the wrong prefix.
-    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?[^\\s\\]]*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
+    // Both the prefix and body must consume nested attribute brackets as a unit.
+    // Stopping at the inner `]` leaves `[&[open]]:(...)` or `hover:([&[open]]:...)`
+    // unexpanded, and the unmatched group can later produce malformed CSS.
+    // Quoted body values can contain literal brackets, as in `content-['[']`.
```
hannoeru, line 19: issue: A space inside a quoted body value splits the value into two tokens

The quoted alternative added here lets a quoted value hold a space. But the body is still split with `\S+` just below, so the space ends the token:

```ts
expandVariantGroup("hover:(content-['a b'] p-2)")
// => "hover:content-['a hover:b'] hover:p-2"   (corrupted)
```

This split predates this change. I ran the previous regex on the same input, and it produced the same corrupted result, so this is not a regression. But the quoted alternative now makes quoted values a first-class case, so it may be worth handling here. A quote-aware split keeps the value whole; I tested the matches below:

```ts
const ITEM_RE = /(?:[^\s'"]|'(?:\\.|[^'\\])*'|"(?:\\.|[^"\\])*")+/g
for (const itemMatch of [...body.matchAll(ITEM_RE)]) {
  // ...
}
// "content-['a b'] p-2"                 => content-['a b'], p-2
// "content-['['] p-2"                   => content-['['], p-2
// "bg-accent/9/oklab text-accent-…"     => two tokens, unchanged
```

packages-engine/core/src/utils/variant-group.ts
CNSeniorious000: Addressed in 279d00fa: extracted `nestedBracket`, `quotedValue`, `prefixBracket`, and `bodyBracket` as suggested. Rebuilt core and verified the variant-group and transformer tests.

packages-engine/core/test/variant-group.test.ts
CNSeniorious000: Addressed in 113e5b14: the test table now carries each expander directly, removing the conditional from the assertions.

packages-engine/core/test/shortcuts.test.ts
CNSeniorious000: Addressed in 45a960cc: added the exact `bg-accent/9/oklab` and `light-dark(...)` reproduction, compared its CSS against explicitly expanded utilities, and checked Lightning CSS minification for all three cases. Declared Lightning CSS as a core test dependency using the existing catalog. All 23 shortcut tests pass.

packages-engine/core/src/utils/variant-group.ts
CNSeniorious000: Fixed separately in 2875b994 using the quote-aware item matcher. Added string and MagicString coverage for spaces in single/double quotes, escaped quotes, literal brackets, and nested groups followed by siblings. Confirmed the new cases fail before the fix and pass after rebuilding core. The complete core and variant-group transformer suite passes: 145 tests across 30 files; ESLint and frozen-lockfile validation also pass.

CNSeniorious000:
Follow-up after local Claude review: 4616120b removes the redundant `&?>?:?` from `prefixBracket`. Those characters are already accepted by the following character class; the optional copies create exponentially many ways to parse repeated arbitrary-variant prefixes when no complete group follows.

Correction to the motivation stated in b8d8e854: that commit changed body tokenization to preserve bracketed values, but did not fix this backtracking path. The expensive matcher was the outer group regex, not `itemRE`. This follow-up addresses that root cause without rewriting the published history.

Validation:
- Added four repeated-prefix cases, each covering absent, unclosed, and valid groups through both string and MagicString inputs; also covered the scanner's bracketed values containing spaces.
- Rebuilt core; independent Claude verification passed 233 tests across 32 files. ESLint passed.
- Original `light-dark(...)` / `/9/oklab` shortcut matched the expanded reference; Lightning CSS minification succeeded with zero warnings.
- On Node, the original repeated-selector shape dropped from about 248 ms for 20 repeats to under 1 ms. Larger fixed inputs also complete; this removes the exponential ambiguity, not a claim that the entire parser is linear.

Claude reviewed the final diff again and reported no blocking findings.

packages-engine/core/src/utils/variant-group.ts
```diff
@@ -4,19 +4,75 @@ import { notNull } from '../utils'
...
+    if (quote) {
+      if (escaped)
+        escaped = false
+      else if (char === '\\')
+        escaped = true
+      else if (char === quote)
+        quote = ''
+      continue
+    }
+    if (char === '\'' || char === '"') {
```
hannoeru, line 29: note: splitVariantGroupBody and bodyBracket encode the same grammar twice

bodyBracket (line 73) only accepts quotes via `quotedValue`, which is always inside `[...]`. The scanner here treats every `'`/`"` as a quote start regardless of bracket depth. A quote outside brackets cannot survive the regex, so the divergence is unreachable today — but the two can drift.

Recommendation (make them agree, one-line guard):

    if (depth > 0 && (char === '\'' || char === '"')) {

and add a comment that this mirrors `bodyBracket` (quotes are only valid inside `[...]`).

packages-engine/core/test/variant-group.test.ts
```diff
@@ -58,12 +59,50 @@ describe('variant-group', () => {
...
+  it.each(['[&[data-a=b]]:', '[&foo]:', '[>foo]:', '[:foo]:'])('repeated arbitrary variant prefixes: %s', (variant) => {
+    // Optional prefix markers used to multiply backtracking paths at each bracket.
+    const prefix = variant.repeat(32)
+    const cases = [
+      [`${prefix}text-red`, `${prefix}text-red`],
+      [`${prefix}(p-1 p-2`, `${prefix}(p-1 p-2`],
+      [`${prefix}(p-1 p-2)`, `${prefix}p-1 ${prefix}p-2`],
+    ]
+    for (const [input, expected] of cases) {
+      expect(expandVariantGroup(input)).toBe(expected)
```
hannoeru, line 101: nitpick: The branch test re-inlines both expanders

The sibling test (lines 70-73) parameterizes `string` vs `MagicString`; this block writes both calls out by hand.

Recommendation — hoist a shallow helper so both tests share one expander shape:

    const expanders = {
      string: (s: string) => expandVariantGroup(s),
      MagicString: (s: string) => expandVariantGroup(new MagicString(s)).toString(),
    }
    // here:
    expect(expanders.string(input)).toBe(expected)
    expect(expanders.MagicString(input)).toBe(expected)

packages-engine/core/src/utils/variant-group.ts
```diff
@@ -4,19 +4,75 @@ import { notNull } from '../utils'
 import { escapeRegExp } from './escape'
 
 const regexCache: Record<string, RegExp> = {}
+function splitVariantGroupBody(body: string) {
```
hannoeru, line 7: nitpick: add a blank line before splitVariantGroupBody

`const regexCache` (line 6) sits directly against the function with no blank line, unlike the rest of the file.

Recommendation: insert one blank line between lines 6 and 7.

packages-engine/core/src/utils/variant-group.ts
```diff
@@ -47,8 +103,9 @@ export function parseVariantGroup(str: string | MagicString, separators = ['-',
         const group: VariantGroup = { length: from.length, items: [] }
         groupsByOffset.set(groupOffset, group)
 
-        for (const itemMatch of [...body.matchAll(/\S+/g)]) {
-          const itemOffset = bodyOffset + itemMatch.index!
+        // Whitespace inside quoted arbitrary values belongs to the utility, not the group.
+        for (const itemMatch of splitVariantGroupBody(body)) {
```
hannoeru, line 107: issue: collapseVariantGroup still splits naively, so round-trips break

The quote-aware split was added on this line, but the inverse `collapseVariantGroup` still does `str.split(/\s+/g)` (line 185). Result:

    collapseVariantGroup("hover:content-['a b'] hover:p-2", ['hover:'])
    // => "hover:(content-['a p-2) b']"   (corrupted)

Same for `grid-cols-[1fr 2fr]`. This feeds `sortRules` (virtual-shared/integration/src/sort-rules.ts:45), so a quoted/bracket value with a space is corrupted on the way back.

Recommendation — share the scanner:

    // collapseVariantGroup, line 185:
    .split(/\s+/g)   →   splitVariantGroupBody(str).map(item => item.value)

Check the leading/trailing-whitespace edge, which `split(/\s+/g)` handled by emitting an empty string and this change drops.

packages-engine/core/src/utils/variant-group.ts
```diff
@@ -4,19 +4,75 @@ import { notNull } from '../utils'
...
     // left the group unexpanded.
-    // The bracket alternative stops at `]` rather than running to the next
-    // whitespace, otherwise a group nested inside an arbitrary variant swallows
-    // the outer `]:(` and captures the wrong prefix.
-    regexCache[key] = new RegExp(`((?:[!@*<~\\w+:_-]|\\[&?>?:?[^\\s\\]]*\\])+?)(${key})\\(((?:[~!<>@*\\w\\s:/\\\\,%#.$?-]|\\[[^\\]]*?\\])+?)\\)(?!\\s*?=>)`, 'gm')
+    // Both the prefix and body must consume nested attribute brackets as a unit.
+    // Stopping at the inner `]` leaves `[&[open]]:(...)` or `hover:([&[open]]:...)`
+    // unexpanded, and the unmatched group can later produce malformed CSS.
+    // Quoted body values can contain literal brackets, as in `content-['[']`.
+    const nestedBracket = '\\[[^\\[\\]]*\\]'
```
hannoeru, line 69: note: double-nested brackets in a prefix still don't expand

`[&[[data-a=b]]]:(p-1 p-2)` stays unexpanded because `nestedBracket` allows one inner level only. It emits no malformed CSS, so #5326's fallback clause is still met.

Recommendation: if arbitrary depth is wanted, a regex can't recurse — use the same depth scanner already in `splitVariantGroupBody` for the prefix. Otherwise, add a test asserting the single-level limit so the behavior is pinned, e.g.

    expect(expandVariantGroup('[&[[data-a=b]]]:(p-1 p-2)'))
      .toBe('[&[[data-a=b]]]:(p-1 p-2)')   // documented as unexpanded

packages-engine/core/src/utils/variant-group.ts
CNSeniorious000: Addressed in cf750b6: quote handling now requires `depth > 0`, with a comment linking the rule to `bodyBracket`. Added coverage for a literal apostrophe outside brackets so the shared scanner does not swallow subsequent utilities.

packages-engine/core/test/variant-group.test.ts
CNSeniorious000: Addressed in 671e3ea: hoisted the string and MagicString expanders into one shared object and reused them in the attribute-selector, repeated-prefix, and nesting-limit tests.

packages-engine/core/src/utils/variant-group.ts
CNSeniorious000: Added the blank line in cf750b6. ESLint passes.

packages-engine/core/src/utils/variant-group.ts
CNSeniorious000: Fixed in cf750b6: `collapseVariantGroup` now shares the bracket/quote-aware scanner and preserves the previous leading/trailing whitespace normalization, including empty and whitespace-only inputs. `sortRules` had a separate whitespace split before collapsing, so it now uses the same scanner too. Added round-trip coverage for quoted spaces, escaped quotes, literal brackets, nested attribute selectors, and grid values, plus sorting regressions. The new corruption cases failed before the fix; after rebuilding core, all 171 tests across 31 files pass, including core, variant-group transformer, and sort-rules tests. ESLint passes.

packages-engine/core/src/utils/variant-group.ts
CNSeniorious000: Addressed in 671e3ea with string and MagicString tests asserting that `[&[[data-a=b]]]:(p-1 p-2)` remains unchanged, plus a comment documenting the one-inner-level prefix limit. Arbitrary-depth prefix parsing remains outside this fix; the supported boundary is now pinned by tests.

zyyv (approved):

Issue #5293, opened 2026-09-01
[vite] Global CSS HMR misses utilities discovered in lazy modules in 66.9.1


ntnyq:
### UnoCSS version

`66.9.1`

### Describe the bug

In Vite development mode, utilities that are first discovered in a
dynamically imported module are generated by UnoCSS, but the browser does not
reload the global UnoCSS virtual module. The newly rendered content therefore
remains unstyled until the page is manually reloaded.

This is a regression from `66.8.1`:

| Version | Five fresh dev-server runs |
| --- | --- |
| `66.8.1` | `display: grid` in 5/5 runs |
| `66.9.1` | `display: block` in 5/5 runs |

Production builds include the expected utilities. The problem is specific to
development-mode incremental extraction/HMR.

### Reproduction

https://github.com/ntnyq/unocss-vite-hmr-lazy-repro

```bash
pnpm install --frozen-lockfile
pnpm exec playwright install chromium
pnpm reproduce
```

The reproduction uses vanilla Vite and one dynamic import; Vue and routing are
not involved. The script starts five fresh Vite dev servers so that UnoCSS's
in-memory token cache cannot hide the issue.

Actual output with `66.9.1`:

```text
run 1: display=block, grid-template-columns=none
run 2: display=block, grid-template-columns=none
run 3: display=block, grid-template-columns=none
run 4: display=block, grid-template-columns=none
run 5: display=block, grid-template-columns=none
Bug reproduced 5/5 times.
```

Expected: after `src/lazy.tsx` is imported and renders an element with
`grid grid-cols-3 gap-4`, its computed `display` should become `grid` without a
full-page reload.

### Suspected cause

This appears to be a follow-up regression from #5284 / commit `75103d5`.

Global mode now stores the UnoCSS entry using the internal virtual-module ID:

```ts
const virtualId = toViteVirtualId(entry)
entries.add(virtualId)
return virtualId
```

However, `sendUpdate()` still sends `mod.url` directly for both HMR paths:

```ts
return {
  acceptedPath: mod.url,
  path: mod.url,
  timestamp: lastServedTime,
  type: 'js-update',
}
```

The observed WebSocket update uses the internal ID:

```text
\0/__uno.css
```

The Vite browser client registered the URL-safe public path instead:

```text
/@id/__x00__/__uno.css
```

`hotModulesMap.get(path)` therefore misses and silently ignores the update. No
timestamped CSS request is made. Rewriting `path` and `acceptedPath` to
[...]

PR #5335, opened 2026-09-11
fix(vite)!: drive global HMR through Vite's hotUpdate hook, drop Vite 5 support

### LINKED ISSUE #5331: [vite] Global CSS HMR stops updating with Vite 8.3
### UnoCSS version

66.10.2

### Describe the bug

After upgrading Vite from **8.2.2 to 8.3.0**, global-mode CSS HMR stops applying utilities first discovered in a dynamically imported module. UnoCSS generates the CSS on the server, but the browser keeps the earlier stylesheet. A full page reload makes the styles available.

This reproduces in a vanilla Vite app with only one lazy module; Vue, routing, APIs, and application-specific configuration are not involved.

This is related to #5293 and #5298, but the mismatch is now reversed after the Vite upgrade. #5298 fixed the encoded-path requirement on Vite 8.2.2; Vite 8.3 uses the raw virtual-module URL for HMR registration.

### Reproduction

https://github.com/ntnyq/unocss-vite83-hmr-repro

```sh
pnpm install --frozen-lockfile
pnpm exec playwright install chromium
pnpm reproduce
```

The script starts three fresh dev servers and browser contexts, clicks a button to import `src/lazy.tsx`, and checks the rendered element's computed style. The lazy module uses `grid grid-cols-3 gap-4`.

Expected: `display: grid` and three columns without reloading.

Actual with Vite 8.3.0: `display: block`, `grid-template-columns: none`, in all three runs. `pnpm reproduce` exits with status 1 to report the missing CSS.

For a manual reproduction, run `pnpm dev`, open the printed URL, and click **Load lazy panel**. Reload and click again to see the generated styles apply. Restart the dev server before repeating the initial-load reproduction because the in-memory token cache can hide it.

### Version comparison

UnoCSS is fixed at 66.10.2 throughout. Both Vite versions are pinned in the reproduction's lockfile.

| Command | Vite | Diagnostic workaround | Styled without reload |
| --- | --- | --- | --- |
| `pnpm verify:82` | 8.2.2 | No | 3/3 |
| `pnpm reproduce` | 8.3.0 | No | 0/3 |
| `pnpm verify:workaround` | 8.3.0 | Yes | 3/3 |

### HMR path mismatch

A failing run reports:

```json
{
  "run": 1,
  "display": "block",
  "columns": "none",
  "registeredPath": "\u0000/__uno.css",
  "browserHasGrid": false,
  "serverHasGrid": true,
  "updatePaths": ["/@id/__x00__/__uno.css"]
}
```

In Vite 8.3, the transformed CSS module registers:

```js
import.meta.hot = __vite__createHotContext("\u0000/__uno.css")
```

UnoCSS's `sendUpdate()` still uses `toViteClientPath(mod.url)` and sends `/@id/__x00__/__uno.css` for both `path` and `acceptedPath`. Vite's `hotModulesMap.get(path)` does not find the registered raw ID and silently skips the update.

The relevant Vite change is https://github.com/vitejs/vite/pull/23172 / commit `67a680767317f8e2cb28b6b0500192f993a567cf`, included between 8.2.2 and 8.3.0. It changes HMR context registration from `normalizeHmrUrl(importerModule.url)` to `importerModule.url`, with URL encoding handled when importing the updated module.

The reproduction's `--workaround` mode rewrites only the outgoing UnoCSS update paths to the raw ID. It leaves extraction, CSS generation, and installed packages untouched, and makes the browser update immediately.

This workaround is diagnostic, **not a proposed unconditional reversal of #5298**: Vite 8.2.2 still registers the encoded path, so an upstream fix needs to preserve compatibility with older Vite versions.

### System Info

- macOS
- Node.js 26.8.1 (automated reproduction runtime)
- pnpm 12.3.4
- Chromium 153.0.8010.12 via Playwright 1.63.0
- UnoCSS 66.10.2, `presetWind3()`, default global mode
- Vite 8.2.2 and 8.3.0
- No dependency patches in the reproduction

The original application also reproduced the issue under Node.js 24.19.0.

### Validations

- [x] Read the contributing guidelines.
- [x] Searched existing issues and did not find a report for this Vite 8.3 regression.
- [x] Confirmed this is a concrete bug using real browser computed styles and HMR payloads.
- [x] Provided a minimal reproduction with exact versions and a comparison against Vite 8.2.2.

hannoeru:
## Motivation

Vite 8.3 removed the HMR URL concept ([vitejs/vite#23172](https://github.com/vitejs/vite/pull/23172)). Global mode built the HMR `Update` payload by hand and guessed the hot-context path from the installed Vite version (`toViteClientPath`). After the upgrade the client no longer found the encoded path, so utilities first discovered in a dynamically imported module were generated on the server but never applied in the browser.

Closes #5331.

The same hand-built update path caused #5293 (utilities discovered in lazy modules missed in 66.9.1) and forced the `hmrTopLevelAwait` workaround for the WebKit crash in #5007.

## Change

- Drive global HMR through Vite's `hotUpdate` hook and return the regenerated virtual CSS module nodes. Vite computes `path` and `acceptedPath` itself, so UnoCSS no longer touches HMR URLs or reads the Vite version.
- Reload config files through the same `hotUpdate` flow.
- Lazily imported modules are transformed after the CSS module already loaded, so they never trigger `hotUpdate`. A `unocss:refresh` custom event asks the browser to re-import the CSS module for that case (#5331, #5293).
- Compare every entry against the hashes captured before regeneration, so entries that share a layer (for example `uno.css` and `uno.css?inline`) are all marked stale when the CSS changes.
- Delete `toViteClientPath`, the Vite version check, the `unocss:hmr` round-trip, the `__uno_hash_` marker, and the `hmrTopLevelAwait` option. The injected code no longer uses top-level await, which also removes the WebKit crash from #5007.
- Raise the minimum Vite version to 6 for `hotUpdate` and `environment.hot` in `@unocss/vite` and `@unocss/astro`.

## Breaking changes

- Vite 5 is no longer supported.
- The `hmrTopLevelAwait` option is removed. It only controlled the injected code that this change deletes.

## Testing

- `test/virtual-modules.test.ts`: the CSS module is returned from `hotUpdate` only when the generated CSS changes, HTML is left to Vite, a lazy transform sends `unocss:refresh`, and every entry for a shared layer is returned when its CSS changes.
- A `vite-compat` CI job runs the watcher-driven and lazy-refresh tests against Vite 6, 7, and 8.
- `pnpm vitest run`, `pnpm run typecheck`, `eslint`, and the `@unocss/vite` build are clean.
- The #5331 Playwright reproduction is the end-to-end browser check.

### DIFF
--- a/examples/sveltekit/vite.config.ts
+++ b/examples/sveltekit/vite.config.ts
@@ -4,11 +4,7 @@ import { defineConfig } from 'vite'
 
 export default defineConfig({
   plugins: [
-    UnoCSS({
-      // top-level await in the HMR code crashes WebKit builds without the fix
-      // for https://bugs.webkit.org/show_bug.cgi?id=242740 (see #5007)
-      hmrTopLevelAwait: false,
-    }),
+    UnoCSS(),
     sveltekit(),
   ],
 })
--- a/packages-integrations/astro/package.json
+++ b/packages-integrations/astro/package.json
@@ -38,7 +38,7 @@
     "dev": "tsdown --config-loader unrun --watch"
   },
   "peerDependencies": {
-    "vite": "^5.0.0-0 || ^6.0.0-0 || ^7.0.0-0 || ^8.0.0-0"
+    "vite": "^6.0.0-0 || ^7.0.0-0 || ^8.0.0-0"
   },
   "peerDependenciesMeta": {
     "vite": {
--- a/packages-integrations/vite/package.json
+++ b/packages-integrations/vite/package.json
@@ -44,7 +44,7 @@
     "dev": "tsdown --config-loader unrun --watch"
   },
   "peerDependencies": {
-    "vite": "^5.0.0-0 || ^6.0.0-0 || ^7.0.0-0 || ^8.0.0-0"
+    "vite": "^6.0.0-0 || ^7.0.0-0 || ^8.0.0-0"
   },
   "dependencies": {
     "@jridgewell/remapping": "catalog:build",
--- a/packages-integrations/vite/src/config-hmr.ts
+++ b/packages-integrations/vite/src/config-hmr.ts
@@ -1,5 +1,10 @@
 import type { UnocssPluginContext } from '@unocss/core'
 import type { Plugin } from 'vite'
+import { normalizePath } from 'vite'
+
+export function isConfigSource(ctx: UnocssPluginContext, file: string) {
+  return ctx.getConfigFileList().some(source => normalizePath(source) === file)
+}
 
 export function ConfigHMRPlugin(ctx: UnocssPluginContext): Plugin | undefined {
   const { ready } = ctx
@@ -9,24 +14,16 @@ export function ConfigHMRPlugin(ctx: UnocssPluginContext): Plugin | undefined {
       await ctx.updateRoot(config.root)
     },
     async configureServer(server) {
-      const { sources } = await ready
-
+      await ready
       ctx.uno.config.envMode = 'dev'
-      if (!sources.length)
+      server.watcher.add(ctx.getConfigFileList())
+    },
+    async hotUpdate({ file, type }) {
+      if (type === 'delete' || this.environment.name !== 'client')
         return
-
-      server.watcher.add(sources)
-      server.watcher.on('change', async (p) => {
-        if (!sources.includes(p))
-          return
-
-        await ctx.reloadConfig()
-
-        server.ws.send({
-          type: 'custom',
-          event: 'unocss:config-changed',
-        })
-      })
+      if (!isConfigSource(ctx, file))
+        return
+      await ctx.reloadConfig()
     },
   }
 }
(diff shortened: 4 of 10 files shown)


hannoeru:
@copilot Fix the code for all comments in [this review thread](https://github.com/unocss/unocss/pull/5335#pullrequestreview-5184029320).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

hannoeru:
@copilot Fix the code for all comments in [this review thread](https://github.com/unocss/unocss/pull/5335#pullrequestreview-5184239411).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

zyyv (approved):

antfu (approved):

antfu:
I guess we would need to do a v67 major release as this would drop Vite 5 support. Do you have any other breaking changes to include by taking this chance? @zyyv 

Or I think we can do a small breaking major as this seems to affecting the user expereinece quite a lot

PR #5340, opened 2026-09-14
test: add public API snapshot testing with tsnapi

antfubot:
## Description

Adds public API snapshot testing across the monorepo using [`tsnapi`](https://github.com/antfu/tsnapi), so unintended changes to any package's runtime exports or type declarations show up as a reviewable diff in git.

### What's included

- **`test/api-snapshot.test.ts`** — a single test using tsnapi's `describePackagesApiSnapshots()`. It auto-discovers workspace packages and snapshots the publishable libraries under `packages-engine/*`, `packages-integrations/*`, and `packages-presets/*`, skipping non-lib workspaces (docs/playground/etc.), the deprecated presets, the private `@unocss/vscode`, and `@unocss/reset` (ships static CSS).
- **Committed snapshots** under `test/__snapshots__/tsnapi/` — a `.snapshot.js` (runtime) and `.snapshot.d.ts` (types) per entry point, covering 37 `@unocss/*` packages plus every entry of the `unocss` meta package.
- **`tsnapi`** added to the `build` catalog and as a root devDependency.

### How it works

- Runs as part of the normal test suite (`pnpm test`) and `test:ci` (which already does `nr build && … && nr test`). tsnapi reads built `dist`, and the suite already resolves packages through `exports → dist`, so the build-before-test expectation is unchanged.
- A comparison run fails on any additive or breaking API change, keeping every change visible in the diff. Update intentionally with `pnpm test:update` (`vitest -u`); removals additionally trip tsnapi's breaking-change guard unless `TSNAPI_ALLOW_BREAKING=1` is set.

This PR was created with the help of an agent.


antfu (approved):
This would help us to catch up unintented breaking changes earlier, however it does make the testing process a bit more verbose. I would let you do the call @zyyv

zyyv (approved):
Thanks to @antfu, I’ve learned how to use tsnapi, and I’m happy to accept it.

PR #5343, opened 2026-09-14
feat(preset-wind4): add scrollbar width and color utilities

anandghegde:
### Description

Tailwind CSS v4.3 added `scrollbar-{auto,thin,none}` and `scrollbar-thumb-*` / `scrollbar-track-*` ([tailwindlabs/tailwindcss#19981](https://github.com/tailwindlabs/tailwindcss/pull/19981), [#20019](https://github.com/tailwindlabs/tailwindcss/pull/20019)). `preset-wind4` already has the `scrollbar-gutter-*` utilities from the same release (#5224), but not these. This PR adds them to `rules/scrolls.ts`.

- `scrollbar-auto` / `scrollbar-thin` / `scrollbar-none` set `scrollbar-width`.
- `scrollbar-thumb-<color>` / `scrollbar-track-<color>` set `--un-scrollbar-thumb` / `--un-scrollbar-track` and `scrollbar-color: var(--un-scrollbar-thumb) var(--un-scrollbar-track)`. This matches Tailwind's output. Both variables are registered with `@property` (`<color>`, initial value `#0000`) and get the usual properties-layer fallback, the same as Tailwind's `--tw-scrollbar-*`. That way, setting only one of the two still gives a valid `scrollbar-color`.
- The colors go through `colorResolver`, so theme colors, `/50` alpha, arbitrary values, `$var` and `current`/`transparent`/`inherit` all work. There are also `scrollbar-(thumb|track)-op-<n>` rules, like the other wind4 color utilities (`accent-op-*`, `caret-op-*`, `ring-op-*`).

```html
<div class="overflow-auto scrollbar-thin scrollbar-thumb-gray-400 scrollbar-track-transparent"></div>
```

```css
.scrollbar-thin{scrollbar-width:thin;}
.scrollbar-thumb-gray-400{--un-scrollbar-thumb:color-mix(in srgb, var(--colors-gray-400) var(--un-scrollbar-thumb-opacity), transparent);scrollbar-color:var(--un-scrollbar-thumb) var(--un-scrollbar-track);}
.scrollbar-track-transparent{--un-scrollbar-track:transparent;scrollbar-color:var(--un-scrollbar-thumb) var(--un-scrollbar-track);}
```

Notes:
- Name clashes: no other built-in rule starts with `scrollbar-` apart from `scrollbar-gutter-*`, and those still work. The community `unocss-preset-scrollbar` uses `scrollbar-(thumb|track)-color-*` and `-radius-*`. Neither of those parses as a color, so these new rules don't match them. Its `scrollbar-thin`/`scrollbar-none` emit the same `scrollbar-width` declaration.
- Autocomplete suggests `scrollbar-(thumb|track)-$colors` and the opacity variants.

### Tests

- Added the new utilities to `test/assets/preset-wind4-targets.ts`. The regenerated snapshot only changes for the new entries.
- Added an inline test to `test/preset-wind4.test.ts` that covers the `@property` output and the preflight fallback.
- `pnpm build`, then `pnpm vitest run`: 100 files, 731 passed / 3 skipped.
- `tsgo --noEmit` passes, and `eslint` on the changed files passes.

This PR was written with help from Claude Code, and I reviewed it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QBWFWqjbUQGS1c9LsM2bxX


packages-presets/preset-wind4/src/rules/scrolls.ts
```diff
@@ -1,8 +1,27 @@
...
 import type { Theme } from '../theme'
-import { defineProperty, directionSize } from '../utils'
+import { colorResolver, defineProperty, directionSize, h } from '../utils'
 
 export const scrolls: Rule<Theme>[] = [
+  ['scrollbar-auto', { 'scrollbar-width': 'auto' }],
+  ['scrollbar-thin', { 'scrollbar-width': 'thin' }],
+  ['scrollbar-none', { 'scrollbar-width': 'none' }],
+
+  [/^scrollbar-(thumb|track)-(.+)$/, function* ([, part, body], ctx) {
```
zyyv, line 10: We also need to add parsing for the size.

Refer to https://github.com/unocss-community/unocss-preset-scrollbar

zyyv (changes requested):

packages-presets/preset-wind4/src/rules/scrolls.ts
anandghegde: Happy to add it, but I want to check which "size" you mean, since the two readings pull the PR in quite different directions.

**1. Sizes for the WebKit scrollbar parts**, like `unocss-preset-scrollbar`'s `scrollbar-w-*`, `scrollbar-h-*` and `scrollbar-radius-*` / `scrollbar-(thumb|track)-radius-*`. Those write `--scrollbar-width` / `--scrollbar-height` and apply them through `::-webkit-scrollbar` selectors.

One thing to weigh if that's the intent: `::-webkit-scrollbar` styling is ignored once the element has a standard `scrollbar-width` or `scrollbar-color`, which is what `scrollbar-thin` and `scrollbar-thumb-*` in this PR set. That's the reason the community preset gates the two behind its `compatible` option. So combining them in one preset means `scrollbar-w-2 scrollbar-thin` silently drops the width. In current UnoCSS the WebKit parts are also already reachable without new rules, e.g. `[&::-webkit-scrollbar]:w-2`.

**2. A parsed value for `scrollbar-width` itself**, replacing the three static rules with something like:

```ts
[/^scrollbar-(?:width-)?(.+)$/, ([, w], { theme }) => ({ 'scrollbar-width': h.bracket.cssvar.global(w) })]
```

so `scrollbar-[thin]` and `scrollbar-$my-width` work alongside `scrollbar-auto|thin|none`. The property only accepts `auto | thin | none` per spec, so a length there is invalid CSS, and the pattern needs care not to shadow `scrollbar-gutter-*` or `scrollbar-thumb-*`.

My own preference is to keep this PR at Tailwind v4.3 parity (`scrollbar-width` keywords plus thumb/track colors) and do the WebKit part sizes as a follow-up, where the interaction above can be handled properly. But it's your call. Tell me which one you want and I'll push it.

PR #5355, opened 2026-09-22
fix(eslint-plugin): sort class utilities in arrow/function slot values and helper-call initializer arguments

### LINKED ISSUE #5317: `@unocss/order` does not sort class strings inside `unoVariables` assigned to helpers like `defu`, or inside arrow-function slot values
### UnoCSS version

66.10.0

### Describe the bug

`unocss/order` can sort class strings on matching `unoVariables` when the initializer is a string or a plain object:

```ts
const buttonClassNames = { default: 'pr1 pl1' }
// autofix → { default: 'pl1 pr1' }
```

Theme-style code often looks like this instead:

```ts
export const themeSelect = defu(
  {
    slots: {
      base: () => 'relative disabled:cursor-not-allowed disabled:(opacity-75)',
    },
  },
  {},
)
```

`eslint --fix` does not reorder those classes.

Two AST gaps:

1. **Call initializer.** If a matching variable is assigned `defu(...)` (or any helper that is not in `unoFunctions`), `VariableDeclarator` does not walk the call arguments.
2. **Arrow / function slot values.** Object walking only looks at string literals and nested objects, so `base: () => '...'` and `base: () => { return '...' }` are skipped.

Configuring `unoFunctions: ['defu']` is not enough on its own, because of (2). Configuring `unoVariables: ['^theme']` is not enough on its own, because of (1) and (2).

### Reproduction

```js
// eslint.config.js
export default [
  {
    plugins: { unocss },
    rules: {
      'unocss/order': ['warn', {
        unoVariables: ['^theme'],
        // also reproduces with: unoFunctions: ['defu']
      }],
    },
  },
]
```

```ts
export const themeSelect = defu(
  {
    slots: {
      base: () => 'mr-1 ml-1',
    },
  },
  {},
)
```

Expected after `--fix`:

```ts
export const themeSelect = defu(
  {
    slots: {
      base: () => 'ml-1 mr-1',
    },
  },
  {},
)
```

Actual: no warning and no fix.

Same skip for a block body:

```ts
base: () => { return 'mr-1 ml-1' }
```

Without `defu` and without the arrow function, a plain matching object is fixed:

```ts
const buttonClassNames = { default: 'pr1 pl1' }
```

### System Info

- `@unocss/eslint-plugin` `66.10.0`
- Class strings live in theme objects merged with `defu`, with slot values as `() => '...'`

### Validations

nikolas-sapa:
Fixes #5317.

`unocss/order` only walked `ObjectExpression` values in `unoVariables` declarations. Two AST gaps caused class lists inside helper-wrapped theme definitions to be skipped:

1. **Function slot values.** `base: () => 'mr-1 ml-1'` (and block-body `() => { return '...' }`) were never inspected, because only `Literal` / `ObjectExpression` property values were recursed.
2. **Helper-call initializer.** `const theme = defu({ slots: {...} }, {})` wasn't handled: the `VariableDeclarator` branch only checked `ObjectExpression` inits, so `defu` arguments were never walked even when `defu` is in `unoFunctions`.

Changes:
- Hoist `handleObjectExpression` / `handleCallArguments` to the rule scope so both `VariableDeclarator` and `CallExpression` visitors share them.
- `handleObjectExpression` now recurses into `ArrowFunctionExpression` / `FunctionExpression` bodies (unwraps TS `as`/`satisfies`), and into nested uno-function call values (e.g. `base: clsx(...)`).
- `VariableDeclarator` walks a uno-function init's arguments (e.g. `defu({ slots: {...} })`).
- Guard `SpreadElement` in call arguments (TS type).

Tests: +3 cases covering the arrow-body, defu-initializer, and a helper-call property value — all previously unfixable, now fixed. Baseline 63 tests still pass (66 total).


zyyv:
@nikolas-sapa Can you handle the changes of #5356?

nikolas-sapa:
Maintainer note: this branch includes the exact `#5356` change through shared commit `f10c26a03de1ae92991d89df46c3d5d0678ba32e` (the constructCSS parent-placeholder fix), alongside the ESLint order fix. I also corrected the duplicate order test so it covers the missing nested helper-call case.

